PR #5, opened 2018-11-09
Add /endpoints/{endpoint_id}/storage API path.

atoaca:
Depends on the policy PR.


coriolis/api/v1/views/endpoint_storage_view.py
```diff
@@ -0,0 +1,23 @@
...
+    def transform(key, value):
+        if keys and key not in keys:
+            return
+        yield (key, value)
+
+    return dict(itertools.chain.from_iterable(
+        transform(k, v) for k, v in storage.items()))
+
+
+def single(req, storage):
```
aznashwan: Considering there is no `/endpoint/$ID/storage/$STORAGE_ID` API call (and probably never will) this method will never get used so please remove it.

PR #9, opened 2018-11-29
Add `source_environment` to `base_transfer_action`.

atoaca:


coriolis/api/v1/migrations.py
```diff
@@ -106,11 +109,18 @@ def create(self, req, body):
                 raise exc.HTTPBadRequest(
                     explanation="Invalid destination "
                                 "environment: %s" % message)
+            is_valid, message = (
```
aznashwan: Please move this check in `_validate_migration_input` to be consistent with the dest-env validation (that one has to be there due to the `network_map` and `storage_mappings` situation with the plugins)

coriolis/api/v1/replicas.py
```diff
@@ -91,11 +94,19 @@ def create(self, req, body):
             raise exc.HTTPBadRequest(
                 explanation="Invalid destination "
                             "environment: %s" % message)
+        is_valid, message = (
```
aznashwan: Please move this check in `_validate_create_body()` to be consistent with the dest-env validation.

coriolis/api/v1/migrations.py
```diff
@@ -106,11 +109,18 @@ def create(self, req, body):
...
+                self._endpoints_api.validate_source_environment(
+                    context, origin_endpoint_id, source_environment))
+            if not is_valid:
```
aznashwan: Could you please define a method/decorator in `coriolis.api.v1.utils` which abstracts away this `is_valid, message = ...` patter and apply it both here, in the Replicas file, and for the dest-env validation?

coriolis/worker/rpc/server.py
```diff
@@ -310,6 +310,22 @@ def validate_endpoint_target_environment(
...
 
+    def validate_endpoint_source_environment(
+            self, ctxt, platform_name, source_env):
+        provider = providers_factory.get_provider(
+            platform_name, constants.PROVIDER_TYPE_EXPORT, None)
+        source_env_schema = provider.get_source_environment_schema()
+
+        is_valid = True
+        message = None
+        try:
```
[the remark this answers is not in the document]
aznashwan: Obsoleted by previous suggestion to define a decorator in `coriolis.api.v1.utils`

aznashwan:
LGTM, merging, thanks a lot!

PR #10, opened 2018-12-05
Validate the 'destination_environment' in API before appending the 'storage_mappings' to it.

aznashwan:


aznashwan:
Tested and confirmed ready. Merging.

PR #12, opened 2018-12-11
Improve resiliency of various API calls and a minor refacor.

aznashwan:


aznashwan:
Merging as-is as all changes are trivial.

PR #13, opened 2018-12-11
Add `source_environment` to `base_transfer_action`

atoaca:


coriolis/providers/base.py
```diff
@@ -309,39 +309,47 @@ def restore_replica_disk_snapshots(self, ctxt, connection_info,
...
+                        instance_name, export_path):
         """Exports the given instance.
 
          Exports the instance given by its name from the given source cloud
         to the provided export directory path using the given connection info.
         """
         pass
 
+    @abc.abstractmethod
+    def get_source_environment_schema(self):
```
aznashwan: Please define a separate `BaseExportInstanceProvider` class for this method and make both `BaseExportProvider` and `BaseReplicaExportProvider` inherit it (this way it's only required in the `BaseExportProvider` ABC and not the Replica one as well)

aznashwan:
LGTM, thanks!

PR #14, opened 2018-12-16
Disable pipelined transfers for SFTP

gabriel-samfira:
This should fix the intermittent lock-ups we have been seeing during OSMorphing.


aznashwan:
Tested and merged, thank you!

PR #15, opened 2018-12-17
Remove check for mandatory pattern when listing endpoint instances.

aznashwan:


aznashwan:
Merging ASAP as this is a critical issue.

PR #18, opened 2019-02-20
Fix os_mount()

gabriel-samfira:
For some reason, we were ignoring the first device in the list because of the assumption that it belonged to the worker VM. This change looks in /proc/mounts for already mounted devices. This way there is no way to confuse the worker VM root device for the root device of the instance we are migrating

### DIFF
--- a/coriolis/osmorphing/osmount/base.py
+++ b/coriolis/osmorphing/osmount/base.py
@@ -222,6 +222,16 @@ class BaseLinuxOSMountTools(BaseSSHOSMountTools):
 
         return new_mountpoints
 
+    def _get_mounted_devices(self):
+        mounts = self._exec_cmd(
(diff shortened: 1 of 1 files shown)


coriolis/osmorphing/osmount/base.py
```diff
@@ -233,9 +243,8 @@ def _get_volume_block_devices(self):
             "lsblk -lnao KNAME").decode().split('\n')[:-1]
         LOG.debug("All block devices: %s", str(volume_devs))
 
-        # Skip this instance's current root device (first in the list)
         volume_devs = ["/dev/%s" % d for d in volume_devs if
-                       not re.match(r"^.*\d+$", d)][1:]
```
aznashwan: Great catch, I wasn't even aware we do this here (I presumed there was no filtering which is why I added the `ignore_devices` logic above)

PR #19, opened 2019-02-21
Fix os_dismount()

gabriel-samfira:
In some situations, during os morphing, some binaries that get called
inside the chroot, may mount aditional filesystems, as a child of some
already mounted FS. For example, some applications that require FUSE
may mount aditional kernel filesystems in /sys. This change unmounts
child filesystems before attempting to unmount parent filesystems.

This is done by looking in /proc/mounts first, sorting the mounts that
have the root filesystem as a prefix, and unmounting them in order.

NOTE: currently testing this.


gabriel-samfira:
rebased

PR #21, opened 2019-03-01
Fix chunk size

gabriel-samfira:
Do as the comment says.

This reduces the number of round trips and makes transfers considerably faster.


aznashwan:
Fantastic catch, merging, thanks!

PR #22, opened 2019-03-05
Add context manager for ephemeral Keystone trusts.

aznashwan:
Add a new context manager for creating a Keystone trust for the duration
of an operation and then removing, and use it all the synchronous
calls in the conductor (ex: listing instances etc...)


aznashwan:
NOTE: this is more of a workaround than an actual necessary fix (arguably all of these synchronous calls should just use the caller's token under all circumstances)

aznashwan:
This has been fixed after completely porting Coriolis to using Keystone v3. Closing now.

PR #23, opened 2019-03-18
Add target-env param in `get_storage()` provider abstract method.

aznashwan:


aznashwan:
Merging as it's trivial but necessary.

PR #24, opened 2019-03-22
Add replicator integration

gabriel-samfira:
This PR adds/fixes the following:

  * Adds replicator integration. The replicator is an implementation of a disk differentiating procedure that allows us to have replicas even for providers that do not have a native way of getting incremental diffs of storage devices.
    * Replicator listens on default port 4433.
    * Replicator implements a ssh tunnel fallback in case 4433 is not directly accessible
    * During initial replica, if destination provider zeros out disks before attaching them to instances, only written blocks will be transfered.
    * on subsequent replica runs, any changes (even zero chunks) are synced.
  * fixes a deadlock in coriolis core which shows up when task_info is larger than can be put on a linux named pipe.
  * changes the info field of ```base_transfer_action``` to ```LONGBLOB```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-* text eol=lf
+* text=auto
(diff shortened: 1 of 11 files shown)


aznashwan:
This has been furtively tested so I'll be merging it as-is for now and we can push any needed improvements later.

PR #29, opened 2019-04-26
Add support for interactions with the licensing server. 

aznashwan:


coriolis/licensing/client.py
```diff
@@ -0,0 +1,188 @@
...
+                "No 'LICENSING_SERVER_BASE_URL' env var present. Cannot "
+                "instantiate licensing client.")
+            return None
+        allow_untrusted = False
+        if os.environ.get("LICENSING_SERVER_ALLOW_UNTRUSTED"):
+            allow_untrusted = True
+
+        # try out client:
+        client = cls(base_url, allow_untrusted=allow_untrusted)
+        try:
```
gabriel-samfira: Not much use if you are only going to log if it fails. If something is wrong with the licensing server, it will fail again on next call, which will yield relevant info to the user. Log messages will simply stream by unnoticed until something does wrong.

This would be useful only if we actually send a notification to a monitoring system or an operator, that something is wrong.

coriolis/licensing/client.py
```diff
@@ -0,0 +1,188 @@
...
+                "Error while testing licensing client. Exception:\n%s",
+                utils.get_exception_details())
+
+        return client
+
+    def _get_url_for_resource(self, resource):
+        """ Provides full URL for subresource.
+        Ex: "licences" -> "http://$host:$port/v1/licences"
+        """
+        return "%s/%s" % (self._base_url, resource.strip('/'))
```
gabriel-samfira: you probably want ```lstrip()``` here. Otherwise both leading and trailing slashes will be stripped.

coriolis/licensing/client.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def get_licences(self):
+        """ Lists all installed licences. """
+        return self._get("/licences", response_key="licences")
+
+    def add_licence(self, licence_data):
+        """ Sends request to add licence (in .PEM format). """
+        return self._post("/licences", licence_data)
+
+    def add_reservation(self, typ, num_vms):
```
gabriel-samfira: typ --> type

No use in abbreviating for the sake of one letter.

coriolis/licensing/client.py
```diff
@@ -0,0 +1,188 @@
...
+        return self._get(
+            "/reservations/%s" % reservation_id, response_key="reservation")
+
+    def check_reservation(self, reservation_id):
+        """ Checks the reservation with the given ID.  """
+        return self._post(
+            "/reservations/%s/check" % reservation_id, None,
+            response_key="reservation")
+
+    def delete_reservation(self, reservation_id):
```
gabriel-samfira: It may be useful to have a parameter like ```fail_on_404=False``` here. In some cases it may be desirable to fail if a reservation was expected to be there, but wasn't. Feel free to ignore this one. Just a thought.

coriolis/licensing/client.py
aznashwan: True, but I was trying to avoid having the conductor crash as much as possible, which is why both this check, and the reservation clearing call are wrapped in a try-except.
Admittedly this call is a bit redundant, so I'll just remove it completely.

coriolis/conductor/rpc/server.py
```diff
@@ -86,9 +87,61 @@ def inner():
...
+            except (Exception, KeyboardInterrupt):
+                LOG.warn(
+                    "Failed to delete reservation with ID '%s' for transfer "
+                    "action with ID '%s'. Skipping. Exception\n%s",
+                    reservation_id, action_id, utils.get_exception_details())
+
+    def _check_create_reservation_for_transfer(
+            self, transfer_action, transfer_type):
+        action_id = transfer_action.base_id
+        if not self._licensing_client:
```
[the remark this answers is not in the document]
aznashwan: Agree that it's repetitive but I did not feel it was warranted considering I wanted the log messages to be explicitly different. (it's just an `if` with a single statement in the body so I think making a decorator for it would be more cumbersome)

coriolis/licensing/client.py
aznashwan: It was intentional, as the Go implementation of the server doesn't like trailing slashes, as the Replicator does :))

coriolis/licensing/client.py
aznashwan: It's named that way so as to avoid shadowing the `type` builtin.

gabriel-samfira: then call it ```reservation_type```. Something a bit more meaningful.

coriolis/licensing/client.py
gabriel-samfira: ahh. Makes sense :).

coriolis/licensing/client.py
gabriel-samfira: This call does not prevent the conductor from throwing errors. It just logs if something goes wrong, and nobody will ever notice that log :).

aznashwan:
[Artificially] tested. Merging as-is for now.

PR #31, opened 2019-05-21
Ignore Exceptions while uninstalling packages on SUSE.

aznashwan:
Considering `zypper` does not have an `--ignore-missing`-type flag, and
returns a non-zero exit code when a package it's trying to uninstall is
missing, Coriolis will now silently skip any uninstallation issues on
SLES.


aznashwan:
Tested and works. Merging.

PR #32, opened 2019-05-21
Make compression configurable

gabriel-samfira:
Currently, compression is enabled by default, and there is no way to disable it via config or from the providers calling SSHBackupWriter. This PR adds the ability to do both.

### DIFF
--- a/coriolis/providers/backup_writers.py
+++ b/coriolis/providers/backup_writers.py
@@ -7,6 +7,7 @@ import errno
 import os
 import threading
 
+from oslo_config import cfg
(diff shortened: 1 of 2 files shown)


coriolis/providers/backup_writers.py
```diff
@@ -8,6 +8,7 @@
 import threading
 
 from oslo_log import log as logging
+from oslo_config import cfg
```
aznashwan: Nit: this should come before `oslo_log`,

aznashwan:
Merging. Thanks!

PR #34, opened 2019-06-01
Improve OSMount and OSMorphing processes.

aznashwan:


aznashwan:
Tested and works. Merging.

PR #35, opened 2019-06-01
Error out of no Import OSMorphing tools are found.

aznashwan:


aznashwan:
Tested and works. Merging.

PR #36, opened 2019-06-03
OSMount fixes

aznashwan:


aznashwan:
Tested and works. Merging.

PR #38, opened 2019-07-03
Enable TLS1.2 in download_file

gabriel-samfira:
TLS 1.0 is broken and a lot of servers have disabled it. This sets TLS 1.2 in powershell before calling download.


aznashwan:
LGTM, merging now.

PR #39, opened 2019-07-05
Skip checking reservations on missing `reservation_id`

aznashwan:
In case of Replicas which did not have a Reservation ID created for
them, skip the reservation check.


aznashwan:
Trivial. Merging.

PR #41, opened 2019-07-12
Split BaseReplicaUpdateableProvider class

ionutbalutoiu:
Split the `BaseReplicaUpdateableProvider` class into separate
classes for source and destination clouds.

This fixes a bug when providers were not being properly loaded,
and the wrong `check_update_environment_params` was called.


aznashwan:
LGTM. Merging. Thanks!

PR #42, opened 2019-07-12
Ensure disk ordering is preserved during Replication.

aznashwan:
Coriolis-core will not both validate the schema of the returned 'volumes_info' of each Replica task, as well as reorder the returned volumes into to match the exact ordering found in `export_info[devices][disks]`.
Also added dedicated schemas for the disk sync resources.
Added the following schemas:
- disk_sync_resources_conn_info_schema.json -- for defining SSH connection params for temporary minion VMs
- disk_sync_resources_info_schema.json -- schema defining volumes_info for disk syncing (both for Migrations and Replicas)

Tested import providers which showed signs of potential disk ordering issues:
- [x] Azure
- [x] OCI -- (up to finalize task)
- [ ] OPC

### DIFF
--- a/coriolis/providers/backup_writers.py
+++ b/coriolis/providers/backup_writers.py
@@ -16,14 +16,16 @@ from coriolis import data_transfer
 from coriolis import exception
 from coriolis import utils
 
-LOG = logging.getLogger(__name__)
+
 CONF = cfg.CONF
 opts = [
     cfg.BoolOpt('compress_transfers',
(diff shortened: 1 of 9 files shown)


aznashwan:
Tested import providers which showed signs of potential disk ordering issues:

- [x] Azure
- [ ] OCI
- [ ] OPC

aznashwan:
Tested by the QA team. Merging.

PR #43, opened 2019-07-12
Remove redundant task dependency

ionutbalutoiu:
The replica source inputs validation task does not
need anything from the get instance info task.

Thus, the tasks dependency is redundant.


aznashwan:
LGTM. Merging. Thanks!

PR #45, opened 2019-07-19
Add compressor support

gabriel-samfira:
This adds the option to use an external, multi-threaded service for compression.


coriolis/data_transfer.py
```diff
@@ -16,20 +104,13 @@ def encode_data(msg_id, path, offset, content, compress=True):
 
     data_len_inflated = len(inflated_content)
 
+    compressed = False
     if compress:
-        data_content = zlib.compress(inflated_content)
+        data_content, compressed = compression_proxy(
+            inflated_content, constants.COMPRESSION_FORMAT_ZLIB)
```
aznashwan: NIT: I suggest we make the compression type configurable just in case

gabriel-samfira: the ```write_data``` binary expects ```zlib``` compression format. So not sure if it's worth making ```encode_data()``` configurable.

gabriel-samfira: ```compression_proxy()``` does accept a compression format type flag.

aznashwan:
LGTM. Merging. Thanks!

PR #46, opened 2019-07-19
Add system startup creation functions

gabriel-samfira:
This change allows the creation of system start-up scripts for upstart and systemd on remote temporary workers.

### DIFF
--- a/coriolis/utils.py
+++ b/coriolis/utils.py
@@ -1,5 +1,6 @@
 # Copyright 2016 Cloudbase Solutions Srl
 # All Rights Reserved.
+# pylint: disable=anomalous-backslash-in-string
 
 import base64
 import binascii
@@ -15,6 +16,7 @@ import socket
 import subprocess
 import time
 import traceback
+import uuid
 
 import OpenSSL
 from oslo_config import cfg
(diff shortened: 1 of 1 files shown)


coriolis/utils.py
```diff
@@ -498,3 +531,78 @@ def filter_chunking_info_for_task(task_info):
             vol["replica_state"]["chunks"] = "<redacted>"
 
     return cpy
+
+
+def _write_systemd(ssh, cmdline, svcname, run_as=None, start=True):
+    serviceFilePath = "/lib/systemd/system/%s.service" % svcname
+
+    if test_ssh_path(ssh, serviceFilePath):
+        return
```
aznashwan: IMO we should try to backup the existing file and create a new one in case there are any differences.

coriolis/utils.py
```diff
@@ -498,3 +531,78 @@ def filter_chunking_info_for_task(task_info):
...
+        "sudo mv /tmp/%s.service %s" % (name, serviceFilePath),
+        get_pty=True)
+    _reload_and_start(start=start)
+
+
+def _write_upstart(ssh, cmdline, svcname, run_as=None, start=True):
+    serviceFilePath = "/etc/init/%s.conf" % svcname
+
+    if test_ssh_path(ssh, serviceFilePath):
+        return
```
aznashwan: Same file backup & re-write observation as above.

coriolis/utils.py
gabriel-samfira: if we do that, we replace the file every time we call this function, whether it's needed or not, and reload systemd && restart application (which you will probably not want to do). If we need to replace a service, I propose we do it explicitly by removing the old config file and writing it again.

aznashwan:
Merging, thanks!

PR #47, opened 2019-07-19
Add http backup writer

gabriel-samfira:
Leverage a HTTP based backup writer. The HTTP backup writer has the added benefit of requiring less resources for large chunks, as it does not need to rely on struct.pack to generate the payload it sends over the wire. Standard HTTP calls are made to a remote API.

 It also does not require a live, SSH channel to be active, and recovery from failure is more likely to succeed. Removing the reliance on paramiko also means that we are no longer constrained by the fact that encryption of data over SSH done by paramiko is single threaded and limited by the power of a single core available to the Coriolis  worker.

It's worth noting that this also implements queueing for the phases of transfer of data from source to destination. There is a ```compression``` queue, and a ```sender``` queue. That means that while we are compressing, we can still read from source, and as we are sending to destination, we can still compress and read from source.

Unrecoverable errors are propagated from the eventlets that are processing the compression/sending and will surface on next call to ```write()``` or ```close()```.

Depends on #45  #46

### DIFF
--- a/coriolis/providers/backup_writers.py
+++ b/coriolis/providers/backup_writers.py
@@ -20,8 +20,8 @@ LOG = logging.getLogger(__name__)
 CONF = cfg.CONF
 opts = [
     cfg.BoolOpt('compress_transfers',
-               default=True,
-               help='Use compression if possible during disk transfers'),
+                default=True,
+                help='Use compression if possible during disk transfers'),
 ]
 CONF.register_opts(opts)
 
--- a/coriolis/providers/backup_writers.py
+++ b/coriolis/providers/backup_writers.py
@@ -260,7 +260,8 @@ class SSHBackupWriter(BaseBackupWriter):
 
 
 class HTTPBackupWriterImpl(BaseBackupWriterImpl):
-    def __init__(self, path, disk_id, compress_transfer=None):
+    def __init__(self, path, disk_id,
+                 compress_transfer=None, compressor_count=3):
         self._offset = None
         self._session = None
         self._ip = None
@@ -272,6 +273,7 @@ class HTTPBackupWriterImpl(BaseBackupWriterImpl):
         self._write_error = False
         self._id = None
         self._exception = None
+        self._compressor_count = compressor_count
         self._comp_q = eventlet.Queue(maxsize=5)
         self._sender_q = eventlet.Queue(maxsize=5)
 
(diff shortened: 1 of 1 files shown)


coriolis/providers/backup_writers.py
```diff
@@ -5,13 +5,19 @@
 import contextlib
 import errno
 import os
+import tempfile
+import time
```
aznashwan: NIT: `time` should be imported after `threading`

coriolis/providers/backup_writers.py
```diff
@@ -250,3 +257,408 @@ def _connect_ssh(self):
...
+        self._id = info.get("id")
+        if not all([self._ip, self._port, self._crt,
+                    self._key, self._ca, self._id]):
+            raise exception.CoriolisException(
+                "Missing required info when creating HTTPBackupWriter")
+
+    @property
+    def _uri(self):
+        return "https://%s:%s/api/v1%s" % (
+            self._ip, self._port, self._path
```
aznashwan: Safety nit: use `"...v1/%s" % self._path.strip('/')` in case the `path` provided in the constructor doesn't have a leading slash (which is bad on the callers part but better safe than sorry)

coriolis/providers/backup_writers.py
```diff
@@ -250,3 +257,408 @@ def _connect_ssh(self):
...
+        sess.verify = self._ca
+        self._session = sess
+
+    def _open(self):
+        self._closing = False
+        self._init_session()
+        self._acquire()
+        self._sender_evt = eventlet.spawn(self._sender)
+        self._compressor_evt = [
+            eventlet.spawn(self._compressor),
```
aznashwan: NIT: we should make the compressor thread count configurable if possible

coriolis/providers/backup_writers.py
```diff
@@ -250,3 +257,408 @@ def _connect_ssh(self):
...
+
+    def _setup_certificates(self, ssh):
+        remote_base_dir = "/etc/coriolis-writer"
+
+        ca_crt_name = "ca-cert.pem"
+        client_crt_name = "client-cert.pem"
+        client_key_name = "client-key.pem"
+
+        srv_crt_name = "srv-cert.pem"
+        srv_key_name = "srv-key.pem"
```
aznashwan: Considering these constants are also used in the `replicator.py`, it might be worth defining separate constants for them just in case.

coriolis/providers/backup_writers.py
gabriel-samfira: indeed!

coriolis/providers/backup_writers.py
gabriel-samfira: indeed. There is a TODO in the replicator to make a common way to do this. I would loke for coriolis to have some sort of PKI implemented instead. As in have a CA in a coriolis component that signs the certificates generated for the services we deploy.

aznashwan:
LGTM. Merging now.

PR #48, opened 2019-07-20
Cleanup replicator integration

gabriel-samfira:
This makes use of utils.create_service() and refactors a bit the way certificates are copied over from the replicator.

Depends on #46

### DIFF
--- a/coriolis/providers/replicator.py
+++ b/coriolis/providers/replicator.py
@@ -7,6 +7,7 @@ import errno
 import os
 import time
 import requests
+import uuid
 
 from coriolis import utils
 from coriolis import exception
@@ -23,8 +24,11 @@ HASH_METHOD_SHA256 = "sha256"
 HASH_METHOD_XXHASH = "xxhash"
 
 REPLICATOR_PATH = "/usr/bin/replicator"
+REPLICATOR_DIR = "/etc/coriolis-replicator"
 REPLICATOR_STATE = "/tmp/replicator_state.json"
 REPLICATOR_USERNAME = "replicator"
+REPLICATOR_SVC_NAME = "coriolis-replicator"
+
 DEFAULT_REPLICATOR_PORT = 4433
 
 replicator_opts = [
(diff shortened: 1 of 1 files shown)


coriolis/providers/replicator.py
```diff
@@ -429,163 +402,133 @@ def _setup_replicator_user(self, ssh):
...
-        """
-        if self._cert_dir is None:
-            self._cert_dir = tempfile.mkdtemp()
+    def _setup_certificates(self, ssh, args):
+        # TODO(gsamfira): coriolis-replicator and coriolis-writer share
+        # the functionality of being able to generate certificates
+        # This will either be replaced with proper certificate management
+        # in Coriolis, and the needed files will be pushed to the services
+        # that need them (userdata or ssh), or the two applications will be
+        # merged into one, and we will deduplicate thif functionallity.
```
ader1990: thif -> this

coriolis/providers/replicator.py
```diff
@@ -390,12 +362,9 @@ def _parse_replicator_conn_info(self, conn_info):
...
         }
 
+    @utils.retry_on_error()
     def _copy_file(self, ssh, localPath, remotePath):
-        tmp = tempfile.mkstemp()[1]
-        try:
-            os.remove(tmp)
-        except BaseException:
-            pass
+        tmp = os.path.join("/tmp", str(uuid.uuid4()))
```
aznashwan: Any reason why we'd not want to use `tempfile.mkstemp` given that the dir is on the local machine?

coriolis/providers/replicator.py
```diff
@@ -429,163 +402,133 @@ def _setup_replicator_user(self, ssh):
...
         }
 
+    @utils.retry_on_error()
+    def _setup_replicator(self, ssh):
+        # copy the binary, set up the service, generate certificates,
+        # start service
+        state_file = self._get_replicator_state_file()
+        self._copy_file(ssh, state_file, REPLICATOR_STATE)
+        utils.exec_ssh_cmd(
+            ssh, "sudo chmod 755 %s" % REPLICATOR_STATE, get_pty=True)
```
aznashwan: If adding `get_pty=True` to `exec_ssh_cmd` is so common, maybe we should consider setting `True` as the default value for the kwarg.

coriolis/providers/replicator.py
gabriel-samfira: The file is remote. tempfile creates a local file. No use in creating a local file just to get a name we could use remotely.

coriolis/providers/replicator.py
gabriel-samfira: get_pty is only needed if we run ```sudo``` on systems where it was explicitly configured to need a real ```pty```. Changing this default is outside the scope of this PR though.

coriolis/providers/replicator.py
```diff
@@ -7,6 +7,7 @@
 import os
 import time
 import requests
+import uuid
```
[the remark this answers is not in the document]
gabriel-samfira: Done

coriolis/providers/replicator.py
aznashwan: My bad, I fleetingly saw the separate `remotePath` and assumed this one would be a local one.

aznashwan:
LGTM. Merging. Thanks!

PR #49, opened 2019-07-23
Fix _expand_archive()

gabriel-samfira:
In cases where System.IO.Compression.ZipFile is available on the
windows minion, if the destination folder exists, it will cause
an error.

Where this component is not available, if the target folder exists
and there are conflicting files inside it with what's inside the
archive, it will error out.

In both cases, ensuring that the destination folder does not exist
will fix the issue.

If the destination folder is the root folder or is within the
%WINDOWS% folder, we do not remove it.

### DIFF
--- a/coriolis/osmorphing/windows.py
+++ b/coriolis/osmorphing/windows.py
@@ -9,6 +9,7 @@ import uuid
 from oslo_log import log as logging
 
 from coriolis import exception
+from coriolis import utils
 from coriolis.osmorphing import base
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


coriolis/osmorphing/windows.py
```diff
@@ -146,9 +147,21 @@ def _dismount_disk_image(self, path):
...
+    def _expand_archive(self, path, destination, overwrite=True):
         LOG.info("Expanding archive \"%(path)s\" in \"%(destination)s\"",
                  {"path": path, "destination": destination})
+
+        if self._conn.test_path(destination):
+            LOG.info("Destination folder %s already exists" % destination)
+            if overwrite:
+                if destination.endswith(":\\") or ":\\Windows" in destination:
+                    LOG.warn(
+                        "Not removing target directory, as it is either the "
```
aznashwan: Do we run the risk of it asking for overwrite confirmation and hanging if we ignore this here?
If so, I suggest we just error out and impose that the plugins not directly unarchive shit in C:\

gabriel-samfira: the expand archive commands are non-interactive.

aznashwan: Cool. LGTM then.

aznashwan:
LGTM. Merging.

PR #50, opened 2019-07-31
Refactor network morphing

gabriel-samfira:
This change does the following:

For situations where replicas are created with ```set_dhcp=True```:

  * Disable predictable NIC names
  * Create DHCP enabled configs for all NICs
  * Make backups of any mutated config files

When migrating from one cloud to another and from one hypervisor to another, there is no way to guarantee NIC names will stay the same on the destination. The only condition where we can ensure NIC names, is if the cloud allows us to preserve both MAC addresses and IP addresses when migrating from source.

It also means the source allows us to introspect the instance using the cloud APIs and fetch the required info. Given that we currently only have a small nr of combinations where this would work, flattening the morphing procedure for networking is the only sane approach, given our agent-less architecture.

### DIFF
--- a/coriolis/osmorphing/debian.py
+++ b/coriolis/osmorphing/debian.py
@@ -2,9 +2,9 @@
 # All Rights Reserved.
 
 import os
-import yaml
-
 from io import StringIO
+
+import yaml
 
 from coriolis import utils
 from coriolis.osmorphing import base
--- a/coriolis/osmorphing/debian.py
+++ b/coriolis/osmorphing/debian.py
@@ -44,13 +44,13 @@ class BaseDebianMorphingTools(base.BaseLinuxOSMorphingTools):
         cfg = utils.Grub2ConfigEditor(contents)
         cfg.append_to_option(
             "GRUB_CMDLINE_LINUX_DEFAULT",
-            {"opt_type": "key_val", "opt_key": "net.ifaces", "opt_val": 0})
+            {"opt_type": "key_val", "opt_key": "net.ifnames", "opt_val": 0})
         cfg.append_to_option(
             "GRUB_CMDLINE_LINUX_DEFAULT",
             {"opt_type": "key_val", "opt_key": "biosdevname", "opt_val": 0})
         cfg.append_to_option(
             "GRUB_CMDLINE_LINUX",
-            {"opt_type": "key_val", "opt_key": "net.ifaces", "opt_val": 0})
+            {"opt_type": "key_val", "opt_key": "net.ifnames", "opt_val": 0})
         cfg.append_to_option(
             "GRUB_CMDLINE_LINUX",
             {"opt_type": "key_val", "opt_key": "biosdevname", "opt_val": 0})
--- a/coriolis/osmorphing/debian.py
+++ b/coriolis/osmorphing/debian.py
@@ -3,6 +3,7 @@
 
 import os
 
+from coriolis import utils
 from coriolis.osmorphing import base
 
 
@@ -20,6 +21,29 @@ class BaseDebianMorphingTools(base.BaseLinuxOSMorphingTools):
             release = self._read_file(
                 debian_version_path).decode().split('\n')[0]
             return ('Debian', release)
+
+    def disable_predictable_nic_names(self):
+        grub_cfg = os.path.join(
+            self._os_root_dir,
+            "etc/default/grub")
+        if self._test_path(grub_cfg) is False:
+            return
+        contents = self._read_file(grub_cfg).decode()
+        cfg = utils.Grub2ConfigEditor(contents)
+        cfg.append_to_option(
+            "GRUB_CMDLINE_LINUX_DEFAULT",
+            {"opt_type": "key_val", "opt_key": "net.ifaces", "opt_val": 0})
+        cfg.append_to_option(
+            "GRUB_CMDLINE_LINUX_DEFAULT",
+            {"opt_type": "key_val", "opt_key": "biosdevname", "opt_val": 0})
+        cfg.append_to_option(
+            "GRUB_CMDLINE_LINUX",
+            {"opt_type": "key_val", "opt_key": "net.ifaces", "opt_val": 0})
+        cfg.append_to_option(
+            "GRUB_CMDLINE_LINUX",
+            {"opt_type": "key_val", "opt_key": "biosdevname", "opt_val": 0})
+        self._write_file_sudo(grub_cfg, cfg.dump())
+        self._exec_cmd_chroot("/usr/sbin/update-grub")
 
     def set_net_config(self, nics_info, dhcp):
         if dhcp:
--- a/coriolis/osmorphing/debian.py
+++ b/coriolis/osmorphing/debian.py
@@ -97,7 +97,6 @@ class BaseDebianMorphingTools(base.BaseLinuxOSMorphingTools):
             return
 
         self.disable_predictable_nic_names()
-        # NOTE: doesn't work with chroot
         if self._test_path("etc/network"):
             ifaces_file = "etc/network/interfaces"
             contents = self._compose_interfaces_config(nics_info)
--- a/coriolis/osmorphing/debian.py
+++ b/coriolis/osmorphing/debian.py
@@ -2,10 +2,22 @@
 # All Rights Reserved.
 
 import os
+import yaml
+
+from io import StringIO
 
 from coriolis import utils
 from coriolis.osmorphing import base
 
+LO_NIC_TPL = """
+auto lo
+iface lo inet loopback
+"""
+
+INTERFACES_NIC_TPL = """
+auto %(device_name)s
+iface %(device_name)s inet dhcp
+"""
 
 class BaseDebianMorphingTools(base.BaseLinuxOSMorphingTools):
     def _check_os(self):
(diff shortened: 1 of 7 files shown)


coriolis/osmorphing/debian.py
```diff
@@ -2,9 +2,22 @@
 # All Rights Reserved.
 
 import os
+import yaml
```
aznashwan: NIT: `io` import should be grouped with the other builtins above (i.e. this newline is extraneous)

coriolis/utils.py
```diff
@@ -606,3 +609,198 @@ def create_service(ssh, cmdline, svcname, run_as=None, start=True):
...
         raise exception.CoriolisException(
             "could not determine init system")
+        
+        
+class Grub2ConfigEditor(object):
+    """This class edits GRUB2 configs, normally found in
+    /etc/default/grub. This class tries to preserve commented
+    and empty lines.
+    NOTE: This class does not actually write to file during
+    commit. Rhather, it will mutate it's internal view iof the
```
aznashwan: s/iof/of

coriolis/utils.py
```diff
@@ -606,3 +609,198 @@ def create_service(ssh, cmdline, svcname, run_as=None, start=True):
...
+    contents of that file with the latest changes made.
+    Use dump() to get the file contents.
+    """
+    def __init__(self, cfg):
+        self._cfg = cfg
+        self._parsed = self._parse_cfg(self._cfg)
+
+    def _parse_cfg(self, cfg):
+        ret = []
+        for line in cfg.split("\n")[:-1]:
```
aznashwan: NIT: can we add a `if not line: continue` at the top to not run redundant checks in case of multiple newlines

coriolis/utils.py
```diff
@@ -606,3 +609,198 @@ def create_service(ssh, cmdline, svcname, run_as=None, start=True):
...
+        """
+        self._validate_value(value)
+        opt_found = False
+        for opt in self._parsed:
+            if opt.get("option_name") == option:
+                opt_found = True
+                opt["option_value"] = value
+                break
+        if not opt_found:
+            self._parsed.append({
```
aznashwan: Side-note: this will need to extended later in case we have .ini files with separate config `[sections]` (in that case just appending the opt at the end might put it in the wrong config section)

coriolis/utils.py
```diff
@@ -606,3 +609,198 @@ def create_service(ssh, cmdline, svcname, run_as=None, start=True):
...
+                found = False
+                for val in opt["option_value"]:
+                    if (val["opt_type"] == "key_val" and 
+                            value["opt_type"] == "key_val"):
+                        if str(val["opt_key"]) == str(value["opt_key"]):
+                            val["opt_val"] = value["opt_val"]
+                            found = True
+                    elif (val["opt_type"] == "single" and 
+                            value["opt_type"] == "single"):
+                        if str(val["opt_val"]) == str(value["opt_val"]):
```
aznashwan: As far as I can tell if we do find an option of type `single` which is already set to something `!= new_value['opt_val']` then we don't set the new value at all?
This is fine as long as it's intended behavior, but I'd suggest we at the very least `LOG.warn()` about it.

coriolis/utils.py
```diff
@@ -606,3 +609,198 @@ def create_service(ssh, cmdline, svcname, run_as=None, start=True):
...
+                if val["opt_type"] == "key_val":
+                    flat.append("%s=%s" % (val["opt_key"], val["opt_val"]))
+                else:
+                    flat.append(str(val["opt_val"]))
+
+            if len(flat) == 0:
+                tmp.write("%s=\n" % line["option_name"])
+                continue
+
+            val = " ".join(flat)
```
aznashwan: NIT: I realize this is application-specific to GRUB config files but IMO we should make the separator for the KV pairs configurable (maybe some weird distro uses commas or whatever instead of a space as a separator)

coriolis/utils.py
```diff
@@ -606,3 +609,198 @@ def create_service(ssh, cmdline, svcname, run_as=None, start=True):
...
+            quoted = line["quoted"]
```
aznashwan: Side-NIT: although I don't expect there are any kernel args with spaces in them, this would also screw up the formatting unless we encase the value in quotes (but this depends on GRUB's parser and IDK for sure)

coriolis/osmorphing/debian.py
```diff
@@ -21,13 +34,89 @@ def _check_os(self):
...
+            "GRUB_CMDLINE_LINUX_DEFAULT",
+            {"opt_type": "key_val", "opt_key": "biosdevname", "opt_val": 0})
+        cfg.append_to_option(
+            "GRUB_CMDLINE_LINUX",
+            {"opt_type": "key_val", "opt_key": "net.ifnames", "opt_val": 0})
+        cfg.append_to_option(
+            "GRUB_CMDLINE_LINUX",
+            {"opt_type": "key_val", "opt_key": "biosdevname", "opt_val": 0})
+        self._write_file_sudo("etc/default/grub", cfg.dump())
+        self._exec_cmd_chroot("/usr/sbin/update-grub")
```
aznashwan: Apart from this last `update-grub` call I believe all of the above code also applies for RHEL-based distros so we could abstract it (or the `update-grub` call) away in a method in the base/child classes.
LE: nvm I wasn't aware of the `new-kernel-pkg` binary on RHELs

coriolis/osmorphing/redhat.py
```diff
@@ -105,12 +127,27 @@ def _get_ifcfgs_by_type(self, ifcfg_type):
...
+        for idx,_ in enumerate(nics_info):
+            dev_name = "eth%d" % idx
+            cfg_path = "etc/sysconfig/network-scripts/ifcfg-%s" % dev_name
+            if self._test_path(cfg_path):
+                self._exec_cmd_chroot(
+                    "cp %s %s.bak" % (cfg_path, cfg_path)
+                )
+            self._write_file_sudo(
+                cfg_path,
+                IFCFG_TEMPLATE % {
```
aznashwan: Couldn't we also add the `HWADDR` (MAC address) option here (granted we know it and the dest provider supports setting it on NICs ofc)

gabriel-samfira: the point of this change is to create a valid config when we don't know the MAC or are unable to retain the NIC name because we cannot retain both MAC and IP. The ordering of eth0,1,2,3 is not guaranteed across reboots, hence, the flat, generic configs that should work across reboots even if eth0 decides to become eth1 after a reboot.

Also, I think that if we want to create a generic approach to something, we need to leave out things that are particular to one component or another, and be as generic as possible without affecting functionality. In this instance, all NICs are set to DHCP, regardless of their MAC.

coriolis/utils.py
gabriel-samfira: I wanted to retain the structure of the file as much as possible. That means comments and newlines, as the maintainers of that system intended that file to be. That is the purpose of the ```raw``` type.  The result is that the only thing that changes, are the values of the options you are mutating. The rest of the file remains the same (spaces, typos, comments, etc).

coriolis/utils.py
gabriel-samfira: this was written specifically for the grub config file. It is not meant to parse ```ini``` files. There is the ```ConfigObj``` module for that. No use in reinventing the wheel. The grub config file is essentially a source-able bash env file, with certain rules.

coriolis/utils.py
gabriel-samfira: Single options are non key/value pair options. For example, a ```key_val``` type is of the form:

```bash
net.ifaces=1
``` 

A single is a simple scalar:
```bash
noapic
noacpi
no_stf_barrier
```

The ```append_to_option()``` function will add an element to an OPTION. An option is one of:

```bash
GRUB_HIDDEN_TIMEOUT
GRUB_CMDLINE_LINUX_DEFAULT
``` 

and so on. So a ```single``` option means adding something like ```noapic``` to one of the grub options.

Now, if a ```single``` option already exists in the grub options you are editing, it will be ignored. If not, it will be **appended**.

So for example, given the option:

```bash
# this already has 2 single flags set
GRUB_CMDLINE_LINUX="noibrs noibpb"
```

calling:

```python
cfg.append_to_option("GRUB_CMDLINE_LINUX", {"opt_type": "single", "opt_val": "noibrs"})
cfg.append_to_option("GRUB_CMDLINE_LINUX", {"opt_type": "single", "opt_val": "noacpi"})
```

will result in:

```bash
GRUB_CMDLINE_LINUX="noibrs noibpb noacpi"
```

because noibrs already existed.

coriolis/utils.py
gabriel-samfira: let's cross that bridge when we come to it. I have yet to see such things in the wild. This is standard grub config file, and the kernel accepts space delimited parameters to be passed to it.

coriolis/utils.py
gabriel-samfira: The scripts that parse the config option and set the grub config with appropriate values, make use of the quotes. It is actually a bash source-able file that gets included. So the quotes are necessary:

```bash
gabriel@rossak:~/$ BLA="with quotes"
gabriel@rossak:~/$ BLA=no quotes
quotes: command not found

```

coriolis/utils.py
aznashwan: Thanks for the explanation, my bad for not getting it the first time.

aznashwan:
All "concerns" addressed, merging now.

PR #52, opened 2019-08-01
Add timeout to retries when connecting

gabriel-samfira:
When connecting to SSH, we need to set a timeout in order to avoid
saturating OpenSSH server's connection pool. This can happen if SSH
starts listening to connections and the guest agent did not yet manage
to write the public keys for the user we are authenticating as.

### DIFF
--- a/coriolis/providers/backup_writers.py
+++ b/coriolis/providers/backup_writers.py
@@ -252,16 +252,21 @@ class SSHBackupWriter(BaseBackupWriter):
             finally:
                 sftp.close()
 
-    @utils.retry_on_error()
+    @utils.retry_on_error(sleep_seconds=30)
     def _connect_ssh(self):
         LOG.info("Connecting to SSH host: %(ip)s:%(port)s" %
                  {"ip": self._ip, "port": self._port})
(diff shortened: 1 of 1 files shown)


coriolis/providers/backup_writers.py
```diff
@@ -252,16 +252,21 @@ def _copy_helper_cmd(self, ssh):
             finally:
                 sftp.close()
 
-    @utils.retry_on_error()
+    @utils.retry_on_error(sleep_seconds=30)
```
aznashwan: The arg is actually used in the method.
I see there's an `auth_timeout` arg to Client.connect(), but I presume it's for timeout during the initial connection and won't help us here.
http://docs.paramiko.org/en/2.6/api/client.html

gabriel-samfira: of no use. Auth timeout is for situations where client waits for the server to authenticate it, and takes longer than usual. If authentication fails due to bad credentials (ssh key not yet inserted into user's home), it fails instantly.

aznashwan:
LGTM. Merging. Thanks!

PR #53, opened 2019-08-05
Add 'env' option during get_instance(s) API calls

aznashwan:
Providers to be updated:
- [x]  OpenStack
- [x]  VMWare
- [x]  AWS
- [x]  Hyper-V
- [x]  OPC
- [x]  Oracle VM

### DIFF
--- a/coriolis/providers/base.py
+++ b/coriolis/providers/base.py
@@ -34,13 +34,15 @@ class BaseEndpointInstancesProvider(BaseEndpointProvider):
     """Defines operations for listing instances off of Endpoints."""
 
     @abc.abstractmethod
-    def get_instances(self, ctxt, connection_info, limit=None,
-                      last_seen_id=None, instance_name_pattern=None):
+    def get_instances(self, ctxt, connection_info, source_environment,
+                      limit=None, last_seen_id=None,
+                      instance_name_pattern=None):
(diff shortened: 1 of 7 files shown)


coriolis/providers/base.py
```diff
@@ -34,13 +34,15 @@ class BaseEndpointInstancesProvider(BaseEndpointProvider):
     """Defines operations for listing instances off of Endpoints."""
 
     @abc.abstractmethod
-    def get_instances(self, ctxt, connection_info, limit=None,
-                      last_seen_id=None, instance_name_pattern=None):
+    def get_instances(self, ctxt, connection_info, source_environment,
```
gabriel-samfira: This will require that all providers be updated. Is source_environment necessary for all providers?

aznashwan: Not all, but some clouds which have various non-connection-related params (ex: resgroup on Azure or AZ on Amazon), and IMO it's better to make it part of the interface for all plugins IMO moving forward.
Additionally, this will allow for the updating of other plugins which already implement these various params as part of the endpoint connection info (ex: the various region params in OpenStack endpoints)

aznashwan:
Tested and works for all the export plugins (listed above)
Merging.

PR #54, opened 2019-08-08
Adds 'force' query parameter to the replica update API.

Dany9966:
When this parameter is set to true, errors concerning replica source
environment updating will be ignored.

### DIFF
--- a/coriolis/api/common.py
+++ b/coriolis/api/common.py
@@ -10,3 +10,8 @@ def get_paging_params(req):
     if limit is not None:
         limit = utils.parse_int_value(limit)
     return marker, limit
+
+
+def get_force_param(req):
+    force_param = req.GET.get('force', '')
+    return (str(force_param).lower() == "true")
--- a/coriolis/api/v1/replicas.py
+++ b/coriolis/api/v1/replicas.py
@@ -5,10 +5,11 @@ from oslo_log import log as logging
 from webob import exc
 
 from coriolis import exception
+from coriolis.api import common
+from coriolis.api import wsgi as api_wsgi
 from coriolis.api.v1 import utils as api_utils
 from coriolis.api.v1.views import replica_view
 from coriolis.api.v1.views import replica_tasks_execution_view
-from coriolis.api import wsgi as api_wsgi
 from coriolis.endpoints import api as endpoints_api
 from coriolis.policies import replicas as replica_policies
 from coriolis.replicas import api
@@ -260,10 +261,11 @@ class ReplicaController(api_wsgi.Controller):
                             "updated")
 
         updated_values = self._validate_update_body(id, context, body)
+        force = common.get_force_param(req)
         try:
             return replica_tasks_execution_view.single(
                 req, self._replica_api.update(req.environ['coriolis.context'],
-                                              id, updated_values))
+                                              id, updated_values, force))
         except exception.NotFound as ex:
             raise exc.HTTPNotFound(explanation=ex.msg)
         except exception.InvalidParameterValue as ex:
--- a/coriolis/conductor/rpc/server.py
+++ b/coriolis/conductor/rpc/server.py
@@ -308,7 +308,8 @@ class ConductorServerEndpoint(object):
                     task_info=task_info.get(task.instance, {}))
 
     @replica_synchronized
-    def execute_replica_tasks(self, ctxt, replica_id, shutdown_instances):
+    def execute_replica_tasks(
+            self, ctxt, replica_id, shutdown_instances, force):
         replica = self._get_replica(ctxt, replica_id)
         self._check_reservation_for_transfer(replica)
         self._check_replica_running_executions(ctxt, replica)
@@ -319,6 +320,15 @@ class ConductorServerEndpoint(object):
         execution.type = constants.EXECUTION_TYPE_REPLICA_EXECUTION
 
         for instance in execution.action.instances:
+            # After a forced replica update, 'force' parameter will be true in
+            # the replica info, therefore it will linger on to the following
+            # replica executions' task_info. For now, we should reset it for
+            # any other non-update replica executions.
+            if hasattr(replica, 'info'):
+                instance_info = replica.info.get(instance, None)
+                if instance_info:
+                    instance_info['force'] = force
+
             get_instance_info_task = self._create_task(
                 instance, constants.TASK_TYPE_GET_INSTANCE_INFO,
                 execution)
@@ -433,7 +443,7 @@ class ConductorServerEndpoint(object):
         db_api.delete_replica(ctxt, replica_id)
 
     @replica_synchronized
-    def delete_replica_disks(self, ctxt, replica_id):
+    def delete_replica_disks(self, ctxt, replica_id, force):
         replica = self._get_replica(ctxt, replica_id)
         self._check_replica_running_executions(ctxt, replica)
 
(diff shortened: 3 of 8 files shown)


coriolis/api/v1/replicas.py
```diff
@@ -8,6 +8,7 @@
 from coriolis.api.v1 import utils as api_utils
 from coriolis.api.v1.views import replica_view
 from coriolis.api.v1.views import replica_tasks_execution_view
+from coriolis.api import common
```
aznashwan: Imports not properly ordered.

coriolis/tasks/replica_tasks.py
```diff
@@ -554,23 +555,34 @@ def run(self, ctxt, instance, origin, destination, task_info,
...
+                source_provider.check_update_source_environment_params(
+                    ctxt, origin_connection_info, instance, volumes_info,
+                    old_source_env, new_source_env))
+            schemas.validate_value(
+                volumes_info, schemas.CORIOLIS_VOLUMES_INFO_SCHEMA)
+
+            task_info['volumes_info'] = volumes_info
+        except Exception as ex:
+            if task_info['force']:
+                LOG.warn('Forced Replica Update is executed, ignoring: %s' % (
```
aznashwan: Please `event_manager.progress_update()` the message so the user knows what went down.

aznashwan: And also, please add a separate `LOG.trace(ex)` so the stacktrace gets logged.

coriolis/tasks/replica_tasks.py
```diff
@@ -554,23 +555,36 @@ def run(self, ctxt, instance, origin, destination, task_info,
...
+                    ctxt, origin_connection_info, instance, volumes_info,
+                    old_source_env, new_source_env))
+            schemas.validate_value(
+                volumes_info, schemas.CORIOLIS_VOLUMES_INFO_SCHEMA)
+
+            task_info['volumes_info'] = volumes_info
+        except Exception as ex:
+            if task_info['force']:
+                event_manager.progress_update(
+                    'Forced Replica Update is executed, ignoring: %s' % (
```
aznashwan: Please:
1) manually prepend `WARNING` to the error string
2) make the message slightly clearer like: `An error occurred whilst updating the Replica's source parameters. Ignoring all source errors and forcefully updating the state of the Replica. This may have unintended consequences on future Replica executions.`

coriolis/tasks/replica_tasks.py
```diff
@@ -554,23 +555,36 @@ def run(self, ctxt, instance, origin, destination, task_info,
...
+                        str(ex)))
+                LOG.trace(utils.get_exception_details())
+            else:
+                raise exception.CoriolisException(
```
aznashwan: 1) Please just `LOG.warn` this message and re-raise the initial exception to keep its trace.
2) In the future, never format `utils.get_exception_details()` into exception messages themselves as it will lead to the whole ugly-ass trace being exposed through the API.

coriolis/tasks/replica_tasks.py
```diff
@@ -554,23 +555,39 @@ def run(self, ctxt, instance, origin, destination, task_info,
...
+                    "source parameters. Ignoring all source errors and "
+                    "forcefully updating the state of the Replica. This may "
+                    "have unintended consequences on future Replica "
+                    "executions.")
+                LOG.warn(utils.get_exception_details())
+            else:
+                LOG.warn(
+                    "An error was detected while updating source environment. "
+                    "The error can be ignored by force updating the replica.")
+                raise exception.CoriolisException(str(ex))
```
aznashwan: Please just ra-raise the existing Replica here.

coriolis/tasks/replica_tasks.py
```diff
@@ -554,23 +555,39 @@ def run(self, ctxt, instance, origin, destination, task_info,
...
+                    "WARNING: An error occurred whilst updating the Replica's "
+                    "source parameters. Ignoring all source errors and "
+                    "forcefully updating the state of the Replica. This may "
+                    "have unintended consequences on future Replica "
+                    "executions.")
+                LOG.warn(utils.get_exception_details())
+            else:
+                LOG.warn(
+                    "An error was detected while updating source environment. "
+                    "The error can be ignored by force updating the replica.")
```
aznashwan: This is pretty redundant considering 99% of users would/should never see the logs.
I think just re-raising the original exception is enough.

coriolis/conductor/rpc/server.py
```diff
@@ -1137,6 +1137,7 @@ def update_replica(
             # so we must operate on a copy:
             inst_info_copy = copy.deepcopy(replica.info[instance])
             inst_info_copy.update(properties)
+            inst_info_copy['force'] = force
```
aznashwan: This has the pretty big unintended side-effect of that 'force' param staying in the Replica's info until further executions.
Indeed it's overridden during Replica updates each time, but maybe some day normal Replica executions will also 'heed' the 'force' param in the task info.
I have no good solution in mind except explicitly setting 'force=False' (at least for the params) before each non-update (which is bad), but for now we should at the very least add a comment saying that force=X will linger in the task_info after the update.

coriolis/tasks/replica_tasks.py
aznashwan: You could simply "rephrase" that whole logic in the `except` body like:
```python
if not task_info['force']:
    raise
<the other steps>
```

coriolis/api/common.py
```diff
@@ -10,3 +10,8 @@ def get_paging_params(req):
     if limit is not None:
         limit = utils.parse_int_value(limit)
     return marker, limit
+
+
+def get_force_param(req):
+    force_param = req.GET.get('force', '')
+    return (force_param.lower() == "true")
```
aznashwan: How sure are we that the `webob` lib we use will always returns strings?
Please add a `str` cast for safety.

coriolis/conductor/rpc/server.py
```diff
@@ -308,7 +308,8 @@ def _begin_tasks(self, ctxt, execution, task_info={}):
                     task_info=task_info.get(task.instance, {}))
 
     @replica_synchronized
-    def execute_replica_tasks(self, ctxt, replica_id, shutdown_instances):
+    def execute_replica_tasks(
+            self, ctxt, replica_id, shutdown_instances, force=False):
```
aznashwan: As mentioned, "middle-tier" code should not make decisions on defaults for params which alter the overall logic/behavior.
Please define this param and its default in the API layer exactly as you did for the `force` flag for the update API call. (and everywhere else along the proverbial stack)
Currently, only Migrations-from-Replicas actually make use of that `force` param, but we might as well define it for all operations as future-proofing.

coriolis/conductor/rpc/server.py
```diff
@@ -433,7 +443,7 @@ def delete_replica(self, ctxt, replica_id):
         db_api.delete_replica(ctxt, replica_id)
 
     @replica_synchronized
-    def delete_replica_disks(self, ctxt, replica_id):
+    def delete_replica_disks(self, ctxt, replica_id, force=False):
```
aznashwan: Same observation as with the `force` flag for `execute_replica_tasks`.

coriolis/conductor/rpc/server.py
```diff
@@ -561,7 +576,7 @@ def _get_provider_types(self, ctxt, endpoint):
 
     @replica_synchronized
     def deploy_replica_instances(self, ctxt, replica_id, clone_disks, force,
-                                 skip_os_morphing=False):
+                                 skip_os_morphing=False, replica_force=False):
```
aznashwan: Same observation as with the force flag for execute_replica_tasks.

aznashwan:
LGTM. Merging. Thanks!

aznashwan:
This param will be re-integrated by the replica-state-related updates which will come in the 2.0 version of the internal model. Closing for now.

PR #56, opened 2019-08-10
Create group for Replicator user.

aznashwan:
Create a group for the Replicator user on the export minion VM.
Additionally, add the SSH user to the group in order to be able to the replicator's certificates.


aznashwan:
Tested and works. Merging.

PR #57, opened 2019-08-19
Adds snapshotting and reverting mechanism during Replica executions

Dany9966:


coriolis/conductor/rpc/server.py
```diff
@@ -915,6 +941,23 @@ def _update_volumes_info_for_migration_parent_replica(
             self._update_replica_volumes_info(
                 ctxt, replica_id, instance, updated_task_info)
 
+    def _update_volumes_info(
+            self, ctxt, instance, execution, volumes_info):
+        if isinstance(execution.action, models.Migration):
```
aznashwan: Checking the operation we're on is bad and we should never have to do it. (neither using `isinstance`, or the type field you add to executions in #58)
We're already being cheeky with what we do in `_handle_post_task_actions`, and it should stay that way IMO.

In this sense, please define separate task classes for the snapshot creation/deletion during Replica executions (and also rename the old tasks please to have `Migration` in the name), and just handle each task normally in `_handle_post_task_actions ` as we currently do.
The actual calls to the providers can stay the same so you'll likely need to duplicate most of the code in the body of the task classes.

coriolis/conductor/rpc/server.py
```diff
@@ -924,19 +967,18 @@ def _handle_post_task_actions(self, ctxt, task, execution, task_info):
...
-                    {"volumes_info": volumes_info})
+                self._update_volumes_info(
+                    ctxt, task.instance, execution, volumes_info)
 
         elif task_type == constants.TASK_TYPE_DELETE_REPLICA_DISK_SNAPSHOTS:
-
-            if not task_info.get("clone_disks"):
+            # If this execution is a replica, volumes_info should NOT be
+            # deleted.
+            clone_disks = task_info.get("clone_disks", True)
```
aznashwan: 1) Any real reason to declare this variable separately considering it's only used once?
2) There is no reason why 'backend code' should make assumptions about params which imply extra work (such as defaulting 'clone_disks' to True here).
This is already done in the API-layer (where it should be), so please remove the default here.

aznashwan:
This issue will be avoided by the replica-state-related updates which will come in the 2.0 version of the internal model. Closing for now.

PR #58, opened 2019-08-20
Add 'type' field to tasks executions.

Dany9966:


coriolis/conductor/rpc/server.py
```diff
@@ -316,6 +316,7 @@ def execute_replica_tasks(self, ctxt, replica_id, shutdown_instances):
         execution.id = str(uuid.uuid4())
         execution.status = constants.EXECUTION_STATUS_RUNNING
         execution.action = replica
+        execution.type = 'replica_sync'
```
aznashwan: Please define constants in `constants.py` for all of these.

aznashwan: Also, call it `replica_execution` to have it in line with existing terminologies.

coriolis/db/sqlalchemy/models.py
```diff
@@ -84,6 +84,7 @@ class TasksExecution(BASE, models.TimestampMixin, models.ModelBase,
                              backref=orm.backref('execution'))
     status = sqlalchemy.Column(sqlalchemy.String(100), nullable=False)
     number = sqlalchemy.Column(sqlalchemy.Integer, nullable=False)
+    type = sqlalchemy.Column(sqlalchemy.String(20))
```
aznashwan: You need to add a transition script in `db/sqlalchemy/migrate_repo/versions` for any updates to the DB.

aznashwan:
PS: please also update the client to show the `type` field of Replica executions. (both for the `coriolis replica execution show/list`, and the plain `coriolis replica show` commands)

aznashwan:
LGTM. Merging. Thanks.

PR #60, opened 2019-08-28
Disable lvm2-lvmetad before disk sync

gabriel-samfira:
This change disables lvm2-lvmetad before starting the disk sync.


aznashwan:
LGTM. Merging. Thanks!

PR #61, opened 2019-08-29
Add _set_selinux_autorelabel()

gabriel-samfira:
Adds _set_selinux_autorelabel() to Linux OS morphers, that when called, will instruct the target system to relabel files on the root partition.


coriolis/osmorphing/base.py
```diff
@@ -211,3 +211,11 @@ def _replace_fstab_entries_device_prefix(
                 "mv -f /%s /%s.bak" % (fstab_chroot_path, fstab_chroot_path))
             self._write_file(
                 fstab_chroot_path, "\n".join(fstab_contents_lines))
+
+    def _set_selinux_autorelabel(self):
```
aznashwan: This is already implemented in the base RHEL tools (which all RHEL-based distro tools should inherit from.
Can you at least remove it from there as it's redundant?

gabriel-samfira: Done.

aznashwan:
LGTM. Merging now.

PR #63, opened 2019-09-03
Fix mount os

gabriel-samfira:
Find and mount root device with all virtual filesystems first, then the boot device if any, then any secondary partitions present in target machine fstab.

### DIFF
--- a/coriolis/osmorphing/osmount/base.py
+++ b/coriolis/osmorphing/osmount/base.py
@@ -347,8 +347,6 @@ class BaseLinuxOSMountTools(BaseSSHOSMountTools):
                 self._exec_cmd('sudo umount %s' % tmp_dir)
                 continue
 
-        if dev_path:
-            devices.remove(dev_path)
         return dev_name
 
     def _find_and_mount_root(self, devices):
@@ -389,6 +387,10 @@ class BaseLinuxOSMountTools(BaseSSHOSMountTools):
             self._exec_cmd(
                 'sudo mount -o bind /%(dir)s/ %(mount_dir)s' %
                 {'dir': directory, 'mount_dir': mount_dir})
+
+        if os_root_device in devices:
+            devices.remove(os_root_device)
+
         return os_root_dir, os_root_device
 
     def mount_os(self):
--- a/coriolis/osmorphing/osmount/base.py
+++ b/coriolis/osmorphing/osmount/base.py
@@ -1,5 +1,6 @@
 # Copyright 2016 Cloudbase Solutions Srl
 # All Rights Reserved.
+# pylint: disable=anomalous-backslash-in-string
 
 import abc
 import base64
@@ -244,7 +245,7 @@ class BaseLinuxOSMountTools(BaseSSHOSMountTools):
             try:
                 self._exec_cmd(mountcmd)
                 new_mountpoints.append(chroot_mountpoint)
-            except Exception as ex:
+            except Exception:
                 LOG.warn(
                     "Failed to run fstab filesystem mount command: '%s'. "
                     "Skipping mount. Error details: %s",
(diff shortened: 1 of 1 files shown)


coriolis/osmorphing/osmount/base.py
```diff
@@ -298,6 +299,99 @@ def _get_volume_block_devices(self):
...
+        if dev_path:
+            devices.remove(dev_path)
+        return dev_name
+
+    def _find_and_mount_root(self, devices):
+        files = ["etc", "bin", "sbin", "boot"]
+        os_root_dir = None
+        os_root_device = self._find_dev_with_contents(
+            devices, all_files=files,
+        )
```
aznashwan: NIT: these trailing parens 9both here and on 377+380) are quite an eyesore, could we just put them at the end of the previous line?

coriolis/osmorphing/osmount/base.py
```diff
@@ -298,6 +299,99 @@ def _get_volume_block_devices(self):
...
+
+        for directory in ['proc', 'sys', 'dev', 'run']:
+            mount_dir = os.path.join(os_root_dir, directory)
+            if not utils.test_ssh_path(self._ssh, mount_dir):
+                LOG.info(
+                    "No '%s' directory in mounted OS. Skipping mount.",
+                    directory)
+                continue
+            self._exec_cmd(
+                'sudo mount -o bind /%(dir)s/ %(mount_dir)s' %
```
aznashwan: Any reason for the trailing slash on the first arg?

coriolis/osmorphing/osmount/base.py
```diff
@@ -340,43 +434,13 @@ def mount_os(self):
...
-                self._exec_cmd('sudo umount %s' % tmp_dir)
+        os_root_dir, os_root_device = self._find_and_mount_root(
+            dev_paths_to_mount)
 
-        if not os_root_dir:
-            raise exception.OperatingSystemNotFound("root partition not found")
+        grub_dirs = ["grub", "grub2"]
+        os_boot_device = self._find_dev_with_contents(
+            dev_paths_to_mount, one_of_files=grub_dirs,
+        )
```
aznashwan: NIT: trailing paren.

coriolis/osmorphing/osmount/base.py
gabriel-samfira: code was moved, not changed. But as a note, the trailing slash disambiguates symlinks from actual folders.

```bash
gabriel@rossak:/tmp/test$ mkdir a_dir
gabriel@rossak:/tmp/test$ ln -s a_dir a_symlink
gabriel@rossak:/tmp/test$ ln -s a_dir2 a_broken_symlink
gabriel@rossak:/tmp/test$
gabriel@rossak:/tmp/test$ ls -l a_dir
total 0
gabriel@rossak:/tmp/test$ ls -l a_symlink
lrwxrwxrwx 1 gabriel gabriel 5 sep  3 16:18 a_symlink -> a_dir
gabriel@rossak:/tmp/test$ ls -l a_symlink/
total 0
gabriel@rossak:/tmp/test$ ls -l a_broken_symlink
lrwxrwxrwx 1 gabriel gabriel 6 sep  3 16:18 a_broken_symlink -> a_dir2
gabriel@rossak:/tmp/test$ ls -l a_broken_symlink/
ls: cannot access 'a_broken_symlink/': No such file or directory
```

coriolis/osmorphing/osmount/base.py
gabriel-samfira: Changed. The python plugin in my IDE automatically does this when I hit enter and the cursor is within the parenthesis.

coriolis/osmorphing/osmount/base.py
aznashwan: Had no clue, thanks for the clarification!

coriolis/osmorphing/osmount/base.py
```diff
@@ -298,6 +299,98 @@ def _get_volume_block_devices(self):
...
+                common = [i for i in one_of_files if i in dirs]
+                if len(common) > 0:
+                    dev_name = dev_path
+                    self._exec_cmd('sudo umount %s' % tmp_dir)
+                    break
+            else:
+                self._exec_cmd('sudo umount %s' % tmp_dir)
+                continue
+
+        if dev_path:
```
aznashwan: 1) I appreciate the idea of removing devices we've successfully identified so as not to pointlessly rescan them but IMO mutating an argument is not ideal and we should avoid it if possible (IMO the like 3s extra wait to rescan the volume is not that big of a deal in the grand scheme of things)
2) even if you decide against it, please declare a separate variable for the `dev_path_to_remove` and pop that away as currently you'll be popping the device last iterated over if and `all_files` arg is provided (there's no break in the first `if`)

coriolis/osmorphing/osmount/base.py
```diff
@@ -298,6 +299,98 @@ def _get_volume_block_devices(self):
...
+                )
+                continue
+
+            LOG.debug("Contents of device %s:\n%s", dev_path, dirs)
+
+            if all_files and dirs:
+                common = [i if i in dirs else None for i in all_files]
+                if not all(common):
+                    self._exec_cmd('sudo umount %s' % tmp_dir)
+                    continue
```
aznashwan: Shouldn't this be a `break`?

coriolis/osmorphing/osmount/base.py
```diff
@@ -298,6 +299,98 @@ def _get_volume_block_devices(self):
...
+                )
+                utils.ignore_exceptions(self._exec_cmd)(
+                    "sudo rmdir %s" % tmp_dir
+                )
+                continue
+
+            LOG.debug("Contents of device %s:\n%s", dev_path, dirs)
+
+            if all_files and dirs:
+                common = [i if i in dirs else None for i in all_files]
```
aznashwan: I agree the list comprehension is way more concise than a `for`, but considering we don't log the list of `common` ones or anything like that, IMO the resulting list should only strictly include booleans instead of `i`/`None`.
This will also prevent the `all()` call below from misbehaving if someone passes a `None`, `False` or `""` inside the `all_files` kwarg (though admittedly them doing it is reprehensible to begin with, but better safe than sorry)

gabriel-samfira: I hope we never allow code that could pass in ```False``` or ```None``` as an element in an array that is supposed to hold file names...

coriolis/osmorphing/osmount/base.py
gabriel-samfira: No. If not ```all``` files are available, this device is ignored. A break is called when using ```one_of_files```

coriolis/osmorphing/osmount/base.py
aznashwan: Absolutely agreed, but why not be safe and sorry if we can afford it?

coriolis/osmorphing/osmount/base.py
gabriel-samfira: The device is removed if, and only if one is found, prior to returning that particular value at the end of this function. Not all devices get removed.

The variable ```all_files``` is supposed to not break. It represents an all or nothing condition. Either all expected files are present, or the device is not relevant and we go on to check the next one in the list.

This ```if dev_path``` is outside of that loop, and ```dev_path``` only gets set to something other than ```None``` if a valid device is found. If none are found, nothing gets removed.

When one is found, it is returned as a value, which can later be used by the caller as a parameter in whichever manner they wish.

I can remove this, and add it to ```_find_and_mount_root()```. That function actually mounts the device, and there is no reason to keep that device in the list and risk mounting it twice, leading to potential cases where we forget about the second mountpoint and fail to ```umount``` it.

coriolis/osmorphing/osmount/base.py
aznashwan: My bad, misread the `not` in the condition above.

coriolis/osmorphing/osmount/base.py
gabriel-samfira: because my hope is that we never become so lazy to not notice someone defining a variable like:

```python
files_to_check = [ "/bin", "/sbin", None, "/home", False ]
```

and just merge it :). Python is not strongly typed, but we should at least be vigilant enough to notice something like that, no?

coriolis/osmorphing/osmount/base.py
aznashwan: Well either way this code will remove the last `dev_path` the `for` iterated over (which isn't necessarily the one we actually found), not to mention that we'll get a `NameError` if `devices` is an empty list (which we could just check for at the top if deemed necessary)
I realise now that you were probably meaning to check and remove the `dev_name` variable you return below, not `dev_path`.

gabriel-samfira: Moved to ```_find_and_mount_root()```

aznashwan:
LGTM. Merging. Thanks!

PR #64, opened 2019-09-04
Call sync in _write_file_sudo() and PEP8

gabriel-samfira:
This change calls sync in _write_file_sudo() to ensure that the files written to disk actually make it on the disk.


aznashwan:
LGTM. Merging. Thanks.

PR #65, opened 2019-09-11
Fix disk size unit printing in `_download_sparse_disk` method.

Dany9966:


aznashwan:
LGTM. Merging. Thanks!

PR #67, opened 2019-09-13
Replace `lvs --reportformat` output with `lvdisplay`.

Dany9966:


coriolis/osmorphing/osmount/base.py
```diff
@@ -124,27 +123,13 @@ def _get_vgnames(self):
...
-        for i, lvm in enumerate(report.get("lv", [])):
-            lvm_name = lvm.get("lv_name", str(i))
-            lvm_path = lvm.get("lv_path")
-            if not lvm_path:
-                LOG.warn("No path for lvm volume '%s'", lvm_name)
-            else:
-                lvm_paths.append(lvm_path)
+        for line in out:
+            lvm_vol = line.strip().split(':')
+            if lvm_vol is not None:
```
aznashwan: List emptiness checks should be just `if list`.
`lvm_vol is not None` will only return True if `lvm_vol = None`, and not if the `lvm_vol = []` so we're still open to the `IndexError` below

coriolis/osmorphing/osmount/base.py
```diff
@@ -258,6 +249,28 @@ def _check_mount_fstab_partitions(
 
         return new_mountpoints
 
+    def _get_symlink_target(self, symlink):
+        target = ''
```
aznashwan: NIT: whenever writing methods like this which may or may not return a thing, always use `None` as the default value

coriolis/osmorphing/osmount/base.py
```diff
@@ -258,6 +249,28 @@ def _check_mount_fstab_partitions(
...
+
+    def _get_device_file_paths(self, symlink_list):
+        """ Reads a list of symlink paths, such as `/dev/GROUP/VOLUME0` or
+        `/dev/mapper/GROUP-VOLUME0` and returns a list of respective links'
+        target device file paths, such as `/dev/dm0`
+        """
+        dev_file_paths = []
+        for link in symlink_list:
+            dev_file = self._get_symlink_target(link)
+            if not dev_file:
```
aznashwan: This means that if for whatever reason there is a file which isn't a symlink it'll get completely "lost in translation"
IMO either this method should be re-written to just return the same path if something is not a symlink

aznashwan:
LGTM. Merging. Thanks!

PR #68, opened 2019-09-17
Directly update parameters of executionless Replicas.

aznashwan:


coriolis/conductor/rpc/server.py
```diff
@@ -1124,13 +1124,24 @@ def update_replica(
...
         self._check_replica_running_executions(ctxt, replica)
-        self._check_valid_replica_tasks_execution(replica, force=True)
+
         execution = models.TasksExecution()
         execution.id = str(uuid.uuid4())
         execution.status = constants.EXECUTION_STATUS_RUNNING
         execution.action = replica
         execution.type = constants.EXECUTION_TYPE_REPLICA_UPDATE
 
+        if not replica.executions:
```
gabriel-samfira: Does replica.executions hold soft deleted executions? Just because there are no previous executions, it does not mean that there were none ran, but deleted. So you may in fact have ```volumes_info``` even on replicas that have no undeleted executions.

aznashwan: No, they should not afaik.
This is only a problem if someone tries to update an executions Replica or if they deleted all the executions of a Replica before issuing an update call.

Either way, I'll be rewriting this to simply error out, as direct Replica updates is a bit more complicated considering the API call should result in an execution, and would better be put up for consideration for a V2 interface.

aznashwan:
This issue will be avoided by the replica-state-related updates which will come in the 2.0 version of the internal model. Closing for now.

PR #69, opened 2019-09-23
Update replicator binary

gabriel-samfira:
Ignores removable devices.


aznashwan:
Tested and works. Merging.

PR #72, opened 2019-10-01
Move 'public_ip_address' out of the 'devices' list in the VM info schema.

aznashwan:
While this mixup doesn't directly functionality (the plugins still put the field in the right place), it warrants fixing for the API to be in line with the schema.


aznashwan:
Trivial, merging.

PR #73, opened 2019-10-07
Copy resolv.conf contents for links during OSMorphing.

aznashwan:
Depending on the OSMorphing worker VM OS, resolv.conf might be a symlink
with a relative path.
This commit makes Coriolis always copy the contents of resolv.conf when
patching it during the OSMorphing process.


aznashwan:
Trivial, merging.

PR #74, opened 2019-10-08
Add 'debug_os_morphing_errors' flag to conductor.

aznashwan:
When set, the conductor will un-schedule all cleanup tasks following an
error'd OSMorphing task, allowing for manual connection and inspection
of the worker setup.


aznashwan:
Tested and works. Merging.

PR #75, opened 2019-10-09
Add caching module

gabriel-samfira:
Adds a caching module that providers may leverage.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,7 @@ keystoneauth1
 keystonemiddleware
 jsonschema
 PyMySQL
+oslo.cache
 oslo.concurrency
 oslo.config
 oslo.context>=2.19.1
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -27,6 +28,7 @@ python-barbicanclient
 python-swiftclient>=3.2
 git+https://github.com/cloudbase/pywinrm.git@requests#egg=pywinrm
 PyYAML
+redis
```
aznashwan: Although it's annoying that we have to specify these dependencies explicitly, I suggest we also add `python-mecached` as a dep so we can cover both most popular caching distros.
(The current Kolla we use install memcached, and only the updated Kolla @ionutbalutoiu is integrating will feature redis (and possible mecached too alongside it)

gabriel-samfira: It's already there:

https://github.com/cloudbase/coriolis/blob/master/requirements.txt#L25

PR #76, opened 2019-10-16
Ignore LVM notations, always get device paths.

Dany9966:
Fixes: https://cloudbasedev.atlassian.net/projects/COR/issues/COR-144


aznashwan:
LGTM. Merging. Thanks!

PR #77, opened 2019-10-16
Switch to using Replica logic for Migrations.

aznashwan:
This patchset switches from the old disk-export-based logic for
Migrations to replying on the same underlying procedures as Replicas.

Now, Migrations will simply be the Replica execution and OSMorphing stages
stiched together.

Changes to Migrations API:
    * 'replication_count' field added. Controls the
      number of Replica syncs to be done on the VM.
    * 'shutdown_instances' field added. Controls whether
      or not to shutdown the source VM during the final sync.

Coriolis UI will need to change how it identifies source Migration
providers. (the identifier for old exporter types has been disabled)


aznashwan:
Tested and works. Merging now and awaiting on UI updates to roll out after that.

PR #78, opened 2019-10-23
Add resources/bin directory

ionutbalutoiu:
Move the static binaries from the `resources` directory
to `resources/bin` directory.

This is useful when running the Coriolis service into a
container, as `resources/bin` can be used as a mount point
with the binaries separately built.


aznashwan:
LGTM. Merging. Thanks!

PR #84, opened 2019-11-07
Only ignore LVM notations when checking for mounted devices.

Dany9966:


aznashwan:
LGTM. Merging. Thanks!

PR #85, opened 2019-11-11
Handle writer error

gabriel-samfira:
* check for out of bounds write
  * check if write_data died before attempting to send a chunk. This will return a proper error code, and not mask it with ```Socket closed``` messages.

### DIFF
--- a/coriolis/providers/backup_writers.py
+++ b/coriolis/providers/backup_writers.py
@@ -34,6 +34,26 @@ _CORIOLIS_HTTP_WRITER_CMD = "coriolis-writer"
 
 LOG = logging.getLogger(__name__)
 
+_WRITER_ERR_MAP = {
+    -1: "ERR_MORE_MSG",
+    0: "ERR_DONE",
+    1: "ERR_READ_MSG_SIZE",
+    2: "ERR_MSG_SIZE",
+    3: "ERR_OPEN_FILE",
+    4: "ERR_DATA",
+    5: "ERR_IO_OPEN",
+    6: "ERR_IO_SEEK",
+    7: "ERR_IO_WRITE",
+    8: "ERR_IO_CLOSE",
+    9: "ERR_NO_MEM",
+    10: "ERR_INVALID_ARGS",
+    11: "ERR_READ_MSG_ID",
+    12: "ERR_MSG_SIZE_INFLATED",
+    13: "ERR_ZLIB",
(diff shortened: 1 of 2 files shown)


coriolis/providers/backup_writers.py
```diff
@@ -300,7 +327,7 @@ def _connect_ssh(self):
                 username=self._username,
                 pkey=self._pkey,
                 password=self._password)
-        except:
+        except BaseException:
```
aznashwan: NIT: I would personally go for an explicit `except (Exception, KeyboardInterrupt)`, as `KeyboardInterrupt` is the only other "level 2 exception" we're actually expecting.

coriolis/providers/backup_writers.py
```diff
@@ -713,7 +740,7 @@ def _connect_ssh(self):
...
```
aznashwan: `except (Exception, KeyboardInterrupt)`

coriolis/resources/write_data.c
```diff
@@ -130,10 +133,21 @@ int handle_msg(FILE* input_stream)
     FILE* f = fopen(path, "rb+");
     if (!f)
         return ERR_OPEN_FILE;
+
+    if (fseek(f, 0, SEEK_END))
```
aznashwan: What does this seek achieve? Shouldn't seeking 0 bytes from the end of the file always work?
Did you mean to `fseek(f, offset + data_size, SEEK_SET)`?
That should also eliminate the need for the added `if (end_write > disk_size)` at the end. (though it will mean we give out the more generic `ERR_IO_SEEK`)

gabriel-samfira: This gives you the disk size. You seek to byte 0 relative to the end, then call ```ftell()``` to give you the offset. That offset is your disk size.

I would rather compare values than try a pythonism in C, where I check for an error rather than compare values.

coriolis/providers/backup_writers.py

coriolis/providers/backup_writers.py
gabriel-samfira: Done

coriolis/resources/write_data.c
aznashwan: Failed to properly notice the `ftell` below, great point.

aznashwan:
LGTM. Merging. Thanks!

PR #89, opened 2019-11-26
Redact volume_info when replicatoris used

gabriel-samfira:
Avoid logging the replicator state.

### DIFF
--- a/coriolis/tasks/replica_tasks.py
+++ b/coriolis/tasks/replica_tasks.py
@@ -9,6 +9,7 @@ from coriolis import exception
 from coriolis.providers import factory as providers_factory
 from coriolis import schemas
 from coriolis.tasks import base
+from coriolis import utils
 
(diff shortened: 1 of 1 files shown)


coriolis/tasks/replica_tasks.py
```diff
@@ -45,12 +45,26 @@ def _check_ensure_volumes_info_ordering(export_info, volumes_info):
 
         ordered_volumes_info.append(matching_volumes[0])
 
+    vol_info_cpy = None
+    if volumes_info:
```
aznashwan: Can you please use the below utility method as it's functionally identical and we can reuse it anywhere we need it.
https://github.com/cloudbase/coriolis/blob/master/coriolis/utils.py#L528-L540

gabriel-samfira: Done

PR #90, opened 2019-11-26
Add disk attachment detection functionality to Replicator.

aznashwan:
For platforms where the disk attachment order is unpredictable, the
plugins can choose to call a method on the Replicator object which
automatically runs and detects the new disk attachment, and returns the
new volumes_info for it.

### DIFF
--- a/coriolis/providers/replicator.py
+++ b/coriolis/providers/replicator.py
@@ -283,6 +283,133 @@ class Replicator(object):
             self._setup_replicator)(self._ssh)
         utils.retry_on_error()(
             self._init_replicator_client)(self._credentials)
+        LOG.debug(
+            "Disk status after Replicator initialization: %s",
+            self._cli.get_status(device=None, brief=True))
+
+    def get_current_disks_status(self):
+        """ Returns a list of the current status of the attached data disks.
+        The root disk of the worker VM is NOT returned.
+        The result is a list with elements of the following format:
+        [{
+            'device-path': '/dev/xvdf',
+            'device-name': 'xvdf',
+            'size': 10737418240,
+            'checksum-algorithm': 'sha256',
+            'chunk-size': 10485760,
+            'logical-sector-size': 512,
+            'physical-sector-size': 512,
+            'alignment-offset': 0,
+            'has-mounted-partitions': False,
+            'checksum-status': {
+                'status': 'running',
+                'total-chunks': 1024,
+                'completed-chunks': 0,
+                'percentage': 0},
+            'partitions': [{
+                'name': 'xvdf1',
+                'sectors': 20969439,
+                'uuid': '73e9659d-2fd9-46ca-a341-5a8637c416ee',
+                'fs': 'ext4',
+                'start-sector': 2048,
+                'end-sector': 20971486,
+                'alignment-offset': 0}]
+        }]
+        """
+        return self._cli.get_status()
+
+    def attach_new_disk(
+            self, disk_id, attachf, previous_disks_status=None,
+            retry_period=30, retry_count=10):
+        """ Returns the volumes_info for the disk attached by running
+        `attachf`. This is achieved by comparing the disk statuses before and
+        after the execution of the attachment operation.
+
+        param disk_id: str(): the 'disk_id' of the info from self._volumes_info
+            for the disk which shall be attached by `attachf()`.
+        param attachf: fun(): argument-less function which attaches the disk.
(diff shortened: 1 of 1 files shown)


coriolis/providers/replicator.py
```diff
@@ -283,6 +283,114 @@ def init_replicator(self):
...
+
+        # get/refresh current device paths:
+        if not previous_disks_status:
+            previous_disks_status = self._cli.get_status()
+        previous_device_paths = [
+            dev['device-path'] for dev in previous_disks_status]
+
+        # run attachment function and get new device paths:
+        attachf()
+        new_disks_status = self._cli.get_status()
```
gabriel-samfira: This should probably be polled several times. Devices may be detected slower than expected.

coriolis/providers/replicator.py
```diff
@@ -283,6 +283,114 @@ def init_replicator(self):
...
+                "no longer detected: %s", [
+                    dev for dev in previous_disks_status
+                    if dev['device-path'] in missing_device_paths])
+
+        new_device_paths = set(new_device_paths) - set(previous_device_paths)
+        if not new_device_paths:
+            raise exception.CoriolisException(
+                "No new device paths have appeared after volume attachment of "
+                "disk with ID: %s" % disk_id)
+        elif len(new_device_paths) > 2:
```
gabriel-samfira: Shouldn't there be just one new device?

coriolis/providers/replicator.py
```diff
@@ -283,6 +283,114 @@ def init_replicator(self):
...
+        """
+        # check if volume with given ID is present in self._volumes_info:
+        matching_vols = [
+            vol for vol in self._volumes_info
+            if vol['disk_id'] == disk_id]
+        if not matching_vols:
+            raise exception.CoriolisException(
+                "No information regarding volume with ID '%s'. "
+                "Cannot track its attachment." % disk_id)
+        elif len(matching_vols) > 2:
```
gabriel-samfira: Shouldn't this be ```!= 1```?

Also, this can be a standalone if, instead of an ```elif```

coriolis/providers/replicator.py
```diff
@@ -283,6 +283,114 @@ def init_replicator(self):
...
+                "disk with ID: %s" % disk_id)
+        elif len(new_device_paths) > 2:
+            raise exception.CoriolisException(
+                "Multiple device paths have appeared after attachment of disk "
+                "with ID %s: %s" % (
+                    disk_id,
+                    [dev for dev in new_disks_status
+                     if dev['device-path'] in new_device_paths]))
+
+        new_dev = [
```
gabriel-samfira: At this point you should have only one disk in ```new_device_paths```, considering you are attaching one disk at a time, or at least that's what the ```attachf()``` function should do.

Wouldn't it be easier to just assign new_device_path to ```vol_info['disk_path']```?

coriolis/providers/replicator.py
aznashwan: Good catch, fixing now

coriolis/providers/replicator.py
aznashwan: As mentioned in the docstring, it's the responsibility of `attachf` to wait as much as it considers necessary.

coriolis/providers/replicator.py
aznashwan: Yes, the check is here in case `attachf` accidentally attaches more.
The check is wrong though, should also `> 1`

coriolis/providers/replicator.py
aznashwan: These checks are all here so as to ensure `attachf()` doesn't accidentally attach more than one disk (which is likely to happen considering most current plugins implement a single `attach_disks` method, so a lazy implementation which reuses that will get caught here)

coriolis/providers/replicator.py
gabriel-samfira: How would it know without checking with the replicator status if the disk has appeared on the server or not?

coriolis/providers/replicator.py
gabriel-samfira: Yes. Understood. But at this point, you already passed the above tests where you check if ```len(new_device_paths) > 1```. It should be safe to trust that we have only 1 element in that set.

aznashwan: You are correct, I got a bit blind-sighted, fixing now.

aznashwan:
Tested and works with latest Replicator patches. Merging.

PR #91, opened 2019-11-27
Schedule VALIDATE_SOURCE tasks before GET_INSTANCE.

aznashwan:
This patch makes the conductor schedule the VALIDATE_SOURCE_*_INPUTS
tasks before the GET_INSTANCE_INFO ones of both Migrations and Replicas.
This is so as to allow the source validation to run before the instance
query (which might depend on params validated by the former), as well as
prevent any state conflicts while the two tasks are running.


aznashwan:
Tested and confirmed to work. Merging.

PR #93, opened 2019-12-05
Add user defined scripts for OS morphing

gabriel-samfira:
These user defined scripts will run before any OS morphing operation
will take place.

### DIFF
--- a/coriolis/api/v1/migrations.py
+++ b/coriolis/api/v1/migrations.py
@@ -98,7 +98,7 @@ class MigrationController(api_wsgi.Controller):
         migration_body = body.get("migration", {})
         context = req.environ['coriolis.context']
         context.can(migration_policies.get_migrations_policy_label("create"))
-
+        user_scripts = migration_body.get("user_scripts")
         replica_id = migration_body.get("replica_id")
         if replica_id:
             clone_disks = migration_body.get("clone_disks", True)
@@ -108,7 +108,8 @@ class MigrationController(api_wsgi.Controller):
             # NOTE: destination environment for replica should have been
             # validated upon its creation.
             migration = self._migration_api.deploy_replica_instances(
-                context, replica_id, clone_disks, force, skip_os_morphing)
+                context, replica_id, clone_disks, force, skip_os_morphing,
+                user_scripts=user_scripts)
         else:
             (origin_endpoint_id,
              destination_endpoint_id,
@@ -127,7 +128,8 @@ class MigrationController(api_wsgi.Controller):
                 source_environment, destination_environment, instances,
                 network_map, storage_mappings, replication_count,
                 shutdown_instances, notes=notes,
-                skip_os_morphing=skip_os_morphing)
+                skip_os_morphing=skip_os_morphing,
+                user_scripts=user_scripts)
 
         return migration_view.single(req, migration)
 
--- a/coriolis/api/v1/migrations.py
+++ b/coriolis/api/v1/migrations.py
@@ -98,7 +98,7 @@ class MigrationController(api_wsgi.Controller):
         migration_body = body.get("migration", {})
         context = req.environ['coriolis.context']
         context.can(migration_policies.get_migrations_policy_label("create"))
-        user_scripts = migration_body.get("user_scripts")
+        user_scripts = migration_body.get("user_scripts", {})
         replica_id = migration_body.get("replica_id")
         if replica_id:
             clone_disks = migration_body.get("clone_disks", True)
--- a/coriolis/conductor/rpc/server.py
+++ b/coriolis/conductor/rpc/server.py
@@ -693,6 +693,7 @@ class ConductorServerEndpoint(object):
         return self.get_migration(ctxt, migration.id)
 
     def _get_instance_scripts(self, user_scripts, instance):
+        user_scripts = user_scripts or {}
         ret = {
             "global": user_scripts.get("global", {}),
             "instances": {},
--- a/coriolis/conductor/rpc/server.py
+++ b/coriolis/conductor/rpc/server.py
@@ -574,7 +574,7 @@ class ConductorServerEndpoint(object):
 
     @replica_synchronized
     def deploy_replica_instances(self, ctxt, replica_id, clone_disks, force,
-                                 skip_os_morphing=False):
+                                 skip_os_morphing=False, user_scripts=None):
         replica = self._get_replica(ctxt, replica_id)
         self._check_reservation_for_transfer(replica)
         self._check_replica_running_executions(ctxt, replica)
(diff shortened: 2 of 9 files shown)


coriolis/conductor/rpc/server.py
```diff
@@ -729,6 +743,8 @@ def migrate_instances(self, ctxt, origin_endpoint_id,
             # NOTE: we must explicitly set this in each VM's info
             # to prevent the Replica disks from being cloned:
             migration.info[instance] = {"clone_disks": False}
+            scripts = self._get_instance_scripts(user_scripts, instance)
+            migration.info[instance]["user_scripts"] = scripts
```
aznashwan: Please .get("user_scripts", {}) to preserve backwards-compatibility.

coriolis/osmorphing/base.py
```diff
@@ -96,6 +100,38 @@ def get_packages(self):
 
         return add, remove
 
+    def run_user_script(self, user_script):
+        if len(user_script) == 0:
+            return
+
+        script_path = "/tmp/coriolis_user_script"
```
aznashwan: NIT: I'd still use a `tempfile` for safety

For posterity, I'd also consider putting the script in some non-tmp dir. (the temp dir of the worker is bind-mounted during the `chroot` operations so it won't be preserved on the migrated VM)

gabriel-samfira: The user has the script, and it will also be a permanent part of the migration task_info. Considering you already have it in 2 places, does it make sense to have it in a 3rd?

coriolis/api/v1/migrations.py
```diff
@@ -98,7 +98,7 @@ def create(self, req, body):
         migration_body = body.get("migration", {})
         context = req.environ['coriolis.context']
         context.can(migration_policies.get_migrations_policy_label("create"))
-
+        user_scripts = migration_body.get("user_scripts")
```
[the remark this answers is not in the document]
gabriel-samfira: Done

coriolis/osmorphing/base.py
gabriel-samfira: I need to clarify this a bit more. This is not meant to be copied inside the migrated VM. This is meant to be copied to the worker itself, and run with one parameter: the root_dir of the migrated instance. From there on out, the script can decide what it needs to do to the offline migrated VM.

coriolis/osmorphing/base.py
```diff
@@ -96,6 +100,38 @@ def get_packages(self):
...
+        try:
+            utils.write_ssh_file(
+                self._conn,
+                script_path,
+                user_script)
+        except Exception as err:
+            LOG.exception(err)
+            raise exception.CoriolisException(
+                "Failed to copy user script to target system."
+                " Error was: %s" % err)
```
aznashwan: NIT: Considering we decided to only support Python 3, you can remove the `LOG.error` call and just say `raise exception.CoriolisException(...) from err` to preserve both stack traces together

coriolis/osmorphing/base.py
```diff
@@ -96,6 +100,38 @@ def get_packages(self):
...
+                "sudo chmod +x %s" % script_path,
+                get_pty=True)
+
+            utils.exec_ssh_cmd(
+                self._conn,
+                'sudo "%s" "%s"' % (script_path, self._os_root_dir),
+                get_pty=True)
+        except Exception as err:
+            LOG.exception(err)
+            raise exception.CoriolisException(
```
aznashwan: NIT: Considering we decided to only support Python 3, you can remove the LOG.error call and just say raise exception.CoriolisException(...) from err to preserve both stack traces together

coriolis/osmorphing/windows.py
```diff
@@ -221,3 +221,34 @@ def _create_service(self, key_name, service_name, image_path,
...
+
+        script_path = "$env:TMP\\coriolis_user_script.ps1"
+        try:
+            utils.write_winrm_file(
+                self._conn,
+                script_path,
+                user_script)
+        except Exception as err:
+            LOG.exception(err)
+            raise exception.CoriolisException(
```
aznashwan: NIT: Considering we decided to only support Python 3, you can remove the LOG.error call and just say raise exception.CoriolisException(...) from err to preserve both stack traces together

coriolis/osmorphing/windows.py
```diff
@@ -221,3 +221,34 @@ def _create_service(self, key_name, service_name, image_path,
...
+        cmd = ('$ErrorActionPreference = "Stop"; powershell.exe '
+               '-NonInteractive -ExecutionPolicy RemoteSigned '
+               '-File "%(script)s" "%(os_root_dir)s"') % {
+            "script": script_path,
+            "os_root_dir": self._os_root_dir,
+        }
+        try:
+            out = self._conn.exec_ps_command(cmd)
+            LOG.debug("User script output: %s" % out)
+        except Exception as err:
```
aznashwan: NIT: Considering we decided to only support Python 3, you can remove the LOG.error call and just say raise exception.CoriolisException(...) from err to preserve both stack traces together

coriolis/tasks/osmorphing_tasks.py
```diff
@@ -25,11 +25,22 @@ def run(self, ctxt, instance, origin, destination, task_info,
...
         osmorphing_info = task_info.get('osmorphing_info', {})
 
+        user_scripts = task_info.get("user_scripts")
+        instance_script = None
+        if user_scripts:
+            instance_script = user_scripts.get("instances", {}).get(instance)
+            if not instance_script:
+                os_type = osmorphing_info.get("os_type")
+                if os_type:
+                    instance_script = user_scripts.get(
```
aznashwan: I'm not entirely sure how we've implemented this in the UI but we should be sure that it's clear to the user that any per-instance scripts will completely override the global ones (I was under the initial assumption that the globals would always get executed, so it's something I imagine some users would make thus supposition too)

coriolis/utils.py
```diff
@@ -204,6 +204,33 @@ def write_ssh_file(ssh, remote_path, content):
...
 
 
+@retry_on_error()
+def write_winrm_file(conn, remote_path, content):
+    """This is a poor man's scp command that transfers small
+    files, in chunks, over WinRM.
+    """
+    if conn.test_path(remote_path):
+        conn.exec_ps_command(
+            'Remove-Item -Force "%s"' % remote_path)
```
aznashwan: For safety, I'd make this method check if the file exists and error out if so.
This would imply either removing the retry or moving it to the inner body. (though I'd just remove it entirely considering the `exec_ps_command` already retries intrinsically)

aznashwan:
LGTM. Merging. Thanks!

PR #95, opened 2019-12-10
Send sudo SSH commands through a pseudo-terminal.

Dany9966:
There are situations where sshd_config can be set to require a
tty when sending sudo commands(Default requiretty). This patch
should always set Coriolis to send sudo commands using pty.

### DIFF
--- a/coriolis/osmorphing/base.py
+++ b/coriolis/osmorphing/base.py
@@ -125,11 +125,12 @@ class BaseLinuxOSMorphingTools(BaseOSMorphingTools):
         return utils.list_ssh_dir(self._ssh, path)
 
     def _exec_cmd(self, cmd):
-        return utils.exec_ssh_cmd(self._ssh, cmd, self._environment)
+        return utils.exec_ssh_cmd(self._ssh, cmd, self._environment,
+                                  get_pty=True)
 
     def _exec_cmd_chroot(self, cmd):
         return utils.exec_ssh_cmd_chroot(
-            self._ssh, self._os_root_dir, cmd, self._environment)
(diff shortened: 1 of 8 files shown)


coriolis/osmorphing/base.py
```diff
@@ -125,11 +125,13 @@ def _list_dir(self, chroot_path):
         return utils.list_ssh_dir(self._ssh, path)
 
     def _exec_cmd(self, cmd):
-        return utils.exec_ssh_cmd(self._ssh, cmd, self._environment)
+        get_pty = 'sudo ' in cmd
```
gabriel-samfira: I think we can just set the default value for ```get_pty``` to ```True```, and skip this check. We set it to ```True``` in most places, as almost all commands we run, require elevated privileges.

coriolis/utils.py
```diff
@@ -274,12 +274,15 @@ def exec_ssh_cmd(ssh, cmd, environment=None, get_pty=False):
             "Command \"%s\" failed with exit code: %s\n"
             "stdout: %s\nstd_err: %s" %
             (cmd, exit_code, std_out, std_err))
+    if get_pty:
```
gabriel-samfira: The python ```splitlines()``` function does a good job in identifying the newline termination properly:

```python
>>> data = "this is \r\ntest \n with carriage return"
>>> print(data)
this is 
test 
 with carriage return
>>> print(data.splitlines())
['this is ', 'test ', ' with carriage return']
```

Do we need to replace ```\r\n``` with ```\n``` ?

aznashwan:
Merging, thanks!

PR #96, opened 2019-12-12
Fix storage rescan

gabriel-samfira:
On older Windows systems there is no Update-HostStorageCache
commandlet. This replaces the powershell rescan with a diskpart
equivalent.


aznashwan:
Trivial, merging.

PR #97, opened 2019-12-13
Use diskpart to bring basic disks online

gabriel-samfira:
This works on all versions of Windows. It makes sense to have one common way to do this.


aznashwan:
LGTM. Merging. Thanks!

PR #98, opened 2020-01-09
Fix pseudo-terminal command output

Dany9966:
Most of the commands sent through SSH require root privileges, which
we have to send through pseudo-terminal. Unfortunately, the pty outputs
an additional '\r' to every newline. This may affect some plugins too.
By replacing '\r\n' with '\n', the outputs should be the same as
before using pseudo-terminal.


aznashwan:
LGTM. Merging. Now.

PR #101, opened 2020-01-13
Log the identified release info during Windows OSMorphing.

aznashwan:


aznashwan:
Tested and work. Merging.

PR #102, opened 2020-01-13
Return a more friendly exception

gabriel-samfira:
The original exception can be quite large and useless to the user.
This wraps the original exception in a more user friendly error
message. It should present a more informative message to the user
while still outputing useful debug info in the logs.

### DIFF
--- a/coriolis/osmorphing/manager.py
+++ b/coriolis/osmorphing/manager.py
@@ -123,7 +123,14 @@ def morph_image(origin_provider, destination_provider, connection_info,
         if packages_add:
             event_manager.progress_update(
                 "Adding packages: %s" % str(packages_add))
-            import_os_morphing_tools.install_packages(packages_add)
+            try:
+                import_os_morphing_tools.install_packages(
+                    packages_add)
+            except Exception as err:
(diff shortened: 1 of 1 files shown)


coriolis/osmorphing/manager.py
```diff
@@ -123,7 +123,13 @@ def morph_image(origin_provider, destination_provider, connection_info,
         if packages_add:
             event_manager.progress_update(
                 "Adding packages: %s" % str(packages_add))
-            import_os_morphing_tools.install_packages(packages_add)
+            try:
+                import_os_morphing_tools.install_packages(
+                    packages_add)
+            except Exception as err:
+                raise exception.CoriolisException(
+                    "Failed to install packages: %s" % ", ".join(
```
aznashwan: Considering we're hiding the real error message in the Replica view, I'd also add a "Please review the logs for more details" to the message

gabriel-samfira: Was actually thinking about this. Adding.

aznashwan:
LGTM. Merging. Thanks!

PR #103, opened 2020-01-14
Parse FSTab entries that don't have final digits

Dany9966:


coriolis/osmorphing/osmount/base.py
```diff
@@ -172,7 +172,7 @@ def _check_mount_fstab_partitions(
         # fstab entry format:
         # <device> <mountpoint> <filesystem> <options> <dump> <fsck>
         fstab_entry_regex = (
-            "^(\s*([^#\s]+)\s+(\S+)\s+(\S+)\s+(\S+)\s+(\d)\s+(\d)\s*)$")
+            "^(\s*([^#\s]+)\s+(\S+)\s+(\S+)\s+(\S+)\s*((\d)\s+(\d)*)*\s*)$")
```
aznashwan: That last `*` can be a `?` to ensure we don't get too multiple matches of the last two digits. (though please double-check in the interpreter in case I'm wrong)

aznashwan:
LGTM. Merging. Thanks!

PR #104, opened 2020-01-15
Use queueing for SSH backup writer

gabriel-samfira:
This should speed up the transfer speed when using the SSH backup
writer. Currently, there are 3 operations that happen:

1) Read from the source
2) Compress and encode payload
3) Send payload to destination

These 3 operations run in sequence for each chunk, which means that
while we are reading from the source, we are not compressing or
sending. While we are compressing and encoding, we are not reading
from the source or sending to destination, etc. Using queues allows
us to do more work in parallel.

### DIFF
--- a/coriolis/providers/backup_writers.py
+++ b/coriolis/providers/backup_writers.py
@@ -3,6 +3,7 @@
 
 import abc
 import contextlib
+import datetime
 import errno
 import os
 import tempfile
@@ -162,13 +163,22 @@ class FileBackupWriter(BaseBackupWriter):
 
 
 class SSHBackupWriterImpl(BaseBackupWriterImpl):
-    def __init__(self, path, disk_id, compress_transfer=None):
+    def __init__(self, path, disk_id, compress_transfer=None,
+                 encoder_count=3):
         self._msg_id = None
         self._stdin = None
         self._stdout = None
         self._stderr = None
         self._offset = None
         self._ssh = None
+        self._sender_q = eventlet.Queue(maxsize=5)
+        self._enc_q = eventlet.Queue(maxsize=5)
+        self._sender_evt = None
+        self._encoder_evt = []
+        self._encoder_cnt = encoder_count
+        self._exception = None
+        self._closing = False
+
         self._compress_transfer = compress_transfer
         if self._compress_transfer is None:
             self._compress_transfer = CONF.compress_transfers
@@ -184,10 +194,10 @@ class SSHBackupWriterImpl(BaseBackupWriterImpl):
         self._stdin, self._stdout, self._stderr = self._ssh.exec_command(
             "chmod +x write_data && sudo ./write_data")
 
-    def _encode_data(self, content):
+    def _encode_data(self, content, offset, msg_id):
         msg = data_transfer.encode_data(
-            self._msg_id, self._path,
-            self._offset, content,
+            msg_id, self._path,
+            offset, content,
             compress=self._compress_transfer)
 
         LOG.debug(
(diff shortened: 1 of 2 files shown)


coriolis/providers/backup_writers.py
```diff
@@ -169,6 +169,13 @@ def __init__(self, path, disk_id, compress_transfer=None):
         self._stderr = None
         self._offset = None
         self._ssh = None
+        self._sender_q = eventlet.Queue(maxsize=5)
+        self._enc_q = eventlet.Queue(maxsize=5)
+        self._sender_evt = None
+        self._encoder_evt = None
+        self._exception = None
+        self._closing = False
```
aznashwan: NIT: I'd rename this to `self._closed` to better indicate it represents the state of being closed or not

coriolis/providers/backup_writers.py
```diff
@@ -213,29 +220,98 @@ def _send_msg(self, data):
...
         self._stdin.write(data)
         self._stdin.flush()
         self._stdout.read(4)
 
     def _open(self):
         self._exec_helper_cmd()
+        self._sender_evt = eventlet.spawn(
+            self._sender)
+        self._encoder_evt = [
+            eventlet.spawn(self._encoder),
```
aznashwan: NIT: I'd suggest we add a configuration option for this instead of hardcoding it to 3

coriolis/providers/backup_writers.py
```diff
@@ -169,6 +169,13 @@ def __init__(self, path, disk_id, compress_transfer=None):
         self._stderr = None
         self._offset = None
         self._ssh = None
+        self._sender_q = eventlet.Queue(maxsize=5)
+        self._enc_q = eventlet.Queue(maxsize=5)
+        self._sender_evt = None
+        self._encoder_evt = None
```
aznashwan: NIT: I'd default this to `[]` to prevent possible accidental iterations over `None`

coriolis/providers/backup_writers.py
```diff
@@ -213,29 +220,98 @@ def _send_msg(self, data):
...
+            "data": data,
+            "msg_id": self._msg_id,
+        }
+        self._enc_q.put(payload)
         self._offset += len(data)
+        self._msg_id += 1
+
+    def _wait_for_queues(self):
+        LOG.info("Waiting for unfinished transfers to complete")
+        while (self._enc_q.unfinished_tasks or
```
aznashwan: I'd add a reasonable timeout here in case either we or eventlet screws up and this hangs.

coriolis/providers/backup_writers.py
gabriel-samfira: It represents a transition to closing state. When we call ```close()```, there may still be chunks in transit. The ```close()``` function waits for eventlets to finish sending chunks before it returns. However, we still need to make sure no new ```write()``` operations are allowed if ```close()``` is called.

So in short, this variable indicates that we have called ```close()``` but some operations may still be pending.

coriolis/providers/backup_writers.py

coriolis/providers/backup_writers.py
gabriel-samfira: Done

coriolis/providers/backup_writers.py
gabriel-samfira: Suggestions?

aznashwan: Considereing we only call this "on the way out" before closing, I'd expect 120s would strike a good balance between and not too long, my only concern is that it doesn't somehow hang forever due to an edge-case

gabriel-samfira: set a 600 second timeout. Just to account for really slow networks. If need be, we can lower it later.

aznashwan:
LGTM. Merging. Thanks!

PR #105, opened 2020-01-15
Execute Replica/Migration source/dest resource tasks in sequence.

aznashwan:
Due to the current Replica state model, the Replica/Migrations
source/target resource deletion/creation tasks can potentially override
each-other's state.
This patch makes the tasks run in sequence to ensure this does not
happen.


aznashwan:
Tested and works. Merging.

PR #106, opened 2020-01-15
Prevent execution status locking from on_error tasks.

aznashwan:
Due to flawed logic, on_error tasks which were in ERROR/COMPLETED state could prevent the whole execution from being marked as 'ERROR'd.


aznashwan:
Trivial but very important, merging.

PR #108, opened 2020-01-27
Show deleted migrations/replicas

gabriel-samfira:
In an effort to make available the soft-deleted migrations and replicas to the API consumer, this branch also includes various fixes that needed to be implemented before we could achieve the desired result. As such, as part of this PR, there is also a slightly reworked way of returning results from the conductor to the API server, in a way that avoids serializing a potentially huge object, that includes task_info.

### DIFF
--- a/coriolis/api/v1/views/migration_view.py
+++ b/coriolis/api/v1/views/migration_view.py
@@ -27,10 +27,10 @@ def _format_migration(req, migration, keys=None):
     migration_dict = dict(itertools.chain.from_iterable(
         transform(k, v) for k, v in migration.items()))
 
-    if len(migration_dict["executions"]):
-
+    if len(migration_dict.get("executions", [])):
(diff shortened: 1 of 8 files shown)


coriolis/api/v1/views/migration_view.py
```diff
@@ -27,10 +27,14 @@ def transform(key, value):
     migration_dict = dict(itertools.chain.from_iterable(
         transform(k, v) for k, v in migration.items()))
 
-    execution = replica_tasks_execution_view.format_replica_tasks_execution(
-        req, migration_dict["executions"][0])
+    if len(migration_dict["executions"]):
```
aznashwan: NIT: I'd make this a `.get("execution", [])` to be extra safe, though afaik every Migration automatically has an execution.

aznashwan:
LGTM. Merging. Thanks!

PR #112, opened 2020-02-20
Refactor backup writer interfaces

gabriel-samfira:
This change adds the ability for the target provider to dictate which backup writer should be used, based on the target environment itself.

Currently Coriolis only really supports the SSH backup writer, which in some cases is not feasible. All ```migr_target_connection_info``` is currently expected to be SSH connection information. This means that if we want to use the ```HTTPBackupWriter()```, we need to feed in the same connection information to it, and expect it to somehow configure itself automagically using only the SSH connection info.

This change allows import providers to set up the ```HTTPBackupWriter()``` whichever way they want, and only return the relevant certificates and IP:port combination to the export provider. The export provider in turn can simply use the ```BackupWriterFactory()``` to instantiate the proper backup writer, using only the information returned by the import provider, which sets up the necessary resources in advance.

This change, breaks current providers. The providers will need to be updated to account for the following changes:

  * add backup writer factory
  * require that the pkey returned by the providers be a encoded RSA private key.
  * require import providers to declare the backup writer type that should be used

### DIFF
--- a/coriolis/providers/backup_writers.py
+++ b/coriolis/providers/backup_writers.py
@@ -10,6 +10,7 @@ import tempfile
 import threading
 import time
 import uuid
+import shutil
 
 import eventlet
 from oslo_config import cfg
@@ -34,6 +35,15 @@ CONF.register_opts(opts)
 _CORIOLIS_HTTP_WRITER_CMD = "coriolis-writer"
 
 LOG = logging.getLogger(__name__)
+BACKUP_WRITER_SSH = "ssh_backup_writer"
+BACKUP_WRITER_HTTP = "http_backup_writer"
+BACKUP_WRITER_FILE = "file_backup_writer"
+
+BACKUP_WRITERS = [
+    BACKUP_WRITER_SSH,
+    BACKUP_WRITER_HTTP,
+    BACKUP_WRITER_FILE
+]
 
 _WRITER_ERR_MAP = {
     -1: "ERR_MORE_MSG",
@@ -81,6 +91,44 @@ def _disable_lvm2_lvmetad(ssh):
         # that have been synced at least once, will be deactivated.
         utils.ignore_exceptions(utils.exec_ssh_cmd)(
             ssh, "sudo vgchange -an", get_pty=True)
+
+
+class BackupWritersFactory(object):
+
+    def __init__(self, writer_connection_info, volumes_info):
+        self._validate_info(writer_connection_info)
+        self._type = writer_connection_info["backend"]
+        self._conn_info = writer_connection_info["connection_details"]
+        self._volumes_info = volumes_info
+
+    def get_writer(self):
+        if self._type == BACKUP_WRITER_SSH:
+            return SSHBackupWriter.from_connection_info(
+                self._conn_info, self._volumes_info)
+        elif self._type == BACKUP_WRITER_HTTP:
+            return HTTPBackupWriter.from_connection_info(
+                self._conn_info, self._volumes_info)
+        elif self._type == BACKUP_WRITER_FILE:
+            return FileBackupWriter.from_connection_info(
+                self._conn_info, self._volumes_info)
+        raise exception.CoriolisException(
+            "Invalid backup writer type: %s" % self._type)
+
+    def _validate_info(self, info):
+        if type(info) is not dict:
+            raise exception.CoriolisException(
+                "Invalid backup writer connection info.")
+        wrt_type = info.get("backend", None)
+        if wrt_type is None:
+            raise exception.CoriolisException(
+                "Missing backend name in connection info")
+        if wrt_type not in BACKUP_WRITERS:
+            raise exception.CoriolisException(
+                "Invalid backup writer type: %s" % wrt_type)
+        wrt_conn_info = info.get("connection_details")
+        if wrt_conn_info is None:
+            raise exception.CoriolisException(
+                "Missing credentials in connection info")
 
 
 class BaseBackupWriterImpl(with_metaclass(abc.ABCMeta)):
@@ -132,6 +180,11 @@ class BaseBackupWriter(with_metaclass(abc.ABCMeta)):
             if impl:
                 impl.close()
 
+    @classmethod
+    @abc.abstractmethod
+    def from_connection_info(cls, info, volumes_info):
+        pass
+
 
 class FileBackupWriterImpl(BaseBackupWriterImpl):
     def __init__(self, path, disk_id):
@@ -160,6 +213,10 @@ class FileBackupWriterImpl(BaseBackupWriterImpl):
 class FileBackupWriter(BaseBackupWriter):
     def _get_impl(self, path, disk_id):
         return FileBackupWriterImpl(path, disk_id)
+
+    @classmethod
+    def from_connection_info(cls, info, volumes_info):
+        return cls()
 
 
 class SSHBackupWriterImpl(BaseBackupWriterImpl):
@@ -356,6 +413,32 @@ class SSHBackupWriter(BaseBackupWriter):
         self._ssh = None
         self._lock = threading.Lock()
 
+    @classmethod
+    def from_connection_info(cls, info, volumes_info):
+        required = ["ip", "port", "username"]
+        ip = info.get("ip")
+        port = info.get("port")
+        username = info.get("username")
+        pkey = info.get("pkey")
+        password = info.get("password")
+
+        if not all([ip, port, username]):
+            raise exception.CoriolisException(
+                "Connection info is invalid for SSHBackupWriter. "
+                "The following fields are required: %s" % ", ".join(required))
+        if pkey is None and password is None:
+            raise exception.CoriolisException(
+                "Either pkey or password are required")
+
+        if pkey:
+            if type(pkey) is not str:
+                raise exception.CoriolisException(
+                    "pkey must be a PEM encoded RSA private key")
+            pkey = utils.deserialize_key(
+                pkey, CONF.serialization.temp_keypair_password)
+
+        return cls(ip, port, username, pkey, password, volumes_info)
+
     def _get_impl(self, path, disk_id):
         ssh = self._connect_ssh()
         _disable_lvm2_lvmetad(ssh)
@@ -439,6 +522,8 @@ class HTTPBackupWriterImpl(BaseBackupWriterImpl):
         self._compressor_evt = None
 
         self._compress_transfer = compress_transfer
+        if self._compress_transfer is None:
+            self._compress_transfer = CONF.compress_transfers
         super(HTTPBackupWriterImpl, self).__init__(path, disk_id)
 
     def _set_info(self, info):
@@ -540,7 +625,6 @@ class HTTPBackupWriterImpl(BaseBackupWriterImpl):
     def _sender(self):
         while True:
             payload = self._sender_q.get()
-            LOG.debug("Got send payload")
             headers = {
                 "X-Write-Offset": str(payload["offset"]),
                 "X-Client-Token": self._id,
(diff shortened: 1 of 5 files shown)


coriolis/providers/backup_writers.py
```diff
@@ -665,21 +843,49 @@ def _inject_iptables_allow(self, ssh):
...
+            "ca_crt": ca_crt_file,
+        }
+        return self._cert_paths
+
     def _get_impl(self, path, disk_id):
-        ssh = self._connect_ssh()
-        _disable_lvm2_lvmetad(ssh)
-        self._setup_writer(ssh)
+        if self._ssh:
+            _disable_lvm2_lvmetad(self._ssh)
```
aznashwan: Couldn't lvmetad be a problem even if the writer was already setup for us (`else` branch)?
I'd still disable either way for safety.

coriolis/providers/factory.py
```diff
@@ -1,7 +1,8 @@
-# Copyright 2016 Cloudbase Solutions Srl
+# Copyright 2016 Cloudbase :Solutions Srl
```
aznashwan: NIT: accidental added colon

coriolis/providers/factory.py
```diff
@@ -14,6 +15,7 @@
                 help='List of provider class paths'),
 ]
 
+LOG = logging.getLogger(__name__)
```
aznashwan: NIT: this isn't actually used anywhere

coriolis/secrets.py
```diff
@@ -10,6 +10,7 @@
 
 
 def get_secret(ctxt, secret_ref):
+    keystone.create_trust(ctxt)
```
aznashwan: NIT: this could lead to a lot of leftover trusts.
If really needed, I'd recycle this old PR I dumped I a while back:
https://github.com/cloudbase/coriolis/pull/22/files#diff-5a2302b43cadb0681e8b33764c884578R37

coriolis/tasks/replica_tasks.py
```diff
@@ -221,11 +216,11 @@ def run(self, ctxt, instance, origin, destination, task_info,
             migr_connection_info = base.marshal_migr_conn_info(
                 migr_connection_info)
             schemas.validate_value(
-                migr_connection_info,
-                schemas.CORIOLIS_DISK_SYNC_RESOURCES_CONN_INFO_SCHEMA,
-                # NOTE: we avoid raising so that the cleanup task
-                # can [try] to deal with the temporary resources.
-                raise_on_error=False)
+               migr_connection_info,
```
aznashwan: Any reason for the reduced indent here? The whole rest of the file (and everything else in the repo except for the JSON schemas) seems to be using 4.

coriolis/tasks/replica_tasks.py
```diff
@@ -221,11 +216,11 @@ def run(self, ctxt, instance, origin, destination, task_info,
...
+               schemas.CORIOLIS_DISK_SYNC_RESOURCES_CONN_INFO_SCHEMA,
```
aznashwan: NOTE: the schema wasn't actually updated in any way, and is tailored for SSH (`username` is required):
https://github.com/cloudbase/coriolis/blob/master/coriolis/schemas/disk_sync_resources_conn_info_schema.json#L28

coriolis/tasks/replica_tasks.py
```diff
@@ -284,15 +279,12 @@ def run(self, ctxt, instance, origin, destination, task_info,
...
-            schemas.CORIOLIS_DISK_SYNC_RESOURCES_CONN_INFO_SCHEMA,
-            # NOTE: we avoid raising so that the cleanup task
-            # can [try] to deal with the temporary resources.
-            raise_on_error=False)
-
+        try:
+            backup_writers.BackupWritersFactory(
+                migr_connection_info, None).get_writer()
+        except BaseException as err:
+            raise exception.CoriolisException(
```
aznashwan: WARN: Raising here will prevent the `task_info` from getting returned and thus the `DeleteReplicaTargetResourcesTask` from actually cleaning them up:
https://github.com/cloudbase/coriolis/blob/master/coriolis/tasks/replica_tasks.py#L309

coriolis/tasks/replica_tasks.py
gabriel-samfira: Comment/Uncomment command in VScode helped with this one. Comments were added during testing in vim, which was diff from what VSCode does. Fixed.

coriolis/tasks/replica_tasks.py
gabriel-samfira: This change is only relevant for backup writers, which get created by the target, not the source. The source connection info remains unchanged for now, as does the ```os_morphing``` connection info.

ie: OS morphing is still SSH/WinRM specific

coriolis/tasks/replica_tasks.py
gabriel-samfira: Indeed! We should not be raising here. We should let this pass, and fail inside the ```replicate_disks``` method, when the export provider attempts to instantiate the backup writer class using the connection info.

Changed it to ```LOG.exception```

gabriel-samfira:
Thanks for the review!

As it stands, this change is really tied to backup writers (SSH, HTTP and file). Traditionally we used SSH for both OS morphing, transferring disk data and everything else we needed. Until recently, we had nothing else.

So if anyone has any input in regards to how this PR has abstracted away that part, it would be great. It would be desirable, if we come up with a proper formal definition of how these separate concerns should be defined in terms of connection info to their respective workers.

ie:

  * Transfering disk data:
    * How do we create the temporary minions?
    * How should the connection info look like?
    * Do we create JSON schemas, or do we do validation as part of the backup writers themselves?
    * Where should we do validation?
  * OS Morphing:
    * Should we allow room for other means of executing the morphing procedures?
    * Should we explicitly expect SSH for *nix, WinRM for Windows?
  * Source connection info:
    * Do we use it anywhere?
    * How do we use it?
    * Should we limit ourselves to SSH?

coriolis/providers/backup_writers.py
```diff
@@ -83,6 +93,44 @@ def _disable_lvm2_lvmetad(ssh):
...
+class BackupWritersFactory(object):
+
+    def __init__(self, writer_connection_info, volumes_info):
+        self._validate_info(writer_connection_info)
+        self._type = writer_connection_info["backend"]
+        self._conn_info = writer_connection_info["connection_details"]
+        self._volumes_info = volumes_info
+
+    def get_writer(self):
+        if self._type == BACKUP_WRITER_SSH:
```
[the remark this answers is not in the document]
aznashwan: Not much apart from the fact that seeing so many chained `elif`s is more harmful to the eyes than an `return WRITERS_DICT[type](...)`, it's not really a problem anyway.

coriolis/providers/backup_writers.py
```diff
@@ -83,6 +93,44 @@ def _disable_lvm2_lvmetad(ssh):
...
+            return HTTPBackupWriter.from_connection_info(
+                self._conn_info, self._volumes_info)
+        elif self._type == BACKUP_WRITER_FILE:
+            return FileBackupWriter.from_connection_info(
+                self._conn_info, self._volumes_info)
+        raise exception.CoriolisException(
+            "Invalid backup writer type: %s" % self._type)
+
+    def _validate_info(self, info):
+        if type(info) is not dict:
```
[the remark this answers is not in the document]
aznashwan: My bad, was thinking about `isinstance`. (both here and below where I commented the same thing)

My point was that we should try to avoid to directly `==/is` types.

coriolis/providers/backup_writers.py
```diff
@@ -83,6 +93,44 @@ def _disable_lvm2_lvmetad(ssh):
...
+            raise exception.CoriolisException(
+                "Invalid backup writer connection info.")
+        wrt_type = info.get("backend", None)
+        if wrt_type is None:
+            raise exception.CoriolisException(
+                "Missing backend name in connection info")
+        wrt_conn_info = info.get("connection_details")
+        if wrt_conn_info is None:
```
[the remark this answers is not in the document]
aznashwan: I saw, just that I'd expect this `_validate_info` to do all the validation upfront, not do a "surface-level" item check and delegate the rest to later (considering this is the first thing called in the factory, it'd be [marginally] better if it failed sooner rather than later)

coriolis/providers/backup_writers.py
```diff
@@ -628,29 +711,124 @@ def close(self):
...
+        required = ["ip", "port"]
+        if not all([ip, port]):
+            raise exception.CoriolisException(
+                "Missing required connection info: %s" % ", ".join(required))
+
+        ssh_cli = None
+        ssh = conn_info.get("ssh_conn_info")
+        certs = conn_info.get("certificates")
+        if ssh:
+            if type(ssh) is not dict:
```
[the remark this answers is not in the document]
aznashwan: My bad, meant `isinstance()`

coriolis/providers/backup_writers.py
```diff
@@ -628,29 +711,124 @@ def close(self):
...
+        port = conn_info.get("port")
+
+        required = ["ip", "port"]
+        if not all([ip, port]):
+            raise exception.CoriolisException(
+                "Missing required connection info: %s" % ", ".join(required))
+
+        ssh_cli = None
+        ssh = conn_info.get("ssh_conn_info")
+        certs = conn_info.get("certificates")
```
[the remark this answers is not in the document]
aznashwan: I was not saying we should move this to the constructor, I was saying that we should check if both `certs` and `ssh ` was provided and error out here, instead of actually opening the SSH connection here:
https://github.com/cloudbase/coriolis/pull/112/files#diff-ae50c984218b903f95f34c20c8660dbeR818

only to have this exact thing checked and fail here:
https://github.com/cloudbase/coriolis/pull/112/files#diff-ae50c984218b903f95f34c20c8660dbeR731

aznashwan:
Some added general comments:

> Do we create JSON schemas, or do we do validation as part of the backup writers themselves?

IMO the JSON schemas are a nicer way to generalize/decouple the structure of things to have both the caller and callee know what to expect, though I'd certainly leave the added validation in the writers too as it sure can't hurt.

> Source connection info:
>     Do we use it anywhere?
>     Should we limit ourselves to SSH?

You mean the connection to the "reader" which does the chunk reads from the source during PMR?

Afaik every plugin which uses PMR creates the export minion VM and sets up the Replicator directly in the `REPLICATE_DISKS` task due to the `task_info` clobbering issue, so no actual "source_connection_info" info gets transferred across tasks.

Considering I don't see ourselves needing to implement a better Replicator, and that the current Replica class is capable of setting itself up after only being given SSH deets to the minion, I see no problem with leaving it like this.

All the non-PMR plugins mostly don't need any prior setup so there's no reason to worry about passing things across tasks for those (maybe apart from pre-creating backups/snapshots in the `DEPLOY_REPLICA_SOURCE_RESOURCES` tasks once `task_info` clobbering will be fixed)

coriolis/providers/backup_writers.py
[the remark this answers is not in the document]
gabriel-samfira: You are correct. This whole mutually exclusive thing bugs me to be honest, and needs to go. I will split up the ```HTTP``` backup writer service bootstrap to its own class, which will return the proper connection info. That in turn will be used to instantiate the ```HTTPBackupWriter()```.

This way providers that want to use the ```HTTPBackupWriter()``` can choose to either bootstrap the thing using ```userdata``` or call the soon to be written ```HTTPBackupWriterBootstrapper()``` and get proper connection info, and avoid this whole mess.

gabriel-samfira:
The bootstrapping logic for the ```HTTPBackupWriter()``` has been moved to its own class. PTAL.

coriolis/providers/backup_writers.py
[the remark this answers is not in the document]
gabriel-samfira: Okay. Done. Should be a bit more sane. Added benefit is, any provider can now easily bootstrap the ```HTTPBackupWriter()``` by simply calling:

```python
writer_conn_info = HTTPBackupWriterBootstrapper(
    ssh_conn_info, writer_port).setup_writer()
```

aznashwan:
LGTM. Merging. Thanks!

PR #113, opened 2020-02-20
Perform whole migration cleanup in one single task.

aznashwan:
Because the disk deletion task is reused for normal Migrations, there is
a chance it may conflict with the failed deployment cleanup task.
This PR combines the functionality of both in an aggregate task which
executes both steps sequentially.


aznashwan:
Closing this for now as the underlying cause is in the conductor’s on_error task scheduling which requires looking into deeper.

PR #114, opened 2020-03-11
Overhaul task execution loop.

aznashwan:
This PR overhauls the existing task execution loop to better facilitate parallelization of tasks and better ensure the scheduling and non-conflict of cleanup tasks.

Performed changes include:
- addition of several interim states for tasks and executions (ex: SCHEDULED or CANCELLING)
- refactor of task scheduling loop so most state advancement logic is aggregated in a single place instead of hooked into several places
- add checks and safety nets for deadlocked executions (which are still running to some degree, so it still will not help with global reboots)
- revamping the `TaskRunner` interface to only return the pieces of state which they have changed so as to be able to safely run them in parallel
- add checks on Execution definition to ensure no state conflicts can occur


coriolis/worker/rpc/server.py
```diff
@@ -96,7 +96,7 @@ def cancel_task(self, ctxt, task_id, process_id, force):
                     process_id, task_id))
             LOG.error(msg)
             self._rpc_conductor_client.confirm_task_cancellation(
-                ctxt, task_id, cancellation_details=msg)
+                ctxt, task_id, cancellation_details)
```
Dany9966: Undefined `cancellation_details` in this method.

aznashwan: Great catch, thank you!

aznashwan:
Tested (both personally and independently) and works. Merging now.

aznashwan:
NOTE: the UI still needs to have this merged for full compatibility: https://github.com/cloudbase/coriolis-web/pull/521

PR #121, opened 2020-03-30
Do not sanitize action info before saving to DB.

aznashwan:


aznashwan:
Tested and works. Merging.

PR #122, opened 2020-03-30
Add 'normalize_mac_address' utility function for use in providers.

aznashwan:


aznashwan:
Tested and works. Merging.

PR #123, opened 2020-03-31
Make writer key serialization/deserialization more robust.

aznashwan:


coriolis/providers/backup_writers.py
```diff
@@ -93,6 +93,24 @@ def _disable_lvm2_lvmetad(ssh):
             ssh, "sudo vgchange -an", get_pty=True)
 
 
+def _check_deserialize_key(key):
+    res = None
+    if isinstance(key, paramiko.RSAKey):
+        LOG.debug("Key is already in the proper format.")
```
gabriel-samfira: this should be a trace.

coriolis/providers/backup_writers.py
```diff
@@ -93,6 +93,24 @@ def _disable_lvm2_lvmetad(ssh):
...
+        res = key
+    elif type(key) is str:
+        LOG.debug("Deserializing PEM-encoded private key.")
```
gabriel-samfira: idem

coriolis/tasks/replica_tasks.py
```diff
@@ -385,8 +385,19 @@ def _run(self, ctxt, instance, origin, destination, task_info,
                     migr_connection_info, None).get_writer()
             except Exception as err:
                 LOG.warn(
-                    "Seemingly invalid connection info. Replica will likely "
-                    "fail during disk Replication. Error is: %s" % str(err))
+                    "Seemingly invalid backup writer conn info. Replica will "
```
gabriel-samfira: We should probably bite the bullet, err here, catch it in core, and run the cleanup functions in the provider with whatever task info is available up to this point.

aznashwan: I agree it would be more logical to do so here but we'd need:
1) duplicate some code between two task classes which I'd like to avoid
2) by "merging" the behavior of two separate task classes like this we increase the risk of developer/user confusion (ex: what if the cleanup errors out too? which error do we raise and what do we tell the user? what if the user manually cancels the task as soon as they see and error out of impatience and thus accidentallyskips the cleanup?) -- keeping it separate simplifies things a lot IMO

gabriel-samfira: 1. Don't see why:
  * Make a list of cleanup asks we already have in each provider. These are the on_error tasks we already have.
  * make a record of the current task_info we have up to the point of raising the error here
  * short-circuit directly to the cleanup tasks

2. There should be no confusion.
  * The cleanup tasks already look for certain values inside task_info to cleanup. Worker instance ID, networks, security groups etc. 
  * Tasks should be idempotent. Do something there only if needed. Do not fail if current status is desired status (ie: could not find anything to delete. If it's not there, no need to delete, just noop.)
  * You can't guard against PEBKAC. If an operator explicitly cancels a cleanup task, that is either what they want or user error. We can always add a confirmation modal on canceling cleanup tasks, or any tasks for that matter.
  * We raise the initial error. Include any subsequent error raised during cleanup as well. No reason to shadow anything. We can also raise a generic error with a detailed post mortem json including tracebacks or exception information that was collected along the way.

There should be no confusion anywhere. It's simply a matter of how we decide to handle an error that, in this case happens as a result of coriolis core validating the task info, and decising it is in error. Simply logging a warning telling the user: "it will probably fail" is not useful.

It is perfectly understandable to do this in a separate PR though. I just wanted to make sure we add this to our roadmap, and start a discussion.

aznashwan: Sorry for the delayed reply, got distracted with something else.

What you're describing makes sense though it'd involve quite a decent refactor is not much removed from the "individual resource tracking" idea which is already on the roadmap to be worth just partially implementing before we tackle the underlying problem as a whole.

The main problem which needs refactoring is that the 'task result' which is returned by the provider method call is only registered if a task completes successfully (the worker doesn't even bother sending anything in the way of a result on task error/cancellation, just the error message)

In its current state, what'll happen here is that the subsequent REPLICATE_DISKS task will fail [relatively] early and cleanup will get triggered afterwards (DELETE_REPLICA_TARGET_RESOURCES will always be triggered if DEPLOY_REPLICA_TARGET_RESOURCES completes successfully), the check/log message here only helps with indicating that the result the provider returned in DEPLOY_REPLICA_* was borked to begin with, so the provider code is the one at fault and should be reviewed)

That being said, all the points are valid, and when the time does eventually come for the "resource tracking" implementation it'll greatly streamline task execution/cleanup.

gabriel-samfira: Wait, but in this case in particular, the previous task returned successfully. It's us that have decided that the returned task_info is probably invalid. If we error here, we have access to the task info returned. So we could save it and pass it on to the cleanup tasks.

Yes, it would require a refactor, but it is a slight one compared to the incremental task_info updates that's on the roadmap. In any case, you are right. It is worth just swapping out the current way we deal with task info with a more granular option.

Like I said, we don't need to do it as part of this PR, so ignore.

aznashwan:
Tested and works. Merging.

PR #124, opened 2020-04-06
Adapt replicator client to source resource deployment changes

Dany9966:


coriolis/providers/replicator.py
```diff
@@ -519,10 +535,10 @@ def _copy_file(self, ssh, localPath, remotePath):
         except IOError as ex:
```
aznashwan, line 535: We have a `test_ssh_path` utility in the `coriolis.utils` module, could you please use that?

coriolis/providers/replicator.py
```diff
@@ -519,10 +535,10 @@ def _copy_file(self, ssh, localPath, remotePath):
...
             if ex.errno != errno.ENOENT:
                 raise
+        finally:
             sftp.put(localPath, tmp)
```
aznashwan, line 539: Please make double-sure that `sftp.put()` doesn't error if the file already exists

coriolis/utils.py
```diff
@@ -724,6 +724,36 @@ def create_service(ssh, cmdline, svcname, run_as=None, start=True):
...
 
 
+@retry_on_error()
+def restart_service(ssh, svcname):
+    if test_ssh_path(ssh, "/lib/systemd/system"):
+        exec_ssh_cmd(ssh, "sudo systemctl restart %s" % svcname, get_pty=True)
+    elif test_ssh_path(ssh, "/etc/init"):
+        exec_ssh_cmd(ssh, "restart %s" % svcname)
+    else:
+        raise exception.CoriolisException("could not determine init system")
```
aznashwan, line 734: Could you please define a proper `UnrecognizedWorkerInitSystem` exception in the `coriolis.exceptions` module (likely inheriting `NotSupportedOperation`, but you can just inherit from the base `CoriolisException` too if needed) and give it a properly-capitalized default error message.

coriolis/utils.py
```diff
@@ -724,6 +724,36 @@ def create_service(ssh, cmdline, svcname, run_as=None, start=True):
...
+
+
+@retry_on_error()
+def start_service(ssh, svcname):
+    if test_ssh_path(ssh, "/lib/systemd/system"):
+        exec_ssh_cmd(ssh, "sudo systemctl start %s" % svcname, get_pty=True)
+    elif test_ssh_path(ssh, "/etc/init"):
+        exec_ssh_cmd(ssh, "start %s" % svcname)
+    else:
+        raise exception.CoriolisException("could not determine init system")
```
aznashwan, line 744: Ditto

coriolis/utils.py
```diff
@@ -724,6 +724,36 @@ def create_service(ssh, cmdline, svcname, run_as=None, start=True):
...
+
+
+@retry_on_error()
+def stop_service(ssh, svcname):
+    if test_ssh_path(ssh, "/lib/systemd/system"):
+        exec_ssh_cmd(ssh, "sudo systemctl stop %s" % svcname, get_pty=True)
+    elif test_ssh_path(ssh, "/etc/init"):
+        exec_ssh_cmd(ssh, "stop %s" % svcname)
+    else:
+        raise exception.CoriolisException("could not determine init system")
```
aznashwan, line 754: Ditto

coriolis/providers/replicator.py
```diff
@@ -440,6 +440,22 @@ def wait_for_chunks(self):
...
+    def start(self):
+        utils.start_service(self._ssh, REPLICATOR_SVC_NAME)
+
+    def stop(self):
+        utils.stop_service(self._ssh, REPLICATOR_SVC_NAME)
+
+    def restart(self):
+        utils.restart_service(self._ssh, REPLICATOR_SVC_NAME)
+
+    def update_state(self, state, restart=False):
```
aznashwan, line 452: Also, considering that afaik we'll always be wanting the replicator to restart when doing this, should we default `restart=True`? (unless there's a good reason against restarting the Replicator each time?)

coriolis/providers/replicator.py
Dany9966: It doesn't

coriolis/providers/replicator.py
Dany9966: For now, yes, we'd want to always restart. 
NOTE: this method is just a hack for now. In the future, an API endpoint will be added to coriolis-replicator, where you can POST the new state; so this method will use that call instead of manually copying state and restarting the service.

coriolis/providers/replicator.py
```diff
@@ -440,6 +440,22 @@ def wait_for_chunks(self):
...
+        state_file = self._get_replicator_state_file()
```
gabriel-samfira, line 453: This should be done explicitly from outside the replicator, and passed in using the ```state``` parameter. The idea is to allow the person using this function, to supply whichever state, previous or otherwise, they see fit.

The ```state``` parameter should be a ```json``` or a serializable python object that when serialized to json, results in a valid replicator state file. Your choice.

coriolis/providers/replicator.py
gabriel-samfira: Usually, disruptive operations (ie: restarting services), or destructive operations (ie: deleting a file), should be off by default and enabled explicitly.

aznashwan:
LGTM. Merging. Thanks!

PR #128, opened 2020-04-22
Make sure replicator gets started after service creation

Dany9966:


coriolis/providers/replicator.py
```diff
@@ -719,6 +719,7 @@ def _setup_replicator(self, ssh):
         certs = self._setup_certificates(ssh, args)
         self._exec_replicator(
             ssh, args["port"], certs["remote"], REPLICATOR_STATE)
+        self.start()
```
aznashwan, line 722: Is this idempotent? (i.e. are we sure it won't error out if we don't run this if the service is already started?)

Dany9966: > Is this idempotent? (i.e. are we sure it won't error out if we don't run this if the service is already started?)

I tested it: on first replica run, the service is already started before this call, and no error occurs on another start call.

aznashwan: Great, thanks for the confirmation!

aznashwan:
LGTM. Merging. Thanks!

PR #129, opened 2020-05-08
Adds EphemeralVolumesDeleted exception

Dany9966:


Dany9966:
Exception classes are too specific for core. Moving to plugin.

PR #131, opened 2020-05-28
Sort tasks during Replica listing operations.

aznashwan:


aznashwan:
Tested and works. Merging.

PR #132, opened 2020-06-02
Fix run_dracut progress message

Dany9966:


aznashwan:
Trivial, merging, thanks!

PR #135, opened 2020-06-24
Delete source disks

aznashwan:


aznashwan:
Tested and works. Merging now.

PR #136, opened 2020-07-08
Handle any connection validation errors in the worker.

aznashwan:


aznashwan:
Tested and works. Merging.

PR #137, opened 2020-07-22
Move `_enable_oracle_repos` to core

Dany9966:


coriolis/osmorphing/oracle.py
```diff
@@ -21,3 +21,23 @@ def check_os_supported(cls, detected_os_info):
     def _run_dracut(self):
         self._run_dracut_base('kernel')
         self._run_dracut_base('kernel-uek')
+
+    def _enable_oracle_repos(self):
+
+        major_version = self.get_supported_release()
```
aznashwan, line 27: `get_supported_release` is only defined in the OpenStack tools.

Please use `self._detected_os_info['release_version']` or `self._version` instead. (NOTE: please ensure that the code will handle the conversion to `int` itself and that minor releases (e.g. 7.5) will NOT affect the code unless needed)

aznashwan:
LGTM. Merging. Thanks!

PR #139, opened 2020-08-12
Add Scheduler component and Service/Region management.

aznashwan:
This patch adds Region and Service management to Coriolis, as well as a Scheduler service.


aznashwan:
Tested (both personally and independently) and works. Merging now.

PR #141, opened 2020-08-20
Add timeout config option for all internal HTTP requests.

Dany9966:


coriolis/data_transfer.py
```diff
@@ -24,6 +24,9 @@
                     'will be done through this service. This value can be '
                     'either a unix socket path (/var/run/compressor.sock '
                     'or an IP:PORT.'),
+    cfg.IntOpt('default_requests_timeout',
```
aznashwan, line 27: There is no need to register the option multiple times, only once. (unless it's in separate option groups (i.e. you pass `CONF.register_opts(<opts>, "OPTION_GROUP_NAME")`)

Please define this in a separate `coriolis/conf.py` module, and simply make sure that you run a `from coriolis import conf` inside the `__init__.py` of the `coriolis.cmd` package, since all the entry points of the projects are located there.

coriolis/providers/replicator.py
```diff
@@ -34,13 +32,20 @@
 
 DEFAULT_REPLICATOR_PORT = 4433
 
+request_opts = [
+    cfg.IntOpt('default_requests_timeout',
```
aznashwan, line 36: Imo it would still be useful to have a separate `CONF.replicator. default_requests_timeout` option for granularity's sake.

aznashwan (changes requested):
Only nit is that the option only needs to be defined once.

For granularity's sake, I'd recommend you still define an option separately within the `replicator` group which will only be used there.

coriolis/providers/replicator.py
```diff
@@ -38,6 +36,9 @@
     cfg.IntOpt('port',
                default=DEFAULT_REPLICATOR_PORT,
                help='The replicator TCP port.'),
+    cfg.IntOpt('replicator.default_requests_timeout',
```
aznashwan, line 39: The extra `replicator.` prefix is not needed when defining the config options, the group name is passed as a second argument to `CONF.register_opts` below.

aznashwan (changes requested):
The replicator's `default_requests_timeout` option declaration needs fixing.
Also please make a `coriolis-docker` PR to add these new options into the default config.

aznashwan (approved):

aznashwan:
LGTM. Merging. Thanks!

PR #145, opened 2020-08-31
Add Minion Machine Pooling features.

aznashwan:
This PR adds support for pooling of Minion Machines.

Minion Machine pools can be pre-allocated ahead of Migration/Replica operations so as to avoid the time overhead associated with spinning up temporary minion VMs on the source/target platforms.

Current minion pool modelling is done similar to Replicas in that:
* they share the concepts of Executions and Statuses
* the same Conductor task processing logic is used for Minion-pool related tasks
* their operations may be inspected and cancelled by the user
* Migrations/Replicas which use minion pools simply have their respective "DEPLOY/DELETE_*_RESOURCES" tasks replaced with tasks which acquire/attach disks to pre-existing minion machines


aznashwan:
Tested and works. Merging.

PR #147, opened 2020-09-09
Fix barbican secret fetching that uses keystone trust_id

Dany9966:


coriolis/secrets.py
```diff
@@ -1,18 +1,35 @@
...
+    return payload
+
+
+def get_secret(ctxt, secret_ref):
+    payload = None
+
+    try:
+        payload = _get_barbican_secret_payload(ctxt, secret_ref)
+    except keystoneauth1.exceptions.http.Unauthorized:
+        ctxt = copy.deepcopy(ctxt)
```
aznashwan, line 29: Please add a LOG.debug("Error occured while fetching secret with trust ID, retrying without, error was: %s", utils.get_exception_details())

aznashwan:
LGTM. Merging. Thanks!

PR #148, opened 2020-09-10
Improve reliability of task cancellation.

aznashwan:
Account for possibility of worker services failing/refusing/missing cancellation requests while advancing Execution states.


aznashwan:
Tested and works. Merging.

PR #150, opened 2020-09-11
Fix osmorphing minion selection and disk attachment.

aznashwan:


aznashwan:
Tested and works. Merging.

PR #151, opened 2020-09-11
Only delete keystone trust when task goes into a finalized status

Dany9966:


aznashwan:
LGTM. Merging. Thanks!

PR #152, opened 2020-09-14
Fix iterating over osmorphing pool mappings.

herciunichita:


aznashwan:
Trivial. Merging. Thanks!

PR #153, opened 2020-09-14
Fix keystone trusts not deleting bug

Dany9966:


aznashwan:
Trivial. Merging. Thanks!

PR #155, opened 2020-10-07
Added OSMorphing utility method for removing ifcfg file settings on RHELs.

aznashwan:


aznashwan:
Tested and works. Merging.

PR #156, opened 2020-10-08
Log error trace when validating endpoint schema or connection

Dany9966:


aznashwan:
LGTM. Merging. Thanks!

PR #158, opened 2020-10-26
Add 'instance_name' property to instance task export info

Dany9966:


coriolis/schemas/vm_export_info_schema.json
```diff
@@ -18,6 +18,10 @@
       "type": "string",
       "description": "Unique identifier of the VM."
     },
+    "instance_name": {
+      "type": "string",
+      "description": "Human-readable identifier of the VM. Its usage is recommended for platforms that do not use ID as VM identifier."
```
aznashwan, line 23: Can you please rephrase that last part to `It can be optionally used as an alternative to 'id' for identifying instances on platforms which also feature a non-ID naming scheme (e.g. VMWare VM paths)`

aznashwan:
Trivial, merging, thanks!

PR #159, opened 2020-11-03
Switch to using urlsafe base64 encoding

Dany9966:


aznashwan:
Trivial, merging, thanks!

PR #160, opened 2020-11-03
Separate Minion Machine management logic into its own sevice.

aznashwan:
This patchset separates all minion machine-related operations into a dedicated service (the `coriolis-minion-manager`), which  is in charge of the following operations:
* CRUD operations for minion pools
* managing access to the minion-related DB tables
* automatic refreshing and healthchecks of existing minion pools
* providing minion machines for the conductor during Migrations/Replicas

Notable internal improvements/additions to facilitate these operations are:
* ability for worker services to run a task synchronously (still thorough AMQP and with no polling involved)
* definition of a base classes which "adapt" `taskflow`'s task models to Coriolis' `tasks.base.TaskRunner` interface
* addition of logic for running taskflow flows inside a forked process
* all of the "Minion Pool Execution"-related abstractions from the Conductor have been rolled over to the improved `taskflow`-based model within the Minion Manager service


aznashwan:
Tested and works. Merging.

PR #161, opened 2020-11-05
Mount fstab entries in the correct order.

aznashwan:
The entries in fstab used to be mounted in the order they appear in.

This can lead to situations where parent mountpoints later in the file
override child mountpoints earlier in it (e.g. /var/log/mysql can be
overriden by mounting /var/log afterwards)


aznashwan:
Tested and works. Merging.

PR #162, opened 2020-11-12
Fix task deps for DEPLOY_MIGRATION_SOURCE_RESOURCES task.

aznashwan:
Given that the VALIDATE_SOURCE_INPUTS and GET_INSTANCE_INFO tasks run
independently, an extra dependency must be declared for the
DEPLOY_MIGRATION_SOURCE_RESOURCES to the GET_INSTANCE_INFO task to
ensure that the 'export_info' will be available.


aznashwan:
Trivial, merging.

PR #163, opened 2020-11-16
Add Powershell script to set static IPs on Windows VMs

Dany9966:


coriolis/osmorphing/windows.py
```diff
@@ -68,6 +69,39 @@
...
+STATIC_IP_SCRIPT_TEMPLATE = """
+$NICS_INFO = ConvertFrom-Json "%(nics_info)s"
+$IPS_INFO = ConvertFrom-Json "%(ips_info)s"
+
+$i = 0
+foreach ($nic in $NICS_INFO) {
+    $mac_address = $nic.mac_address.replace(':', '-').toUpper()
+    $interface = Get-NetAdapter | Where MacAddress -eq $mac_address
+    if (-Not $interface) {
+        $interface = @(Get-NetAdapter | Where Status -eq Up)[$i]
```
aznashwan, line 84: I don't get this part, is this assuming that the new NIC devices will always be in order on the new machine? What is said order based on? What guarantees do we have that it will be consistent?

Ideally please write a comment somewhere in this script for posterity.

coriolis/osmorphing/windows.py
```diff
@@ -68,6 +69,39 @@
 REQUIRED_DETECTED_WINDOWS_OS_FIELDS = [
     "version_number", "edition_id", "installation_type", "product_name"]
 
+INTERFACES_PATH_FORMAT = (
+    "HKLM:\\%s\\ControlSet001\\Services\\Tcpip\\Parameters\\Interfaces")
+
+STATIC_IP_SCRIPT_TEMPLATE = """
```
aznashwan, line 75: In case Cloudbase-init doesn't do it already, please add `$ErrorActionPreference = "Stop"` at the beginning of the script.

coriolis/osmorphing/windows.py
```diff
@@ -433,8 +467,89 @@ def _install_cloudbase_init(self, download_url,
 
         return cloudbaseinit_base_dir
 
+    def _compile_static_ip_conf_from_registry(self, key_name):
+        ips_info = []
+        interfaces_reg_path = INTERFACES_PATH_FORMAT % key_name
+        interfaces = self._conn.exec_ps_command(
+            "(((Get-ChildItem -Path '%s').Name | Select-String -Pattern "
```
aznashwan, line 474: Please confirm that all of these registry keys exist on the oldest currently supported Windows (2012R2) as well as the latest one by running a test of each so as to cover all bases.

coriolis/osmorphing/windows.py
```diff
@@ -433,8 +467,89 @@ def _install_cloudbase_init(self, download_url,
...
-        pass
+        if dhcp:
+            return
+
+        key_name = str(uuid.uuid4())
+        self._load_registry_hive(
+            "HKLM\\%s" % key_name,
+            "%sWindows\\System32\\config\\SYSTEM" % self._os_root_dir)
+        try:
+            cbslinit_base_dir = "%sCloudbase-Init" % self._os_root_dir
```
aznashwan, line 546: Please abstract this as a method/constant as it's used through the rest of this module too and we should avoid "desyncs".

coriolis/osmorphing/windows.py
```diff
@@ -433,8 +467,89 @@ def _install_cloudbase_init(self, download_url,
...
+            raise exception.OSMorphingException(
+                "The following IP(s): %s found on the source VM's NICs were "
+                "not found in the registry. Please check whether the static "
+                "IP configurations on the source machine are properly set up "
+                "and retry the migration." % diff)
+
+    def _write_static_ip_script(self, base_dir, nics_info, ips_info):
+        scripts_dir = self._get_cbslinit_scripts_dir(base_dir)
+        script_path = "%s\\01-static-ip-config.ps1" % scripts_dir
+        contents = STATIC_IP_SCRIPT_TEMPLATE % {'nics_info': nics_info,
```
aznashwan, line 533: Just formatting the JSONs into the script itself could be risky.
Please either:
1) b64encode before formatting and b64decode within the script (ideal)
2) write the JSON in a separate file and have the script read that file's contents.

coriolis/osmorphing/windows.py
```diff
@@ -68,6 +69,39 @@
...
+            New-NetIPAddress -InterfaceIndex $interface.ifIndex -IPAddress $ip_info.ip_address -PrefixLength $ip_info.prefix_length -DefaultGateway $ip_info.default_gateway
+        } else {
+            New-NetIPAddress -InterfaceIndex $interface.ifIndex -IPAddress $ip_info.ip_address -PrefixLength $ip_info.prefix_length
+        }
+
+        if ($ip_info.dns_addresses) {
+            Set-DnsClientServerAddress -InterfaceIndex $interface.ifIndex -ServerAddresses $ip_info.dns_addresses
+        }
+    }
+}
```
aznashwan, line 102: How/where will errors in the script be handled/logged?
If it won't be in a logical place (e.g. in the Cloudbase-init logs), please add a separate try-catch within this script and write the error to a file or something.

coriolis/osmorphing/windows.py
```diff
@@ -68,6 +69,39 @@
...
+$i = 0
+foreach ($nic in $NICS_INFO) {
+    $mac_address = $nic.mac_address.replace(':', '-').toUpper()
+    $interface = Get-NetAdapter | Where MacAddress -eq $mac_address
+    if (-Not $interface) {
+        $interface = @(Get-NetAdapter | Where Status -eq Up)[$i]
+    }
+    $i++
+
+    Remove-NetIPAddress -InterfaceIndex $interface.ifIndex -Confirm:$false
```
aznashwan, line 88: Please apply the script structuring standards specific to powershell (i.e. split the above and the below logic within their own well-defined functions, and put the "main() logic" (i.e. the JSON decodes and calls to said functions) at the bottom.)

aznashwan (changes requested):
Some technical and stylistic concerns were raised need addressing please.

coriolis/osmorphing/windows.py
```diff
@@ -68,6 +69,39 @@
 REQUIRED_DETECTED_WINDOWS_OS_FIELDS = [
     "version_number", "edition_id", "installation_type", "product_name"]
 
+INTERFACES_PATH_FORMAT = (
+    "HKLM:\\%s\\ControlSet001\\Services\\Tcpip\\Parameters\\Interfaces")
+
+STATIC_IP_SCRIPT_TEMPLATE = """
+$NICS_INFO = ConvertFrom-Json "%(nics_info)s"
```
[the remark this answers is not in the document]
Dany9966: I tested the script on a 2012R2, so I expected it to be supported on the later versions too. Anyway, it turns out the command works on all actively supported Windows OSes.

coriolis/osmorphing/windows.py
Dany9966: This is the proposed backdoor logic. If we can't find the NIC by MAC address, the interface will be selected by the nics_info order.

coriolis/osmorphing/windows.py
Dany9966: The errors are logged in `Cloudbase-Init/Logs` dir.

coriolis/osmorphing/windows.py
aznashwan: This unfortunately has the potential to screw things up harder when our assumption is wrong than any "gains" we get from it being right.

We could hypothetically add better "identification logic" in the future (by exporting more info from the source platform like source network range/gateway) to make more educated matches.

Until then, I'd suggest we simply drop this and just log out somehwere that no matching interface was found for the MAC address (requiring MACs to be preserved seems like a more than reasonable enough requirement IMO)

aznashwan:
LGTM. Merging. Thanks!

PR #164, opened 2020-11-20
Add 'user_scripts' column to migration table

Dany9966:


coriolis/db/sqlalchemy/migrate_repo/versions/017_adds_user_scripts_to_migration.py
```diff
@@ -0,0 +1,17 @@
...
+import sqlalchemy
+
+
+def upgrade(migrate_engine):
+    meta = sqlalchemy.MetaData()
+    meta.bind = migrate_engine
+
+    # add 'user_scripts' column to 'migration':
+    migration = sqlalchemy.Table(
+        'migration', meta, autoload=True)
```
aznashwan, line 13: Please add the column to the base_transfer_action table so we can have Replicas “remember” their user script settings too so users won’t need to specify them each Replica deployment.

aznashwan (changes requested):
Please also add some validation of the user script’s format in api/v1/{migrations,replicas}.py.

Checks should include:
1) if the user script it specified, it must be a dict
2) every instance ID from the user _scripts dict must be present in “instances” (though NOT necessarily vice-versa)
3) each user script is b64decode’d properly

coriolis/db/sqlalchemy/migrate_repo/versions/017_adds_user_scripts_to_migration.py
Dany9966: It is possible to add this to replicas too, but only after a first deployment has been done. Currently, there's no way to add user_scripts when creating a replica, is this what is actually needed here?

coriolis/api/v1/replicas.py
```diff
@@ -171,12 +176,31 @@ def _update_storage_mappings(original_storage_mappings,
...
         return storage_mappings
 
+    @staticmethod
+    def _update_user_scripts(original_user_scripts, new_user_scripts):
+        global_scripts = original_user_scripts.get('global', {})
+        new_global_scripts = new_user_scripts.get('global', {})
+        if new_global_scripts:
+            global_scripts.update(new_global_scripts)
+
+        instance_scripts = original_user_scripts.get('instances', {})
```
aznashwan, line 186: We should ideally provide a way to "unset" a certain user script.
By convention, let's say that is someone posts a `null` or an empty string, that script should be unset.

aznashwan (commented):
First review pass done. Minor comments provided.

coriolis/api/v1/replicas.py
Dany9966: It already does that

coriolis/api/v1/utils.py
```diff
@@ -85,3 +86,31 @@ def _build_keyerror_message(resource, method, key):
             resource, method_mapping[method], key)
 
     return msg
+
+
+def validate_user_scripts(user_scripts):
+    if not isinstance(user_scripts, dict):
+        raise exception.InvalidInput(
+            reason='"user_scripts" must be of JSON object format')
```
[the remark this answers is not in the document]
Dany9966: This does not check whether you pass instances key or not. It actually checks whether user_scripts is a valid dict (i.e. the user passes {"global": blabla, "instances": blablabla} )

coriolis/db/sqlalchemy/migrate_repo/versions/019_adds_user_scripts_column.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2020 Cloudbase Solutions Srl
```
aznashwan, line 1: This should be 18 not 19. Please just revert it back to 17 and I'll make the other update 18.

aznashwan (approved):

aznashwan:
Merging as-is and changing the DB migration script order later.

PR #167, opened 2020-12-20
Optimize RPC client instantiations in conductor.

aznashwan:


aznashwan:
Trivial, merging.

coriolis/conductor/rpc/server.py
```diff
@@ -196,23 +196,35 @@ def inner():
...
+    def _worker_client(self):
+        if not getattr(self, '_worker_client_instance'):
+            self._worker_client_instance = (
+                rpc_worker_client.WorkerClient())
+        return self._worker_client_instance
 
     @property
     def _scheduler_client(self):
-        return rpc_scheduler_client.SchedulerClient()
+        if not getattr(self, '_scheduler_client_instance'):
```
gabriel-samfira, line 217: This is only useful if you re-use this instance of the client, and you don't create a new one using ```WorkerClient.from_service_definition()```.

PR #168, opened 2020-12-21
Ensure RPC call connection cleanup across all components.

aznashwan:
Due to some underlying issue related to the usage/implementation of the
oslo_messaging.RPCClient, cross-components connecions are not getting
properly closed, leading to numerous hanging connections and eventual
saturation of all ports available on the Rabbit host.
    
This PR adds a base class in the `coriolis.rpc` module which facilitates
automatic cleanup for all RPC calls.


aznashwan:
Tested and works. Merging.

PR #169, opened 2021-01-19
Fix leaking RPC clients

gabriel-samfira:


gabriel-samfira:
Will be fixed in a later PR

PR #170, opened 2021-01-25
Add --worker-process-count argument to relevant components.

aznashwan:
This patch adds a new command line argument ('--worker-process-count')
to the components which need it (API, Conductor, and Worker) to allow
for CLI-based control on the number of worker processes spawned.

Note that this will NOT stop the components from forking if they are
programmed to do so, but merely stops oslo_service from forking the main
process at initial startup.


aznashwan:
Tested and works. Merging.

PR #172, opened 2021-02-03
Reconfigure cloud-init to retain original user credentials

Dany9966:


coriolis/osmorphing/base.py
```diff
@@ -342,3 +343,28 @@ def _set_selinux_autorelabel(self):
                 "touch /.autorelabel")
         except Exception as err:
             LOG.warning("Failed to set autorelabel: %r" % err)
+
+    def _retain_user_credentials(self):
```
aznashwan, line 347: This method name is pretty uninformative.
Please rename it to `_configure_cloud_init_user_retention` or something clearly indicating it's only cloud-init related.

coriolis/osmorphing/base.py
```diff
@@ -342,3 +343,28 @@ def _set_selinux_autorelabel(self):
...
+        cloud_cfg_path = "/etc/cloud/cloud.cfg"
```
aznashwan, line 348: Considering it's theoretically possible for these things we care about to be defined within separate files in '/etc/cloud/cloud.cfg.d/*.cfg', please make this method iterate through all files and apply the modifications to all of them.

coriolis/osmorphing/manager.py
```diff
@@ -194,7 +194,7 @@ def morph_image(origin_provider, destination_provider, connection_info,
 
     import_os_morphing_tools = import_os_morphing_tools_cls(
         conn, os_root_dir, os_root_dev, hypervisor_type,
-        event_manager, detected_os_info)
+        event_manager, detected_os_info, osmorphing_info)
```
aznashwan, line 197: Note that the Windows tools have not been updated so you'll get a guaranteed `TypeError` here for Windows morphers.

aznashwan (changes requested):
Some important lapses need to be addressed.

Additionally, the current version will never work with Windows.

coriolis/osmorphing/base.py
```diff
@@ -342,3 +343,28 @@ def _set_selinux_autorelabel(self):
...
+        self._event_manager.progress_update(
+            "Reconfiguring cloud-init to retain original user credentials")
+        try:
+            self._exec_cmd_chroot(
+                "cp %s %s.bak" % (cloud_cfg_path, cloud_cfg_path))
+
+            cloud_cfg = yaml.load(self._read_file(cloud_cfg_path),
+                                  Loader=yaml.SafeLoader)
+
+            cloud_cfg['disable_root'] = 0
```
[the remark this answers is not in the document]
Dany9966: Yes. 0 worked fine from my testing. Although I could switch to 'True' and 'False' to be more safe.

coriolis/osmorphing/base.py
```diff
@@ -342,3 +343,28 @@ def _set_selinux_autorelabel(self):
...
+            cloud_cfg['ssh_pwauth'] = 1
```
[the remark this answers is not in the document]
Dany9966: I tested this with 'dvincze' user, and also a 'cloudbase' user. No password got reset. I'm thinking maybe because we enable password login though SSH, cloud-init will stop reseting them. I can check that also.

aznashwan:
LGTM. Merging now. Thanks!

PR #173, opened 2021-02-05
Filter out minion devices matching 'root=' kernel argument

Dany9966:


coriolis/osmorphing/osmount/base.py
```diff
@@ -309,6 +309,22 @@ def _get_mounted_devices(self):
...
+            "cat /proc/cmdline | grep -oe 'root=[^ ]*'").decode().splitlines()
+        _, root_token = kernel_root_arg[0].split('=', 1)
+        if root_token.startswith('UUID') or (
+                root_token.startswith('PARTUUID')) or (
+                root_token.startswith('LABEL')):
+            cmd = 'blkid -o device --match-token %s' % root_token
+            root_devs = self._exec_cmd(cmd).decode().splitlines()
+            ret += root_devs
+        elif root_token.startswith('/dev'):
+            if root_token not in ret:
```
gabriel-samfira, line 323: There is a chance that this will hold a value like: ```/dev/root```, which has the same major:minor number as ```/dev/vda1``` for example (essentially pointing to the same device). You still need to identify the disk, and make sure you add any other device alias in ```ret```. Get the major:minor number of the device, and make sure you add all other block devices that point to the same physical device.

gabriel-samfira: Also, the UUID/label may be an unreliable way to exclude the minion disks from consideration. If both the VM we are migrating and the minion have the same base image, there is a high chance the UUID and the label will be identical. The UUID will differ only if the minion uses a different image. The UUID will survive a partition resize.

gabriel-samfira (approved):

aznashwan:
LGTM. Merging. Thanks!

PR #174, opened 2021-02-15
Prevent minion reuse error for Replica deployments.

aznashwan:
Ensure that the minion manager does not bother trying to reuse
destination transfer minions for Replica deployments or if no transfer
minions were requested to begin with.


aznashwan:
Trivial. Merging.

PR #178, opened 2021-02-18
Make task-related worker-conductor interactions synchronous.

aznashwan:
This prevents possible desync problems when it comes to operations like
confirming task running or process setting.


aznashwan:
Trivial but necessary. Merging.

PR #179, opened 2021-02-22
Minion healthcheck fix and unmount all devs when looking for guest contents

Dany9966:


aznashwan:
LGTM. Merging. Thanks!

PR #181, opened 2021-03-10
Fix paramiko command channel hang

Dany9966:


coriolis/exception.py
```diff
@@ -470,3 +468,27 @@ class FailedPackageUninstallationException(PackageManagerOperationException):
...
+    pass
+
+
+class OSMorphingSSHOperationTimeout(OSMorphingOperationTimeout):
+    message = (
+        "Pending SSH command %(cmd)s timed out after %(timeout)s seconds. "
+        "Coriolis may have encountered connection issues to the minion machine"
+        " or the command execution time exceeds the timeout set. Try extending"
+        " the timeout by editing the 'default_osmorphing_operation_timeouts' "
+        "in /etc/coriolis/coriolis.conf."
```
aznashwan, line 483: NIT: please no explicit paths like `/etc/coriolis/coriolis.conf` since those are not necessarily set in stone.

Please use "Coriolis' static configuration file" or similar.

coriolis/exception.py
```diff
@@ -470,3 +468,27 @@ class FailedPackageUninstallationException(PackageManagerOperationException):
...
+    )
+
+
+class OSMorphingWinRMOperationTimeout(OSMorphingOperationTimeout):
+    message = (
+        "Pending WinRM command %(cmd)s timed out after %(timeout)s seconds. "
+        "Coriolis may have encountered connection issues to the minion machine"
+        " or the command execution time exceeds the timeout set. Try extending"
+        " the timeout by editing the 'default_osmorphing_operation_timeouts' "
+        "in /etc/coriolis/coriolis.conf."
```
aznashwan, line 493: Same as above please.

coriolis/osmorphing/base.py
```diff
@@ -120,10 +122,11 @@ class BaseLinuxOSMorphingTools(BaseOSMorphingTools):
     _packages = {}
 
     def __init__(self, conn, os_root_dir, os_root_dev, hypervisor,
-                 event_manager, detected_os_info, osmorphing_parameters):
+                 event_manager, detected_os_info, osmorphing_parameters,
+                 operation_timeout):
```
aznashwan, line 126: This runs the risk of some provider tools which override `__init__` to fail to call `super().__init__`, could you please make `operation_timeout` a kwarg which defaults to `None` or `-1` (which, by convention, we can consider to be no timeout)?

Please ensure the actual command execution calls can handle a `timeout=None` (most notably wsman.py)

coriolis/utils.py
```diff
@@ -306,10 +307,13 @@ def exec_ssh_cmd(ssh, cmd, environment=None, get_pty=False):
         "environment %s: '%s'", remote_str, environment, cmd)
 
     _, stdout, stderr = ssh.exec_command(
-        cmd, environment=environment, get_pty=get_pty)
+        cmd, environment=environment, get_pty=get_pty, timeout=timeout)
+    try:
+        std_out = stdout.read()
+        std_err = stderr.read()
+    except socket.timeout:
+        raise exception.OSMorphingSSHOperationTimeout(cmd=cmd, timeout=timeout)
```
aznashwan, line 315: `OSMorphingSSHOperationTimeout` is too specific an exception to raise here since this method can be also used to run non-OSMorphing commands.

Please make a separate `MinionMachineCommandTimeout` exception class and raise that here, and make the base `OSMorphingOperationTimeout` inherits from.
Afterwards, in the OSMorphing tools classes, use `try-except MinionMachineCommandTimeout: raise OSMorphing(SSH|WinRM)OperationTimeout(str(ex)) from ex`

coriolis/wsman.py
```diff
@@ -86,9 +88,16 @@ def from_connection_info(cls, connection_info):
     def disconnect(self):
         self._protocol = None
 
+    def set_timeout(self, timeout):
+        if timeout > 0:
+            self._protocol.transport.timeout = timeout
```
aznashwan, line 93: Has this been tested and works? (I'm asking because in the `protocol.Protocol` constructor the `transport` is merely a string)

coriolis/wsman.py
```diff
@@ -20,15 +21,16 @@
...
         self._protocol = None
+        self._conn_timeout = timeout
 
     EOL = "\r\n"
 
     @utils.retry_on_error()
     def connect(self, url, username, auth=None, password=None,
                 cert_pem=None, cert_key_pem=None):
-        protocol.Protocol.DEFAULT_TIMEOUT = 3600
+        protocol.Protocol.DEFAULT_TIMEOUT = self._conn_timeout
```
aznashwan, line 33: Considering you found that we can set `protocol.Protocol().transport.timeout` I'd remove this static setting completely and just set that after the `self._protocol` declaration below.

coriolis/osmorphing/base.py
```diff
@@ -256,14 +259,19 @@ def _list_dir(self, chroot_path):
         path = os.path.join(self._os_root_dir, chroot_path)
         return utils.list_ssh_dir(self._ssh, path)
 
-    def _exec_cmd(self, cmd):
+    def _exec_cmd(self, cmd, timeout=0):
+        if timeout is not None:
+            timeout = timeout or self._osmorphing_operation_timeout
```
aznashwan, line 264: This line will never allow timeout to be `0`, thus making the above `timeout is not None` check redundant.

We must convene that either:
* timeout of `0` is valid and settable, and thus should just make the default kwarg `None` and just write `if timeout is None: timeout = self._osmorphing_operation_timeout`
* timeout of `0` is semantically the same as `None` (i.e. no timeout), so we can just write it like `if not timeout: timeout = self._osmorphing_operation_timeout`

coriolis/osmorphing/base.py
```diff
@@ -256,14 +259,19 @@ def _list_dir(self, chroot_path):
...
         return utils.exec_ssh_cmd(
-            self._ssh, cmd, environment=self._environment, get_pty=True)
+            self._ssh, cmd, environment=self._environment, get_pty=True,
+            timeout=timeout)
 
-    def _exec_cmd_chroot(self, cmd):
+    def _exec_cmd_chroot(self, cmd, timeout=0):
+        if timeout is not None:
```
aznashwan, line 270: Same as above.

aznashwan (changes requested):
Minor comment on exception classes and the semantics of `None` vs `0` timeouts.

coriolis/osmorphing/base.py
Dany9966: We can't really consider `None` the global equivalent of "no timeout", because pywinrm doesn't accept None or 0 as timeout, only > 0 allowed. `paramiko` on the other side accepts both.

coriolis/osmorphing/base.py
Dany9966: I don't think we want to ever set paramiko channel timeout to 0. Usually, for no timeout, we should set the timeout to `None`. This way, the channels will be in 'blocking' mode - this is what we used before (more info here: https://github.com/paramiko/paramiko/blob/5cf145ffd3d7cd4d96a511376267a7f27cce5eed/paramiko/channel.py#L588).
I wanted to make a logic where `_exec_cmd` would have a way to disable the channel timeout (by setting it to `None`), and also ignore `self._osmorphing_operation_timeout`. If you don't think that's needed, I can just do what you said above.

coriolis/wsman.py
Dany9966: Tested and works. Only acceptable values are ints bigger than 0. No way to disable the timeout.

coriolis/osmorphing/base.py
aznashwan: In that case we add an extra check for WSMan to always default to an internal 1h or something if nothing is provided.

Point is we must decide on a reasonable default for this constructor kwarg to not have the provider tools potentially fail for not setting any.

As mentioned in a different comment, we can convene to either have "None" mean "no timeout", or always demand a numeric input (with values <= 0 counting as no timeout), but we need to be consistent about it.

coriolis/osmorphing/base.py
aznashwan: Anything works as long as we are internally consistent.

From my PoV having both `None` and `0` represent no timeout is excessive, though I'd recommend we make the kwarg itself here default to `None` to be clear that no timeout is set (a timeout of `0` could either mean no timeout ever or expecting each operation to be instantaneous so confusion might arise there)

coriolis/wsman.py
```diff
@@ -49,7 +50,7 @@ def connect(self, url, username, auth=None, password=None,
             cert_key_pem=cert_key_pem)
 
     @classmethod
-    def from_connection_info(cls, connection_info):
+    def from_connection_info(cls, connection_info, timeout):
```
aznashwan, line 53: This has the potential of breaking everywhere it's called. Please make it a kwarg which defaults to `DEFAULT_TIMEOUT`.

coriolis/osmorphing/manager.py
```diff
@@ -13,6 +13,13 @@
 from coriolis.osmorphing.osmount import factory as osmount_factory
 from coriolis.osmorphing.osdetect import manager as osdetect_manager
 
+
+opts = [
+    cfg.IntOpt('default_osmorphing_operation_timeouts',
```
aznashwan, line 18: NIT: `timeouts` should be singular. If anything should be plural it's `operations`, but the general `operation` should be fine IMO.

coriolis/wsman.py
```diff
@@ -86,9 +87,17 @@ def from_connection_info(cls, connection_info):
     def disconnect(self):
         self._protocol = None
 
+    def set_timeout(self, timeout):
+        if timeout > 0:
```
aznashwan, line 91: This bears the risk of of a `TypeError` if `None` somehow gets passed here so please either:
a) account for a timeout of `None` meaning "no timeout at all"
b) add a check for `timeout is not None`, but in that case a simple `if timeout:` would be semantically identical.

aznashwan (changes requested):
Almost there. Just a few minor observations.

coriolis/exception.py
```diff
@@ -470,3 +468,31 @@ class FailedPackageUninstallationException(PackageManagerOperationException):
...
+class OSMorphingOperationTimeout(MinionMachineCommandTimeout):
+    pass
+
+
+class OSMorphingSSHOperationTimeout(OSMorphingOperationTimeout):
+    message = (
+        "Pending SSH command %(cmd)s timed out after %(timeout)s seconds. "
+        "Coriolis may have encountered connection issues to the minion machine"
+        " or the command execution time exceeds the timeout set. Try extending"
+        " the timeout by editing the 'default_osmorphing_operation_timeouts' "
```
aznashwan, line 486: `timeout` singular

aznashwan (approved):

aznashwan:
LGTM. Merging. Thanks.

PR #183, opened 2021-03-16
Update DB layer for SQLAlchemy 1.4.

aznashwan:
Changes are:
    - switching base_transfer_action.info base sqlalchemy type from
      to `LongBinary` since plain `Binary` no longer exists.
      This should not affect existing deployments since the underlying
      SQL type is still `longblob`
    - use `filter` instead of `filter_by` for `max` queries.


aznashwan:
Tested and works. Merging.

PR #184, opened 2021-03-17
Check yum repositories before enabling them

Dany9966:


coriolis/osmorphing/oracle.py
```diff
@@ -22,6 +22,14 @@ def _run_dracut(self):
         self._run_dracut_base('kernel')
         self._run_dracut_base('kernel-uek')
 
+    def _find_oracle_repos(self, repos_to_enable):
```
aznashwan, line 25: This appears to generally apply to any RHEL-based system so please move it to the base RHEL tools class

coriolis/osmorphing/oracle.py
```diff
@@ -22,6 +22,14 @@ def _run_dracut(self):
...
+        for repo in repos_to_enable:
+            cmd = 'egrep "^\[.*%s.*\]$" -R /etc/yum.repos.d | cut -f2 -d:'
+            available_repos = self._exec_cmd_chroot(
+                cmd % repo).decode().splitlines()
+            self._enable_repos += [
```
aznashwan, line 30: Having "magical" class attributes only set during calls to certain methods is bad IMO.

I know this was here before, but can you please refactor the `_enable_oracle_repos` to be called in `install_packages` (both here, in the base RHEL tools, as well as checking all the provider OSMorphing tools for this.

aznashwan (changes requested):

coriolis/osmorphing/oracle.py
```diff
@@ -34,10 +34,15 @@ def _enable_oracle_repos(self):
...
+            repos = self._find_yum_repos(repos_to_enable)
         else:
             self._yum_install(['oraclelinux-release-el%s' % major_version])
+            repos_to_enable = ["ol%s_appstream" % major_version,
+                               "ol%s_UEKR6" % major_version]
+            found_repos = self._find_yum_repos(repos_to_enable)
             self._exec_cmd_chroot(
-                "yum config-manager --enable ol%(release)s_appstream "
-                "ol%(release)s_UEKR6" % {"release": major_version})
+                "yum config-manager --enable %s" % " ".join(found_repos))
```
aznashwan, line 46: Looking through some of the providers it seems that the repos returned by this method are also passed to `_yum_install` (and thus temporarily enabled during the actual package installation, which is preferred to enabling them outright here IMO to prevent modifying the VMs as much as possible during OSMorphing)

Is there any reason/counterexample to still keep the actual enablement command here?

If not, I'd like to request you have this method only return the repo names (and rename the method to something like `_get_oracle_repos`.

coriolis/osmorphing/oracle.py
chenzhouzhao: how to  install on linux  server？  so  like  centos  Ubuntu  ，&& suse

aznashwan (approved):

aznashwan:
LGTM. Merging. Thanks!

PR #185, opened 2021-03-29
Skip replication when there are no changed chunks

Dany9966:


coriolis/providers/replicator.py
```diff
@@ -795,27 +795,27 @@ def replicate_disks(self, source_volumes_info, backup_writer):
                 self._event_manager.progress_update(
                     "No new chunks to replicate for disk \"%s\" (%s)" % (
                         volume['disk_id'], devName))
```
aznashwan, line 797: Why not just add the following two lines here?
```python
self._repl_state = curr_state
return self._repl_state
```

aznashwan (changes requested):
Can be simplified.

coriolis/providers/replicator.py
Dany9966: Logically the same, but can do that too.

aznashwan:
LGTM. Merging. Thanks!

PR #186, opened 2021-03-30
[WiP] Prevent API service from resetting RPC connections.

aznashwan:
This PR prevents the Coriolis API service from resetting RPC transports
every time like the other components since in the RPC service's case
there is no "internal" forking involved and instead the service processes
are created on startup.


aznashwan:
Superseded by #189

PR #187, opened 2021-04-01
Avoid returning None for method `_get_symlink_target`

Dany9966:


coriolis/osmorphing/osmount/base.py
```diff
@@ -289,12 +289,12 @@ def _check_mount_fstab_partitions(
         return new_mountpoints
 
     def _get_symlink_target(self, symlink):
-        target = None
         try:
             target = self._exec_cmd('readlink -en %s' % symlink).decode()
             LOG.debug("readlink %s returned: %s" % (symlink, target))
         except Exception:
             LOG.warn('Target not found for symlink: %s' % symlink)
+            target = symlink
```
aznashwan, line 297: Please just move this line before the `try` to make it more readable and avoid any possible scoping issues.

Also, please mention in the `LOG.warn` that the original link path will be returned.

aznashwan:
LGTM. Merging. Thanks!

PR #189, opened 2021-04-01
Add version requirement for oslo_messaging and kombu.

aznashwan:
The latest versions oslo_messaging and kombu cause exterme RAM usage in
the API service. Coriolis can be ruled out as the cause as individual
testing showed kombu to the single biggest factor.
This PR mitigates this by pinning both to their prior major releases.


aznashwan:
Tested and works. Merging.

PR #191, opened 2021-04-08
Ensure on-error tasks with exclusively on-error parents are started.

aznashwan:
Due to a logical lapsus, on-error tasks which only depended on other
on-error tasks were unscheduled by the conductor.
(most notable being the RESTORE_REPLICA_DISKS_SNAPSHOTS task)


aznashwan:
Tested and works. Merging.

PR #195, opened 2021-04-22
Add default value to user_scripts column

Dany9966:
Necessary when upgrading from older releases


aznashwan:
All LGTM. Merging, Thanks!

PR #197, opened 2021-05-11
Use diskpart to bring disks offline

Dany9966:


coriolis/osmorphing/osmount/windows.py
```diff
@@ -81,6 +84,8 @@ def _service_disks_with_status(
             "Servicing disks with status '%s' (%s) from disk list: %s",
             status, servicable_disk_ids, disk_list)
         for disk_id in servicable_disk_ids:
+            if disk_id in disk_ids_to_skip:
```
aznashwan, line 87: Do please add a log message here so we know when/if a disk gets skipped.

coriolis/osmorphing/osmount/windows.py
```diff
@@ -105,6 +110,30 @@ def _service_disks_with_status(
                             raise
                     break
 
+    def _get_disk_ids_from_drive_letters(self, drive_letters):
+        disk_ids = []
+        volume_entry_re = r"\s+Volume ([0-9]+)\s+([A-Z]+)\s+"
+        volume_list_script = "LIST VOLUME\r\nEXIT"
+        vol_list = self._run_diskpart_script(volume_list_script)
```
aznashwan, line 117: Small nit but please move these two declarations right before the `for volume in vol_list...` since that's where they're used.

coriolis/osmorphing/osmount/windows.py
```diff
@@ -105,6 +110,30 @@ def _service_disks_with_status(
...
+
+        def get_disk_ids(vol_id):
```
aznashwan, line 119: Stylistic yet ultimately pointless nit: please prefix the function name with `_` since it technically counts as "private to this scope"

coriolis/osmorphing/osmount/windows.py
```diff
@@ -105,6 +110,30 @@ def _service_disks_with_status(
                             raise
                     break
 
+    def _get_disk_ids_from_drive_letters(self, drive_letters):
+        disk_ids = []
+        volume_entry_re = r"\s+Volume ([0-9]+)\s+([A-Z]+)\s+"
```
aznashwan, line 115: Nit: in such cases where we're after a very specific match it's always best to make regexes as specific as possible.
Considering the 'Label' in the `LIST VOLUME` can be whatever it wants, please use a leading '^' for the regex so we always catch the beginning of the line.

coriolis/osmorphing/osmount/windows.py
```diff
@@ -105,6 +110,30 @@ def _service_disks_with_status(
...
+        volume_list_script = "LIST VOLUME\r\nEXIT"
+        vol_list = self._run_diskpart_script(volume_list_script)
+
+        def get_disk_ids(vol_id):
+            vol_detail_script = (
+                "SELECT VOLUME %s\r\nDETAIL VOLUME\r\nEXIT" % vol_id)
+            vol_details = self._run_diskpart_script(vol_detail_script)
+            vol_disk_re = r"\*\s+Disk ([0-9]+)\s+"
```
aznashwan, line 123: Leading '^' for the regex.

coriolis/osmorphing/osmount/windows.py
```diff
@@ -149,11 +178,16 @@ def _set_basic_disks_rw_mode(self):
...
-        self._conn.exec_ps_command(
-            "Get-Volume |? DriveLetter -eq \"%s\" | Get-Partition | "
-            "Get-Disk | Set-Disk -IsOffline $True" % drive_letter,
-            ignore_stdout=True)
+    def _bring_disks_offline(self, drives_to_skip=None):
+        disk_ids_to_skip = None
+        if drives_to_skip:
+            disk_ids_to_skip = self._get_disk_ids_from_drive_letters(
+                drives_to_skip)
+        offline_disk_script_fmt = "SELECT DISK %s\r\nOFFLINE DISK\r\nEXIT"
```
aznashwan, line 186: Iirc if we have Dynamic Disks, then bringing one of the physical disks which comprise the dynamic disk offline will (rightfully) bring all the others offline.
In this sense, please double-check that running `OFFLINE DISK` on an already offline disk won't lead to an error code.

coriolis/osmorphing/windows.py
```diff
@@ -426,24 +426,24 @@ def _write_cloudbase_init_conf(self, cloudbaseinit_base_dir,
...
-            "logfile = cloudbase-init.log\n"
-            "default_log_levels = "
-            "comtypes=INFO,suds=INFO,iso8601=WARN,requests=WARN\n"
-            "allow_reboot = false\n"
-            "plugins = %(plugins)s\n"
-            "debug = true\n"
-            "san_policy = OnlineAll\n"
-            "metadata_services = %(metadata_services)s\n"
-            "logging_serial_port_settings = %(com_port)s,9600,N,8\n" %
+            "[DEFAULT]\r\n"
```
aznashwan, line 429: Although I can understand the confusion of `Notepad.exe` users when opening non-CRLF files, is there any functional reason for Cloudbase-init to need the `\r`s?

coriolis/utils.py
```diff
@@ -276,7 +276,7 @@ def write_winrm_file(conn, remote_path, content, overwrite=True):
             data = data.encode()
         asb64 = base64.b64encode(data).decode()
         cmd = ("$ErrorActionPreference = 'Stop';"
-               "$x = [System.IO.FileStream]::new(\"%s\", "
+               "$x = New-Object System.IO.FileStream(\"%s\", "
```
aznashwan, line 279: Out of curiosity, is there any behavioral difference for this change?

aznashwan (commented):
Mostly LGTM with the exception of some minor nits.

coriolis/osmorphing/osmount/windows.py
Dany9966: I've tested this with a source VM that has actual spanned dynamic disks, and it didn't error out. Will double-check

coriolis/osmorphing/windows.py
Dany9966: No functional reason, besides beautifying. Figured we could write CRLF text files since it's standard on Windows.

coriolis/utils.py
Dany9966: This didn't work on Powershell 4.0. We have PSv4 on the WS2012R2 image from our site.

aznashwan:
LGTM. Merging. Thanks!

PR #199, opened 2021-05-19
Fix schema loading for Jinja2 3.0+.

aznashwan:


aznashwan:
Tested and works with Jinja2 both below and above 3.0.

PR #204, opened 2021-06-23
Refresh volume group metadata upon activation

Dany9966:
In some cases, upon volume group activation, its logical volumes' partitions don't get recognized on the minion, only after a metadata refresh


aznashwan:
LGTM. Merging. Thanks!

PR #207, opened 2021-07-06
Futureproof setup.cfg by replacing dashes with underscores.

aznashwan:
`setuptools` will deprecate usage of dashes in the metadata fields from
setup.cfg.


aznashwan:
Trivival, merging,

PR #208, opened 2021-08-16
Only load libraries of required plugin during task execution.

aznashwan:


aznashwan:
Tested and works. Merging.

PR #209, opened 2021-08-25
Enable non-interactive package installation for Debian-based tools

Dany9966:


aznashwan:
LGTM. Merging. Thanks!

PR #213, opened 2021-09-22
Fix mock module import syntax

smiclea:


aznashwan:
LGTM. Merging, Thanks!

PR #218, opened 2021-10-11
Fix memory leak in RPC clients

gabriel-samfira:
This PR fixes a memory leak in the RPC clients. Transport is only created once, and saved inside the instance of the class.

### DIFF
--- a/coriolis/rpc.py
+++ b/coriolis/rpc.py
@@ -66,7 +66,7 @@ def get_server(target, endpoints, serializer=None):
                                     serializer=serializer)
 
 
-def init(reconnect=False):
+def init():
     global _TRANSPORT
     if _TRANSPORT is None:
         _TRANSPORT = _get_transport()


coriolis/rpc.py
```diff
@@ -65,6 +66,13 @@ def get_server(target, endpoints, serializer=None):
                                     serializer=serializer)
 
 
+def init(reconnect=False):
```
aznashwan, line 69: Note that at one point there were issues relating to the worker service class forking with the RPC clients already initialized. (symptom was that all messages sent from the worker task processes (the processes we fork ourselves) would get a MessageTimeout).

If this works in a multi-process worker setting however, then there must have been some internal changes to oslo_messaging to allow for forking so all should be well.

https://github.com/cloudbase/coriolis/blob/master/coriolis/worker/rpc/server.py#L55-L65

gabriel-samfira: That's why init_rpc is ```False``` for the worker, and defaults to ```True``` for non-forking services (all the rest). The downside is that wor the worker, every thread that tries to communicate via RPC will create a new transport which will be cached in the RPC client class. This is fine, as the worker does not spawn many threads. All work is done in a subprocess. The API and conductor create many threads and benefit the most from this change.

See:

https://github.com/cloudbase/coriolis/pull/218/files#diff-aad4a476daa8e8f3405ebdb0dd4c7824fbb722aee7f4e4c2eb27f6d28efc8513R34

That ```reconnect``` parameter is not used. Will remove.

gabriel-samfira: This function:

https://github.com/cloudbase/coriolis/pull/218/files#diff-d5681a671184b0caff615870a5631792f3ef362d3deb6315b13f657af34be0aaR94

will use the transport stored in the global variable, or create a new one which will be cached in the RPC class. Workers will not initialize the global variable, so any forked multiprocess will create its own transport for each RPC client and use that.

PR #221, opened 2021-10-27
Consolidate read-only API access policies

Dany9966:
This patch will set some default policies for project reader and member users.
Reader users will have access to read-only API operations (i.e. list, show).
Write-specific API operations (i.e. create, delete, update) will require
at least project member role for users.


aznashwan (approved):

aznashwan:
LGTM. Merging. Thanks!

PR #222, opened 2021-11-10
Check replica volumes exist before raising storage mapping update error

Dany9966:
Any storage mapping update automatically fails, without checking whether the `volumes_info` actually exists or not. `volumes_info` argument wasn't used up to this point, this may be the main reason we kept passing it.


aznashwan:
LGTM. Merging. Thanks!

PR #224, opened 2021-11-15
Fix yum repos bugs

Dany9966:
This PR includes the following fixes:

1. Filter files ending in '.repo' when searching for yum repos
    
    yum internally only works with repos that reside in files ending
    with '.repo'. This patch should make sure that Coriolis only looks
    for repos inside files with filenames ending in '.repo'.

2. Manually configure yum repos instead of using yum-config-manager
    
    `yum-config-manager --add-repo` automatically adds and enables the
    repos from a URL. This being against our ideal OSMorphing
    functionality, this patch will manually download the repo file
    and should also make sure to disable all downloaded repos.

### DIFF
--- a/coriolis/osmorphing/oracle.py
+++ b/coriolis/osmorphing/oracle.py
@@ -1,6 +1,8 @@
 # Copyright 2016 Cloudbase Solutions Srl
 # All Rights Reserved.
 
+import uuid
+
 from coriolis.osmorphing import redhat
 from coriolis.osmorphing.osdetect import oracle as oracle_detect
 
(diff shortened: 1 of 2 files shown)


coriolis/osmorphing/oracle.py
```diff
@@ -26,15 +26,15 @@ def _get_oracle_repos(self):
...
         if major_version < 8:
-            self._yum_install(
-                ['yum-utils'],
-                self._find_yum_repos(['ol%s_latest' % major_version]))
-            # TODO(apilotti): for ULN users, use the corresponding repos
-            # e.g.: ol7_x86_64_addons
+            repo_file_path = (
+                '/etc/yum.repos.d/new-public-yum-ol%s.repo' % major_version)
+            self._exec_cmd_chroot(
+                "curl -L http://public-yum.oracle.com/public-yum-ol%s.repo "
```
aznashwan, line 32: I'm assuming this would be equivalent to the `ol%s_latest` repo we `yum install`d previously, but `yum-utils` no longer gets installed.

Are we sure none of the providers expect something from `yum-utils` to be available?

coriolis/osmorphing/redhat.py
```diff
@@ -219,7 +219,9 @@ def _yum_clean_all(self):
     def _find_yum_repos(self, repos_to_enable=[]):
         found_repos = []
         for repo in repos_to_enable:
-            cmd = 'egrep "^\[.*%s.*\]$" -R /etc/yum.repos.d | cut -f2 -d:'
+            cmd = (
+                'egrep "^\[.*%s.*\]$" -R /etc/yum.repos.d | '
```
aznashwan, line 223: This is a pretty underhanded way of getting the repos we want.

Presuming that a repo named `X` will always be in a file named `X.repo`, a more logical course of action imo would be to first search for `X.repo` in `/etc/yum.repos.d` (ideally through Python and not with `find` or `ls | grep`) and then process the file we found.

Presuming the repo can be in a file not named `X.repo`, it'd be better to first "parse" all the files in `/etc/yum.repos.d` and record all the repos and their files, then check for all the `repos_to_enable` you'll want in that.

aznashwan (changes requested):
We should generally strive to minimize our reliance on shell commands during OSMount/OSMorphing in situations we can avoid it (e.g. plaintext config file loading/processing)

coriolis/osmorphing/oracle.py
Dany9966: We don't modify yum in any way (as we should), so yum-utils is not needed for providers. Even if that was the case, this `yum-utils` installation only occured for OL guests. And yes, it is the same repo URL as the former.

coriolis/osmorphing/redhat.py
Dany9966: Your presumption is not necessarily correct. A `.repo` file can (and ususally does) store information for tens and hundreds of repos. Here we need to go through all the `.repo` files, fetch all declared repo names, and then check whether everything from `repos_to_enable` list is found.

aznashwan:
LGTM. Merging. Thanks!

aznashwan (approved):

PR #227, opened 2022-02-16
Remove migration `status` from API response view

Dany9966:
`last_execution_status` key from DB should be used instead.


aznashwan:
LGTM. Merging. Thanks!

PR #245, opened 2023-01-20
Ensure packages for backup writer are installed

Dany9966:
Before using packages like `iptables`, we will need to make sure these packages are installed first, especially on newer worker operating systems, where some classic tools are being replaced.


Dany9966:
Replaced  by: https://github.com/cloudbase/coriolis/pull/246

PR #250, opened 2023-02-09
Add unit test for `migrate_instances`

smiclea:
Tests the `migrate_instances` function of the `ConductorServerEndpoint` class.


coriolis/tests/conductor/rpc/test_server.py
```diff
@@ -1556,3 +1556,282 @@ def create_task_side_effect(
...
+            ]
+
+        instances = [
+            mock.sentinel.instance_1,
+            mock.sentinel.instance_2,
+        ]
+        instance_osmorphing_minion_pool_mappings = {}
+        # use morphing pool mappings only
+        # if origin and destination minion pools are used,
+        # this way we can test the case when we don't use any minion pools
```
Dany9966, line 1665: Why wouldn't/couldn't we also test osmorphing pools without having origin/destination pools? You can configure this separately in your yml file, add a `has_osmorphing_pool`, and check that instead of either the origin or destination pools

coriolis/tests/conductor/rpc/data/migrate_instances_config.yml
```diff
@@ -0,0 +1,221 @@
...
+      depends_on: ['DEPLOY_INSTANCE_RESOURCES']
+    - type: CLEANUP_FAILED_INSTANCE_DEPLOYMENT
+      depends_on: ['DEPLOY_INSTANCE_RESOURCES', 'FINALIZE_INSTANCE_DEPLOYMENT']
+      on_error_only: True
+    - type: CLEANUP_INSTANCE_TARGET_STORAGE
+      depends_on: ['CREATE_INSTANCE_DISKS', 'CLEANUP_INSTANCE_SOURCE_STORAGE', 'DELETE_MIGRATION_TARGET_RESOURCES', 'CLEANUP_FAILED_INSTANCE_DEPLOYMENT']
+      on_error_only: True
+
+- config:
+    skip_os_morphing: False
```
Dany9966, line 149: After adding that `has_os_morphing_pool` option, I'd prefer if we'd have the following extra cases:
- `skip_os_morphing: False`, `has_os_morphing_pool: False`, `has_origin/destination_pool: True`;
- `skip_os_morphing: False`, `has_os_morphing_pool: True`, `has_origin/destination_pool: True`;
- `skip_os_morphing: False`, `has_os_morphing_pool: True`, `has_origin/destination_pool: False`;
- `has_os_morphing_pool: True`  (implicitly, `skip_os_morphing` is `True` here)

Dany9966 (changes requested):

coriolis/tests/conductor/rpc/test_server.py
smiclea: I initially considered skipping tests specific to the osmorphing pool for some reason. However, I have now updated the test to address this issue.

coriolis/tests/conductor/rpc/data/migrate_instances_config.yml
smiclea: The new cases have been added.

smiclea:
@Dany9966 The issues have been addressed.

PR #257, opened 2023-05-01
Add unit tests for Coriolis Scheduler RPC Server

smiclea:


coriolis/tests/scheduler/rpc/test_server.py
```diff
@@ -0,0 +1,78 @@
...
+@ddt.ddt
+class SchedulerServerEndpointTestCase(test_base.CoriolisBaseTestCase):
+    """Test suite for the Coriolis Scheduler Worker RPC server."""
+
+    def setUp(self):
+        super(SchedulerServerEndpointTestCase, self).setUp()
+        self.server = server.SchedulerServerEndpoint()
+
+    @mock.patch.object(db_api, "get_regions")
+    @mock.patch.object(db_api, "get_services")
```
Dany9966, line 24: I don't feel like mocking these here makes sense in the unit test scenario. I get what you tried doing here, but it'd make more sense to test `get_workers_for_specs` as a "unit", so please mock `_get_all_worker_services` instead, and move the `get_services` mock in future a `test_get_all_worker_services` case. In this particular case you should really just check `_get_all_worker_services` is called. Apply the same for the rest of the methods (mock `_filter_regions`, etc.)

Dany9966 (changes requested):
Please add independent test cases for each of the methods (public or not). Please also make sure to do the best assertions that make sense for the particular test case.

coriolis/tests/scheduler/rpc/test_server.py
```diff
@@ -0,0 +1,78 @@
...
+        mock_get_services.return_value = config_obj.services_db or []
+        mock_get_regions.return_value = config_obj.regions_db or []
+
+        kwargs = {
+            "enabled": enabled,
+            "region_sets": region_sets,
+            "provider_requirements": provider_requirements,
+        }
+        if expected_exception:
+            exception_mapping = {
```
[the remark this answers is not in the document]
smiclea: Fixed.

coriolis/tests/scheduler/rpc/test_server.py
smiclea: Makes sense. Fixed. Though this means that some of the exceptions in the config yaml need to be tested in different unit tests when testing the relevant functions, so those test configurations are removed for now.

smiclea:
@Dany9966 All comments have been resolved.

smiclea:
@Dany9966 `test_get_workers_for_specs` function name renamed from `test_get_workers_for_specs_refactored`

coriolis/tests/scheduler/rpc/test_server.py
```diff
@@ -0,0 +1,98 @@
...
+            for region_set in region_sets:
+                mock_filter_regions.assert_any_call(
+                    mock.sentinel.context,
+                    region_set,
+                    enabled=True,
+                    check_all_exist=True
+                )
+
+        mock_get_weighted_filtered_services.assert_called_once_with(
+            mock_get_all_worker_services.return_value, mock.ANY
```
Dany9966, line 93: Since we're not gonna build up `filters` to pass here, let's at least mock the `trivial_filters` calls. Therefore, please do the following:
1. if `enabled is not None`: check `EnabledFilter` called with `enabled`
2. on each `region_set`: check `RegionsFilter` called with `region_set, any_region=True`
3. if `provider_requirements`: check `ProviderTypesFilter` called with `provider_requirements`

Dany9966 (changes requested):
Almost there!

coriolis/tests/scheduler/rpc/test_server.py
smiclea: @Dany9966 Changes made.

Dany9966 (approved):
LGTM

PR #259, opened 2023-05-14
Redact unnecessary `volumes_info` keys in errors

smiclea:
The `volumes_info` errors present in replica tasks are now redacted. As a result, the `chunks` field is no longer fully displayed.

### DIFF
--- a/coriolis/tasks/replica_tasks.py
+++ b/coriolis/tasks/replica_tasks.py
@@ -31,24 +31,35 @@ def _check_ensure_volumes_info_ordering(export_info, volumes_info):
     instance = export_info.get(
         'instance_name',
         export_info.get('name', export_info['id']))
+
+    vol_info_cpy = utils.sanitize_task_info(
+        {"volumes_info": volumes_info}).get("volumes_info", [])
+
     ordered_volumes_info = []
     for disk in export_info['devices']['disks']:
         disk_id = disk['id']
         matching_volumes = [
             vol for vol in volumes_info if vol['disk_id'] == disk_id]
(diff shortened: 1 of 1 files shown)


coriolis/tasks/replica_tasks.py
```diff
@@ -39,17 +43,14 @@ def _check_ensure_volumes_info_ordering(export_info, volumes_info):
...
             raise exception.InvalidActionTasksExecutionState(
                 "Could not find source disk '%s' (ID '%s') in Replica "
-                "volumes info: %s" % (disk, disk_id, volumes_info))
+                "volumes info: %s" % (disk, disk_id, vol_info_cpy))
         elif len(matching_volumes) > 1:
             raise exception.InvalidActionTasksExecutionState(
-                "Multiple disks with ID '%s' foind in Replica "
-                "volumes info: %s" % (disk_id, volumes_info))
+                "Multiple disks with ID '%s' found in Replica "
+                "volumes info: %s" % (disk_id, vol_info_cpy))
```
Dany9966, line 50: Same here, keep the message, change to `LOG.error()`;
Raise exception with message:
```python
raise exception.InvalidActionTasksExecutionState(
    "Multiple disks with ID '%s' found in Replica volumes info. Please check that the instance doesn't have the "
    "same volume attached twice, or whether there's a UUID collision between its disks. Check the logs "
    "for more information." % disk_id)
```

Dany9966 (changes requested):

smiclea:
@Dany9966 Changes done.

PR #266, opened 2023-09-28
added a try-block to catch the SSH error

OneBuildMan:
The purpose of this PR is to provide a more user-friendly error messages when SSH exceptions are encountered. For example, instead of printing a generic message like "Authentication failed" it's better to provide more context by indicating that the error is related to the SSH client. Also, an optional timeout was added to wait for the SSH banner to be presented.

### DIFF
--- a/coriolis/providers/replicator.py
+++ b/coriolis/providers/replicator.py
@@ -476,7 +476,7 @@ class Replicator(object):
             return ssh
         except paramiko.ssh_exception.SSHException as ex:
             raise exception.CoriolisException(
-                "Failed to setup SSH client: %s" % str(ex))
+                "Failed to setup SSH client: %s" % str(ex)) from ex
 
     def _parse_source_ssh_conn_info(self, conn_info):
         # if we get valid SSH connection info we can
--- a/coriolis/providers/replicator.py
+++ b/coriolis/providers/replicator.py
@@ -16,7 +16,7 @@ import requests
 
 from coriolis import exception
 from coriolis import utils
-
+from coriolis.exception import NotAuthorized
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


coriolis/providers/replicator.py
```diff
@@ -16,7 +16,7 @@
 
 from coriolis import exception
 from coriolis import utils
-
+from coriolis.exception import NotAuthorized
```
alexpilotti, line 19: You can’t import a class name. As per guidelines, it’s:

coriolis/providers/replicator.py
```diff
@@ -470,10 +470,13 @@ def _get_ssh_client(self, args):
...
-        ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-        ssh.connect(**args)
-        return ssh
+        try:
+            ssh = paramiko.SSHClient()
+            ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+            ssh.connect(**args)
+            return ssh
+        except paramiko.ssh_exception.SSHException as ex:
+            raise NotAuthorized("Failed to setup SSH client: %s" % str(ex))
```
alexpilotti, line 479: Why would any ssh exception be related to authorization issues?

Dany9966: I asked him to pick this one (I had the "Authentication failed" message in mind), but you're right, we should use the more generic `CoriolisException` one. The gist of this PR is to present a more friendly message when SSH exceptions occur. For example, "Authentication failed." just printed out in the task failure message is pretty vague, we need to specify this is a SSH client error. 

Therefore, @OneBuildMan please fix the import statement above, and change the exception type to a `CoriolisException`.

coriolis/providers/replicator.py
```diff
@@ -470,10 +469,14 @@ def _get_ssh_client(self, args):
...
-        ssh.connect(**args)
-        return ssh
+        try:
+            ssh = paramiko.SSHClient()
+            ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+            ssh.connect(**args)
+            return ssh
+        except paramiko.ssh_exception.SSHException as ex:
+            raise exception.CoriolisException(
+                "Failed to setup SSH client: %s" % str(ex))
```
Dany9966, line 479: Also add a `from ex` to the raise, so we don't lose the original trace

Dany9966 (changes requested):

Dany9966 (approved):

PR #275, opened 2023-11-08
Fix pep8 cache unit test

Cristi1324:
This PR fixes cache unit test pep8 error.

### DIFF
--- a/coriolis/tests/test_cache.py
+++ b/coriolis/tests/test_cache.py
@@ -2,8 +2,9 @@
 # All Rights Reserved.
 
 from unittest import mock
-from coriolis import exception
+
 from coriolis import cache
+from coriolis import exception
 from coriolis.tests import test_base
 
 


coriolis/tests/test_cache.py
```diff
@@ -1,10 +1,10 @@
 # Copyright 2023 Cloudbase Solutions Srl
 # All Rights Reserved.
 
-from unittest import mock
-from coriolis import exception
 from coriolis import cache
+from coriolis import exception
 from coriolis.tests import test_base
+from unittest import mock
```
Dany9966, line 7: I think the pep8 issue here is actually the fact that there was no free line between Python builtin module imports and third-party ones. Please try the following:

```python
from unittest import mock

from coriolis import cache
```

Dany9966 (changes requested):

PR #277, opened 2023-11-10
Add unit tests for the `context.py` module

mihaelabalutoiu:
This PR adds unit tests for the `context.py` module.

### DIFF
--- a/coriolis/tests/test_context.py
+++ b/coriolis/tests/test_context.py
@@ -1,0 +1,101 @@
+# Copyright 2023 Cloudbase Solutions Srl
+# All Rights Reserved.
+
+from unittest import mock
+
+from coriolis import context
+from coriolis import exception
+from coriolis.tests import test_base
+from coriolis.tests import testutils
+
+
+class RequestContextTestCase(test_base.CoriolisBaseTestCase):
+    """Test suite for the Coriolis RequestContext class."""
+
+    def setUp(self):
+        super(RequestContextTestCase, self).setUp()
+        self.req_context = context.RequestContext(
+            user="user",
+            project_id='test_project_id',
+        )
+
+    @mock.patch.object(context.timeutils, 'utcnow')
+    def test__init__(self, mock_utcnow):
+        context_req = testutils.get_wrapped_function(context.RequestContext)(
+            user="user",
+            project_id='test_project_id',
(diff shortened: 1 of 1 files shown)


coriolis/tests/test_context.py
```diff
@@ -0,0 +1,60 @@
...
+        self.assertEqual(new_context.roles, [])
+
+    def test_to_dict(self):
+        result = self.req_context.to_dict()
+
+        self.assertEqual(result['timestamp'],
+                         self.req_context.timestamp.isoformat())
+
+    def test_from_dict(self):
+        result = self.req_context.to_dict()
```
Dany9966, line 53: Here you don't need to check whether the timestamp is handled, you already do that in your `__init__` test. I think you should either:
1. Check that the class is called with `**result`
2. Remove the test case altogether. It's fine if `from_dict` is not covered, it doesn't really do much.

Dany9966 (changes requested):
Mostly LGTM

coriolis/tests/test_context.py
mihaelabalutoiu: Gotcha :) Thank you!

mihaelabalutoiu:
@Dany9966 I updated the code, please take a look now, thank you!

coriolis/tests/test_context.py
```diff
@@ -0,0 +1,74 @@
...
+    def test_can_policy_not_authorized(self, mock_check_policy):
+        mock_check_policy.side_effect = exception.PolicyNotAuthorized(
+            action='test_action')
+
+        self.assertRaises(exception.PolicyNotAuthorized,
+                          self.req_context.can, 'test_action')
+
+        mock_check_policy.assert_called_once_with(
+            self.req_context, 'test_action', {'project_id': 'test_project_id',
+                                              'user_id': 'user'})
```
Dany9966, line 74: Please also add a test case where you pass a custom `target`, and also `fatal=False`

Dany9966 (changes requested):
Aside from some additions, LGTM

mihaelabalutoiu:
> Also please add test for `get_admin_context`

Done,  I made the changes, as well. Thank you!

Dany9966 (approved):

PR #279, opened 2023-11-17
Add unit testing for the `data_tranfer.py` module

mihaelabalutoiu:
This PR adds unit tests for the `data_tranfer.py` module.

### DIFF
--- a/coriolis/tests/test_data_transfer.py
+++ b/coriolis/tests/test_data_transfer.py
@@ -1,0 +1,165 @@
+# Copyright 2023 Cloudbase Solutions Srl
+# All Rights Reserved.
+
+import stat
+import struct
+from unittest import mock
+from urllib import parse
+
+import requests
+import requests_unixsocket
+
+from coriolis import data_transfer
+from coriolis import exception
+from coriolis.tests import test_base
+
+
+class DataTranferTestCase(test_base.CoriolisBaseTestCase):
+    """Collection of tests for the Coriolis data transfer module."""
+
+    def setUp(self):
+        super(DataTranferTestCase, self).setUp()
+        self.data_content = 'test-content'.encode()
+        self.fmt = 'gzip'
+        self.mock_session = mock.Mock()
+        self.mock_url = 'http://localhost:8000/'
+        self.mock_response = mock.Mock()
+        self.mock_response.content = self.data_content
+        self.mock_session.post.return_value = self.mock_response
+        self.msg_id = 1
+        self.path = 'test-path'
+        self.offset = 0
+
+    @mock.patch.object(data_transfer, 'CONF')
+    def test_get_session_and_address(self, mock_conf):
+        mock_conf.compressor_address = 'localhost:8000'
+        result = data_transfer._get_session_and_address()
+
+        self.assertIsInstance(result[0], requests.Session)
+        self.assertEqual(result[1], "http://%s/" %
+                         mock_conf.compressor_address)
+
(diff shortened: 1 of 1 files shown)


coriolis/tests/test_data_transfer.py
```diff
@@ -0,0 +1,152 @@
...
+    @mock.patch.object(data_transfer.os, 'stat')
+    @mock.patch.object(data_transfer, 'CONF')
+    def test_get_session_and_address_invalid_unix_socket(self, mock_conf,
+                                                         mock_stat):
+        mock_conf.compressor_address = '/var/run/compressor.sock'
+        mock_stat.return_value.st_mode = stat.S_IFREG
+
+        self.assertRaises(exception.CoriolisException,
+                          data_transfer._get_session_and_address)
+
```
Dany9966, line 64: The following `_get_session_and_address` branches are not explicitly covered: 
https://github.com/cloudbase/coriolis/blob/master/coriolis/data_transfer.py#L41
https://github.com/cloudbase/coriolis/blob/master/coriolis/data_transfer.py#L56

Please add cases for these too

Dany9966: The first can be done by literally just calling the function and checking the result is `(None, None)`

The second should be done by NOT mocking `os.stat`, therefore `os.path.exists` will always return `False`, because the mocked `compressor_address` does not in fact exist on the disk

coriolis/tests/test_data_transfer.py
```diff
@@ -0,0 +1,152 @@
...
+        mock_get_session_and_address.return_value = (
+            self.mock_session, self.mock_url)
+        result = data_transfer.compression_proxy(self.data_content, self.fmt)
+
+        self.mock_session.post.assert_called_once_with(
+            self.mock_url, data=self.data_content,
+            headers={'X-Compression-Format': self.fmt},
+            timeout=mock_conf.default_requests_timeout)
+        self.assertEqual(
+            result, (self.mock_session.post.return_value.content, False))
```
Dany9966, line 93: Additionally, please check that `self.mock_session.close` is called once, thus making sure the session gets closed after the request is sent

coriolis/tests/test_data_transfer.py
```diff
@@ -0,0 +1,152 @@
...
+    @mock.patch.object(data_transfer, '_COMPRESS_FUNC')
+    def test_compression_proxy_with_exception(self, mock_compress_func,
+                                              mock_get_session_and_address):
+        mock_get_session_and_address.return_value = (self.mock_session,
+                                                     self.mock_url)
+        self.mock_session.post.side_effect = Exception("mock_message")
+
+        result = data_transfer.compression_proxy(self.data_content, self.fmt)
+        self.assertEqual(
+            result, (mock_compress_func[self.fmt].return_value, True))
```
Dany9966, line 105: Ditto: session close check

Dany9966 (changes requested):

coriolis/tests/test_data_transfer.py
```diff
@@ -0,0 +1,152 @@
...
+# All Rights Reserved.
+
+import stat
+import struct
+
+import requests
+import requests_unixsocket
+
+from unittest import mock
+from urllib import parse
```
Dany9966, line 11: These two imports should be grouped in the first group (where stdlib modules go). Please move these under `import struct`

mihaelabalutoiu:
I made the changes. Please take a look now. Thank you!

PR #290, opened 2024-01-26
Add unit tests for `coriolis/scheduler` and `coriolis/services` modules

mihaelabalutoiu:
This PR adds unit tests for the following modules:
* `services/api.py`
* `scheduler/scheduler_utils.py`
* `scheduler/rpc/client.py`
* `scheduler/rpc/server.py`
* `scheduler/filters/base.py`
* `scheduler/filters/trivial_filters.py`

### DIFF
--- a/coriolis/tests/scheduler/filters/test_trivial_filters.py
+++ b/coriolis/tests/scheduler/filters/test_trivial_filters.py
@@ -1,0 +1,137 @@
+# Copyright 2024 Cloudbase Solutions Srl
+# All Rights Reserved.
+
+from unittest import mock
+
+import ddt
+
+from coriolis.scheduler.filters import trivial_filters
+from coriolis.tests import test_base
+
+
+@ddt.ddt
+class RegionsFilterTestCase(test_base.CoriolisBaseTestCase):
+    """Test suite for the RegionsFilter class."""
+
+    def setUp(self):
+        super(RegionsFilterTestCase, self).setUp()
+        self.regions = [mock.sentinel.region1, mock.sentinel.region2]
+        self.any_region = False
+        self.regions_filter = trivial_filters.RegionsFilter(
+            self.regions, self.any_region)
+
+    def test__repr__(self):
+        result = repr(self.regions_filter)
+
+        self.assertEqual(result, "<RegionsFilter(regions=%s, "
+                                 "any_region=%s)>" % (self.regions,
+                                                      self.any_region))
+
+    @ddt.data(
+        (None, [mock.sentinel.region1, mock.sentinel.region2], 100, False),
+        ([mock.sentinel.region1, mock.sentinel.region2],
+            [mock.sentinel.region1, mock.sentinel.region2], 100, False),
+        ([mock.sentinel.region1, mock.sentinel.region2], [], 0, False),
+        ([mock.sentinel.region1, mock.sentinel.region2],
+            [mock.sentinel.region1], 0, False),
+        ([mock.sentinel.region1, mock.sentinel.region2],
+            [mock.sentinel.region1], 100, True)
+    )
+    @ddt.unpack
+    def test_rate_service(self, regions, mapped_regions, expected, any_region):
+        self.regions_filter._regions = regions
+        self.regions_filter._any_region = any_region
+        mock_service = mock.Mock()
+        mock_service.mapped_regions = [mock.Mock(id=region) for region in
+                                       mapped_regions]
+
+        result = self.regions_filter.rate_service(mock_service)
+
+        self.assertEqual(result, expected)
+
+
+class TopicFilterTestCase(test_base.CoriolisBaseTestCase):
+    """Test suite for the TopicFilter class."""
+
+    def setUp(self):
+        super(TopicFilterTestCase, self).setUp()
+        self.service = mock.Mock()
+        self.topic = mock.sentinel.topic
+        self.topic_filter = trivial_filters.TopicFilter(self.topic)
+
+    def test__repr__(self):
+        result = repr(self.topic_filter)
+        self.assertEqual(result, "<TopicFilter(topic=%s)>" % self.topic)
+
+    def test_rate_service(self):
+        self.service.topic = self.topic
+
+        result = self.topic_filter.rate_service(self.service)
+        self.assertEqual(result, 100)
+
+    def test_rate_service_false(self):
+        self.service.topic = mock.sentinel.other_topic
+
+        result = self.topic_filter.rate_service(self.service)
+        self.assertEqual(result, 0)
+
+
+class EnabledFilterTestCase(test_base.CoriolisBaseTestCase):
+    """Test suite for the EnabledFilter class."""
+
+    def setUp(self):
+        super(EnabledFilterTestCase, self).setUp()
+        self.service = mock.Mock()
+        self.enabled = True
+        self.enabled_filter = trivial_filters.EnabledFilter()
+
+    def test__repr__(self):
+        result = repr(self.enabled_filter)
+        self.assertEqual(result, "<EnabledFilter(enabled=%s)>" % self.enabled)
+
+    def test_rate_service(self):
+        self.service.enabled = True
+
+        result = self.enabled_filter.rate_service(self.service)
+        self.assertEqual(result, 100)
+
+    def test_rate_service_false(self):
+        self.service.enabled = False
+
+        result = self.enabled_filter.rate_service(self.service)
+        self.assertEqual(result, 0)
+
+
+@ddt.ddt
+class ProviderTypesFilterTestCase(test_base.CoriolisBaseTestCase):
+    """Test suite for the ProviderTypesFilter class."""
+
+    def setUp(self):
+        super(ProviderTypesFilterTestCase, self).setUp()
+        self.service = mock.Mock()
+        self.provider_types = [mock.sentinel.provider_type1,
+                               mock.sentinel.provider_type2]
+        self.provider_types_filter = trivial_filters.ProviderTypesFilter(
+            self.provider_types)
+
+    def test__repr__(self):
+        result = repr(self.provider_types_filter)
+        self.assertEqual(result,
+                         "<ProviderTypesFilter(provider_requirements=%s)>" %
+                         self.provider_types)
+
+    @ddt.data(
+        ({'platform1': ['type1']}, {'platform1': {'types': ['type1']}}, 100),
+        ({'platform1': ['type1']}, {'platform2': {'types': ['type1']}}, 0),
+        ({'platform1': ['type1', 'type2']},
+            {'platform1': {'types': ['type1']}}, 0)
+    )
+    @ddt.unpack
+    def test_rate_service(self, provider_requirements, providers, expected):
+        self.provider_types_filter.\
+            _provider_requirements = provider_requirements
+        self.service.providers = providers
+
+        result = self.provider_types_filter.rate_service(self.service)
+
+        self.assertEqual(result, expected)
--- a/coriolis/tests/scheduler/rpc/test_client.py
+++ b/coriolis/tests/scheduler/rpc/test_client.py
@@ -1,0 +1,242 @@
+# Copyright 2024 Cloudbase Solutions Srl
+# All Rights Reserved.
+
+import logging
+from unittest import mock
+
+import oslo_messaging
+
+from coriolis import constants
+from coriolis import exception
+from coriolis.scheduler.rpc import client
+from coriolis.tasks import factory as tasks_factory
+from coriolis.tests import test_base
+
+
+class SchedulerClientTestCase(test_base.CoriolisBaseTestCase):
+    """Test suite for the Coriolis Scheduler Worker RPC client."""
+
+    def setUp(self):
+        super(SchedulerClientTestCase, self).setUp()
+        self.client = client.SchedulerClient()
+        self.task = {'id': 'task_id', 'task_type': 'task_type'}
+        self.origin_endpoint = {
+            'id': 'origin_id',
+            'mapped_regions': [{'id': 'region1'}, {'id': 'region2'}],
+            'type': 'origin_type'
+        }
+        self.destination_endpoint = {
+            'id': 'destination_id',
+            'mapped_regions': [{'id': 'region3'}, {'id': 'region4'}],
+            'type': 'destination_type'
+        }
+
+    @mock.patch('coriolis.scheduler.rpc.client.CONF')
+    @mock.patch.object(oslo_messaging, 'Target')
+    def test__init__(self, mock_target, mock_conf):
+        expected_timeout = 120
+        mock_conf.scheduler.scheduler_rpc_timeout = expected_timeout
+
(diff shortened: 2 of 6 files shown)


coriolis/tests/scheduler/filters/test_trivial_filters.py
```diff
@@ -0,0 +1,134 @@
...
+    @ddt.unpack
+    def test_rate_service(self, regions, mapped_regions, expected):
+        self.regions_filter._regions = regions
+        mock_service = mock.Mock()
+        mock_service.mapped_regions = [mock.Mock(id=region) for region in
+                                       mapped_regions]
+
+        result = self.regions_filter.rate_service(mock_service)
+
+        self.assertEqual(result, expected)
```
Dany9966, line 47: There's a missed case here. Please account for the case where `any_regions` is set to `True`, and also at least one (not all) of the regions are found in `mapped_regions` (just like the last test case).

coriolis/tests/scheduler/rpc/test_client.py
```diff
@@ -0,0 +1,221 @@
...
+        result = client.SchedulerClient()
+        mock_target.assert_called_once_with(
+            topic='coriolis_scheduler', version=client.VERSION)
+
+        self.assertEqual(result._target, mock_target.return_value)
+        self.assertEqual(result._timeout, expected_timeout)
+
+    def test_init_without_timeout(self):
+        result = client.SchedulerClient()
+        self.assertEqual(result._timeout, 60)
```
Dany9966, line 49: What if `timeout` is explicitly set? :D

coriolis/tests/scheduler/rpc/test_client.py
```diff
@@ -0,0 +1,221 @@
...
+        self.assertEqual(result, mock_call.return_value)
+
+    @mock.patch.object(client.SchedulerClient, '_call')
+    def test_get_any_worker_service(self, mock_call):
+        ctxt = mock.sentinel.ctxt
+        raise_if_none = mock.sentinel.raise_if_none
+
+        self.client.get_any_worker_service(ctxt, raise_if_none=raise_if_none)
+
+        mock_call.assert_called_once_with(
```
Dany9966, line 82: Please explicitly mock and check `get_workers_for_spec` instead of checking the underlying functionality of `get_workers_for_spec`.

Dany9966: Also, please check what is returned. You can mock `random.choice` and set the value you want.

coriolis/tests/scheduler/rpc/test_client.py
```diff
@@ -0,0 +1,221 @@
...
+    def test_get_worker_service_for_specs_random_choice(
+            self, mock_get_workers):
+        service_mock1 = {'id': 'test_id1'}
+        service_mock2 = {'id': 'test_id2'}
+        mock_get_workers.return_value = [service_mock1, service_mock2]
+
+        result = self.client.get_worker_service_for_specs(
+            mock.sentinel.ctxt, random_choice=True)
+
+        self.assertIsInstance(result, dict)
```
Dany9966, line 154: Please also add a case where you set `random_choice=False`, check result.

coriolis/tests/scheduler/rpc/test_client.py
```diff
@@ -0,0 +1,221 @@
...
+
+    @mock.patch.object(client.SchedulerClient, 'get_workers_for_specs')
+    @mock.patch.object(tasks_factory, 'get_task_runner_class')
+    def test_get_worker_service_for_task_different_platforms(
```
Dany9966, line 158: This is basically an empty test, it just calls the method. Please at least check result

coriolis/tests/scheduler/rpc/test_client.py
```diff
@@ -0,0 +1,221 @@
...
+        service_mock1 = {'id': 'test_id1'}
+        service_mock2 = {'id': 'test_id2'}
+        mock_get_workers.return_value = [service_mock1, service_mock2]
+
+        result = self.client.get_worker_service_for_specs(
+            mock.sentinel.ctxt, random_choice=True)
+
+        self.assertIsInstance(result, dict)
+
+    @mock.patch.object(client.SchedulerClient, 'get_workers_for_specs')
```
Dany9966, line 156: Also why mocking this? You should be mocking `get_worker_service_for_specs` directly.

coriolis/tests/scheduler/rpc/test_client.py
```diff
@@ -0,0 +1,221 @@
...
+                    constants.PROVIDER_PLATFORM_SOURCE: 'provider_type'}
+            mock_get_workers.return_value = [{'id': 'test_id'}]
+
+            self.client.get_worker_service_for_task(
+                mock.sentinel.ctxt, self.task, self.origin_endpoint,
+                self.destination_endpoint)
+
+    @mock.patch.object(client.SchedulerClient, 'get_workers_for_specs')
+    @mock.patch.object(tasks_factory, 'get_task_runner_class')
+    def test_get_worker_service_for_task_different_provider_types(
```
Dany9966, line 176: This is bascally the same as above. You can remove this

coriolis/tests/scheduler/rpc/test_client.py
```diff
@@ -0,0 +1,221 @@
...
+        mock_get_task_runner_class.return_value.get_required_platform.\
+            return_value = constants.TASK_PLATFORM_SOURCE
+        mock_get_task_runner_class.return_value.get_required_provider_types.\
+            return_value = {constants.PROVIDER_PLATFORM_SOURCE: 'type'}
+        mock_get_workers.side_effect = Exception()
+
+        self.assertRaises(
+            exception.NoSuitableWorkerServiceError,
+            self.client.get_worker_service_for_task, mock.sentinel.ctxt,
+            self.task, self.origin_endpoint, self.destination_endpoint)
```
Dany9966, line 221: You could also set `retry_period=0` to make it run faster. You could apply this anywhere.

Dany9966 (changes requested):

coriolis/tests/scheduler/test_scheduler_utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def setUp(self):
+        super(SchedulerUtilsTestCase, self).setUp()
+        self.scheduler_client = mock.MagicMock()
+        self.rpc_client_class = mock.MagicMock()
+        self.service = mock.MagicMock()
+        self.ctxt = mock.MagicMock()
+
+    def test_get_rpc_client_for_service(self):
+        scheduler_utils.RPC_TOPIC_TO_CLIENT_CLASS_MAP[
```
Dany9966, line 28: I don't like directly changing the dict. For safety's sake, can you please use `mock.patch.dict` here? 
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch.dict

Dany9966: Also, what if the service topic is not `constants.WORKER_MAIN_MESSAGING_TOPIC`? Please add a case for that too.

coriolis/tests/scheduler/test_scheduler_utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+        self.scheduler_client.get_workers_for_specs.return_value = [
+            service_mock1, service_mock2]
+
+        get_service_mock.side_effect = [service_mock1, service_mock2]
+
+        result = scheduler_utils.get_any_worker_service(
+            self.scheduler_client, self.ctxt, random_choice=True)
+
+        self.assertIsInstance(result, dict)
```
Dany9966, line 69: For full control, please mock `random.choice` and set a value of your choice. Since you call `get_any_worker_service` with `raw_dict=False`, this statement kinda contradicts that. The final result should be the return value of `db_api.get_service`.

coriolis/tests/scheduler/test_scheduler_utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @mock.patch('coriolis.scheduler.scheduler_utils.db_api.get_service')
+    def test_get_any_worker_service_raw_dict(self, get_service_mock):
+        service_mock = {'id': 'test_id'}
+
+        self.scheduler_client.get_workers_for_specs.return_value = [
+            service_mock]
+
+        get_service_mock.return_value = service_mock
```
Dany9966, line 78: By the logic of the original method, it's the other way around. If `raw_dict=True`, `get_service_mock` should not have been called, so it's impossible to have its return value as result, but since you set the actual return value here, the test coincidentally passed.

coriolis/tests/scheduler/test_scheduler_utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+        result = scheduler_utils.get_any_worker_service(
+            self.scheduler_client, self.ctxt, raw_dict=True)
+
+        self.assertEqual(result, get_service_mock.return_value)
+
+    def test_get_worker_rpc_for_host(self):
+        scheduler_utils.RPC_TOPIC_TO_CLIENT_CLASS_MAP[
```
Dany9966, line 86: Please use `mock.patch.dict` here also.

Dany9966 (changes requested):
As a general note, can you please start using 2024 as your copyright year at the start of your new files?

mihaelabalutoiu:
Thanks a lot, I updated the code, please take a look now. :)

coriolis/tests/scheduler/rpc/test_client.py
```diff
@@ -0,0 +1,241 @@
...
+            enabled=True)
+        self.assertEqual(result, service_mock1)
+
+    @mock.patch('random.choice')
+    @mock.patch.object(client.SchedulerClient, 'get_workers_for_specs')
+    def test_get_worker_service_for_specs_no_random_choice(
+            self, get_workers_for_specs, mock_random_choice):
+        mock_service = mock.MagicMock()
+        get_workers_for_specs.return_value = [mock_service]
+        mock_random_choice.return_value = mock_service
```
Dany9966, line 179: We don't care about `random`'s return value here. Please add a `mock_random_choice.assert_not_called()` instead

coriolis/tests/scheduler/test_scheduler_utils.py
```diff
@@ -0,0 +1,125 @@
...
+from coriolis.scheduler import scheduler_utils
+from coriolis.tests import test_base
+
+
+class CoriolisTestException(Exception):
+    pass
+
+
+class SchedulerUtilsTestCase(test_base.CoriolisBaseTestCase):
+    """"Test suite for the Coriolis scheduler utils package."""
```
Dany9966, line 17: NIT: docstring has too many double-quotes (4 instead of 3)
```suggestion
    """Test suite for the Coriolis scheduler utils package."""
```

coriolis/tests/scheduler/rpc/test_server.py
```diff
@@ -52,7 +172,7 @@ def test_get_workers_for_specs(
         provider_requirements = config.get("provider_requirements", None)
 
         # Convert the config dict to an object, skipping the providers
-        # providers is the only field used as dict in the code
+        # is the only field used as dict in the code
```
Dany9966, line 175:
```suggestion
        # as it's the only field used as dict in the code
```

mihaelabalutoiu: Done, thank you!

coriolis/tests/scheduler/rpc/test_client.py
```diff
@@ -0,0 +1,241 @@
...
+            self.assertIsInstance(result, dict)
+
+    @mock.patch.object(client.SchedulerClient, 'get_workers_for_specs')
+    @mock.patch.object(tasks_factory, 'get_task_runner_class')
+    def test_get_worker_service_for_task_retry(
+            self, mock_get_task_runner_class, mock_get_workers):
+        mock_get_task_runner_class.return_value.get_required_platform.\
+            return_value = constants.TASK_PLATFORM_SOURCE
+        mock_get_task_runner_class.return_value.get_required_provider_types.\
+            return_value = {constants.PROVIDER_PLATFORM_SOURCE:
```
Dany9966, line 217: Please edit this to `constants.PROVIDER_PLATFORM_DESTINATION` for full coverage

coriolis/tests/scheduler/rpc/test_client.py
```diff
@@ -0,0 +1,241 @@
...
+            get_worker_service_for_specs.return_value = {'id': 'test_id'}
+
+            result = self.client.get_worker_service_for_task(
+                mock.sentinel.ctxt, self.task, self.origin_endpoint,
+                self.destination_endpoint, retry_period=0)
+
+            self.assertEqual(result, {'id': 'test_id'})
+            self.assertIsInstance(result, dict)
+
+    @mock.patch.object(client.SchedulerClient, 'get_workers_for_specs')
```
Dany9966, line 210: This should've mocked `get_worker_service_for_specs`

coriolis/tests/scheduler/rpc/test_client.py
```diff
@@ -0,0 +1,241 @@
...
+                            'provider_type'}
+        mock_get_workers.side_effect = [Exception(), [{'id': 'test_id'}]]
+
+        with self.assertLogs('coriolis.scheduler.rpc.client',
+                             level=logging.WARN):
+            self.client.get_worker_service_for_task(
+                mock.sentinel.ctxt, self.task, self.origin_endpoint,
+                self.destination_endpoint, retry_period=0)
+
+    @mock.patch.object(client.SchedulerClient, 'get_workers_for_specs')
```
Dany9966, line 227: Same: This should've mocked `get_worker_service_for_specs`

Dany9966: This will also help with coverage, as currently `get_workers_for_specs` raises `NoSuitableWorkerServiceError`, instead of the tested method

mihaelabalutoiu: Gotcha, thank you!

coriolis/tests/scheduler/rpc/test_client.py
```diff
@@ -0,0 +1,242 @@
...
+    @mock.patch.object(client.SchedulerClient, 'get_worker_service_for_specs')
+    @mock.patch.object(tasks_factory, 'get_task_runner_class')
+    def test_get_worker_service_for_task_no_suitable_worker(
+            self, mock_get_task_runner_class, mock_get_worker_service):
+        mock_get_task_runner_class.return_value.get_required_platform.\
+            return_value = constants.TASK_PLATFORM_SOURCE
+        mock_get_task_runner_class.return_value.get_required_provider_types.\
+            return_value = {constants.PROVIDER_PLATFORM_SOURCE: 'type'}
+        mock_get_worker_service.\
+            side_effect = [exception.NoSuitableWorkerServiceError()]
```
Dany9966, line 236: NIT: We usually prefer this code style instead:
```suggestion
        mock_get_worker_service.side_effect = [
            exception.NoSuitableWorkerServiceError()]
```

Please apply this style from now on

PR #299, opened 2024-03-15
Set default drive letter to all Windows partitions

Dany9966:
This patch makes sure all mounted partitions get a drive letter assigned, especially useful for VirtIO disks which on some Windows guests are implemented as removable drives, which get no boot drive letter by default.


gluka-CBS:
tested OK

PR #300, opened 2024-03-15
DB: warn and truncate on overly-long task event messages.

aznashwan:


aznashwan:
@Dany9966 pls consider this for release since some Plugin which shall not be named's error messages has like 5k chars from the server and this can lead to error events not getting registered and shown to users

PR #307, opened 2024-04-02
[OSMount] Add waits/retry during Ubuntu minion prep stage.

aznashwan:
Add wait/retry to the pre-setup stage of the Ubuntu OSMorphing minions for installing packages required by the OSMount process.


aznashwan:
Tested in semi-contrived scenario (cloud-init-metadata mandated package update on minion in first boot) and does the trick fine. Merging now.

PR #308, opened 2024-04-03
[OSMorphing] Add automatic `dpkg --configure` step before installing packages.

aznashwan:


aznashwan:
Tested and works. Merging now.

PR #317, opened 2024-05-22
Add unit tests for `migrations/api.py` module

mihaelabalutoiu:
This PR adds test cases for `coriolis/migrations/api.py` module

### DIFF
--- a/coriolis/tests/migrations/test_api.py
+++ b/coriolis/tests/migrations/test_api.py
@@ -1,0 +1,93 @@
+# Copyright 2024 Cloudbase Solutions Srl
+# All Rights Reserved.
+
+from unittest import mock
+
+from coriolis.migrations import api as migrations_module
+from coriolis.tests import test_base
+
+
+class APITestCase(test_base.CoriolisBaseTestCase):
+    """Test suite for the Coriolis Migrations API."""
+
+    def setUp(self):
+        super(APITestCase, self).setUp()
+        self.api = migrations_module.API()
+        self.rpc_client = mock.MagicMock()
+        self.api._rpc_client = self.rpc_client
+        self.ctxt = mock.sentinel.ctxt
+        self.migration_id = mock.sentinel.migration_id
+
+    def test_migrate_instances(self):
+        origin_endpoint_id = mock.sentinel.origin_endpoint_id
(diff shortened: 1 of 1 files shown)


coriolis/tests/migrations/test_api.py
```diff
@@ -0,0 +1,97 @@
...
+        storage_mappings = mock.sentinel.storage_mappings
+        replication_count = mock.sentinel.replication_count
+        shutdown_instances = mock.sentinel.shutdown_instances
+
+        result = self.api.migrate_instances(
+            self.ctxt, origin_endpoint_id, destination_endpoint_id,
+            origin_minion_pool_id, destination_minion_pool_id,
+            instance_osmorphing_minion_pool_mappings, source_environment,
+            destination_environment, instances, network_map, storage_mappings,
+            replication_count, shutdown_instances, notes=None,
```
ionutbalutoiu, line 41: Do not explicitly give the `kwargs` (`notes=None, skip_os_morphing=False, user_scripts=None`) when calling the tested function.

This way, we can cover with the unit test if those default values change, the assertion below will fail.

coriolis/tests/migrations/test_api.py
```diff
@@ -0,0 +1,97 @@
...
+            notes=None, skip_os_morphing=False, user_scripts=None)
+        self.assertEqual(result,
+                         self.rpc_client.migrate_instances.return_value)
+
+    def test_deploy_replica_instances(self):
+        replica_id = mock.sentinel.replica_id
+        instance_osmorphing_minion_pool_mappings = (
+            mock.sentinel.instance_osmorphing_minion_pool_mappings)
+
+        result = self.api.deploy_replica_instances(
```
ionutbalutoiu, line 59: Same as above, remove the explicit `kwargs`, and keep the same assertion.

ionutbalutoiu (changes requested):
Two small comments. The rest LGTM.

coriolis/tests/migrations/test_api.py
mihaelabalutoiu: Gotcha! You are right!

ionutbalutoiu (approved):

PR #321, opened 2024-06-10
Add `minion_manager.rpc.client.py/tasks.py` unit tests

mihaelabalutoiu:
This PR adds test cases for the following module:

- `minion_manager.rpc.tasks.py`
- `minion_manager.rpc.client.py`

### DIFF
--- a/coriolis/tests/minion_manager/rpc/test_tasks.py
+++ b/coriolis/tests/minion_manager/rpc/test_tasks.py
@@ -1,0 +1,2038 @@
+# Copyright 2024 Cloudbase Solutions Srl
+# All Rights Reserved.
+
+from unittest import mock
+
+from taskflow.types import failure
+
+from coriolis.conductor.rpc.client import ConductorClient
+from coriolis import exception
+from coriolis.minion_manager.rpc.client import MinionManagerClient
+from coriolis.minion_manager.rpc import tasks
+from coriolis.taskflow import base
+from coriolis.tests import test_base
+
+
+class CoriolisTestException(Exception):
+    pass
+
+
+class MinionManagerTaskEventMixinTestCase(test_base.CoriolisBaseTestCase):
+    """Test suite for the Coriolis MinionManagerTaskEventMixin class."""
+
+    def setUp(self):
+        super(MinionManagerTaskEventMixinTestCase, self).setUp()
+        self.task = tasks.MinionManagerTaskEventMixin()
+        self.task._minion_pool_id = mock.sentinel.minion_pool_id
+
+    def test__conductor_client(self):
+        with mock.patch(
+            'coriolis.conductor.rpc.client.ConductorClient',
+                return_value=mock.MagicMock(spec=ConductorClient)) as \
+                mock_Conductor_client:
+            result = self.task._conductor_client
+
+            self.assertEqual(
+                result, mock_Conductor_client.return_value
+            )
+            mock_Conductor_client.assert_called_once_with()
+
+    def test__conductor_client_already_set(self):
+        with mock.patch(
+            'coriolis.conductor.rpc.client.ConductorClient') as \
+                mock_Conductor_client:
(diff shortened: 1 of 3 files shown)


Dany9966 (changes requested):
First review pass done! Please look out for messages (log, event, progress updates, whatever) we do not want to include in tests, and remove all of them please.

mihaelabalutoiu (commented):
I updated the code, please take a look now. Thank you!

coriolis/tests/minion_manager/rpc/test_tasks.py
```diff
@@ -0,0 +1,2009 @@
...
+        test_base.CoriolisBaseTestCase):
+
+    def setUp(self):
+        super(ConfirmMinionAllocationForMigrationTaskTestCase, self).setUp()
+        self.action_id = mock.sentinel.action_id
+        self.allocated_machine_id_mappings = [
+            mock.sentinel.allocate_machine_id_mappings]
+
+        self.task = tasks.ConfirmMinionAllocationForMigrationTask(
+            mock.sentinel.action_id, self.allocated_machine_id_mappings)
```
Dany9966, line 487: You don't use `self.allocated_machine_id_mappings` anywhere. You can pass that sentinel directly
```suggestion
        self.task = tasks.ConfirmMinionAllocationForMigrationTask(
            mock.sentinel.action_id,
            mock.sentinel.allocated_machine_id_mappings)
```

coriolis/tests/minion_manager/rpc/test_tasks.py
```diff
@@ -0,0 +1,2009 @@
...
+class ConfirmMinionAllocationForReplicaTaskTestCase(
+        test_base.CoriolisBaseTestCase):
+
+    def setUp(self):
+        super(ConfirmMinionAllocationForReplicaTaskTestCase, self).setUp()
+        self.allocated_machine_id_mappings = [
+            mock.sentinel.allocate_machine_id_mappings]
+
+        self.task = tasks.ConfirmMinionAllocationForReplicaTask(
+            mock.sentinel.action_id, self.allocated_machine_id_mappings)
```
Dany9966, line 526: Same as above

Dany9966 (changes requested):

Dany9966:
Mostly lgtm, you only have a little missed case in `HealthcheckMinionMachineTask.execute`. Please handle that too.

coriolis/tests/minion_manager/rpc/test_tasks.py
```diff
@@ -0,0 +1,2038 @@
...
+    @mock.patch.object(tasks.BaseMinionManangerTask, 'execute')
+    @mock.patch.object(
+        tasks.MinionManagerTaskEventMixin,
+        '_set_minion_machine_allocation_status'
+    )
+    def test_execute_with_exception_and_fail_on_error(
+            self, mock_set_minion_machine_allocation, mock_execute,
+            mock_add_minion_pool_event, mock_get_minion_machine):
+        mock_get_minion_machine.return_value = self.minion_machine
+        mock_execute.side_effect = CoriolisTestException
```
Dany9966, line 1679: I don't get what you're testing here. You mock the tested method, make it raise and then check that the raise has occurred? 
For this case, you need to:
1. set `mock_get_minion_machine.return_value = None`;
2. remove the `mock_execute` entirely;
3. keep `_fail_on_error` on `True`;
4. Check that the call raises `InvalidMinionMachineState`.

Dany9966 (changes requested):

coriolis/tests/minion_manager/rpc/test_tasks.py
Dany9966: Nevermind this, this case was already covered. This case is also good. Thank you!

Dany9966 (approved):
LGTM. Thank you!

PR #322, opened 2024-06-19
OSMount: fix internal VG name referencing conflict.

aznashwan:
The Coriolis OSMount code would internally keep a mapping between names of VGs it should activate and their PVs, which could potentially lead to VGs with multiple PVs to be referneced both by their original name, as well as a new one Coriolis generates to avoid naming conflicts in /dev/$VG_NAME/$LVM_NAME.


Dany9966 (commented):
Code LGTM, but please fix the pep8 errors. You got a couple of "line too longs" :D

PR #324, opened 2024-07-01
Add unit tests for `coriolis.db.sqlalchemy.*` modules

Cristi1324:
This PR add unit tests for `coriolis.db.sqlalchemy.*` modules

### DIFF
--- a/coriolis/tests/db/sqlalchemy/test_migration.py
+++ b/coriolis/tests/db/sqlalchemy/test_migration.py
@@ -1,0 +1,61 @@
+# Copyright 2024 Cloudbase Solutions Srl
+# All Rights Reserved.
+
+import os
+from unittest import mock
+
+from oslo_db.sqlalchemy import migration as oslo_migration
+
+from coriolis.db.sqlalchemy import migration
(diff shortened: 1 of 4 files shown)


coriolis/tests/db/sqlalchemy/test_migration.py
```diff
@@ -0,0 +1,50 @@
...
+    def test_db_sync(self, mock_db_sync):
+        result = migration.db_sync(mock.sentinel.engine, mock.sentinel.version)
+
+        self.assertEqual(
+            mock_db_sync.return_value,
+            result
+        )
+        mock_db_sync.assert_called_once_with(
+            mock.sentinel.engine,
+            mock.ANY,
```
Dany9966, line 24: Instead of passing `mock.ANY`, let's please mock `os.path.abspath`, make it return something like `"/abspath"`. Then you can replace this with `"/abspath/migrate_repo"`. Please apply this to all of these test methods

Dany9966 (changes requested):
First review pass done, mostly LGTM

PR #337, opened 2024-08-29
[OSMorphing] Check installed packages before installing packages

Cristi1324:
This PR adds check_installed_package for Redhat, Debian and SUSE.

### DIFF
--- a/coriolis/osmorphing/debian.py
+++ b/coriolis/osmorphing/debian.py
@@ -122,6 +122,14 @@ class BaseDebianMorphingTools(base.BaseLinuxOSMorphingTools):
             cfg_name = "%s/coriolis_netplan.yaml" % netplan_base
             self._write_file_sudo(cfg_name, new_cfg)
 
+    def get_installed_packages(self):
+        cmd = "dpkg-query -f '${binary:Package}\\n' -W"
+        try:
+            self.installed_packages = self._exec_cmd_chroot(
+                cmd).decode('utf-8').splitlines()
+        except exception.CoriolisException:
+            pass
+
     def pre_packages_install(self, package_names):
         super(BaseDebianMorphingTools, self).pre_packages_install(
             package_names)
--- a/coriolis/osmorphing/manager.py
+++ b/coriolis/osmorphing/manager.py
@@ -239,6 +239,11 @@ def morph_image(origin_provider, destination_provider, connection_info,
         LOG.info("Post packages uninstall")
         export_os_morphing_tools.post_packages_uninstall(packages_remove)
 
+    LOG.info("Checking for packages already installed")
+    import_os_morphing_tools.get_installed_packages()
+    packages_add = list(
+        set(packages_add) - set(import_os_morphing_tools.installed_packages))
+
     LOG.info("Pre packages install")
     import_os_morphing_tools.pre_packages_install(packages_add)
 
--- a/coriolis/osmorphing/redhat.py
+++ b/coriolis/osmorphing/redhat.py
@@ -209,15 +209,14 @@ class BaseRedHatMorphingTools(base.BaseLinuxOSMorphingTools):
                                                     mac_addresses)
         self._add_net_udev_rules(net_ifaces_info)
 
-    def _has_package_installed(self, package_name):
-        cmd = 'rpm -q %s' % ("".join(package_name))
+    def get_installed_packages(self):
+        cmd = 'rpm -qa --qf "%{NAME}\\n"'
         try:
-            self._exec_cmd_chroot(cmd)
-            return True
+            self.installed_packages = self._exec_cmd_chroot(
+                cmd).decode('utf-8').splitlines()
         except exception.CoriolisException:
-            LOG.warning(f"Package ${package_name} is not installed")
+            LOG.warning("Failed to get installed packages")
             LOG.trace(utils.get_exception_details())
-            return False
 
     def _yum_install(self, package_names, enable_repos=[]):
         try:
@@ -295,7 +294,7 @@ class BaseRedHatMorphingTools(base.BaseLinuxOSMorphingTools):
         super(BaseRedHatMorphingTools, self).pre_packages_install(
             package_names)
         self._yum_clean_all()
-        if not self._has_package_installed('grubby'):
+        if 'grubby' not in self.installed_packages:
             self._yum_install(['grubby'])
         else:
             LOG.debug("Skipping package 'grubby' as it's already installed")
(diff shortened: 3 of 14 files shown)


coriolis/osmorphing/debian.py
```diff
@@ -122,6 +122,14 @@ def set_net_config(self, nics_info, dhcp):
             cfg_name = "%s/coriolis_netplan.yaml" % netplan_base
             self._write_file_sudo(cfg_name, new_cfg)
 
+    def check_installed_package(self, package_name):
+        cmd = 'dpkg-query -W %s' % ("".join(package_name))
```
Dany9966, line 126: I think it would be better to just send a single command that lists all the packages installed on the migrated system. Parse that list and save it on the tools object. Then when you need to check a package, you just do a `if package in self.installed_packages: return` or do a set difference like below to get the list of packages that need to be installed. 

The point of this is to send a minimum amount of SSH commands, instead of sending N+1commands for each package we want to install. For Debian-based, you can use `apt list --installed`, please check a command for the others too.

Dany9966: You also don't need to `"".join(package_name)` here, you can simply pass the string value to the string format.

Dany9966 (changes requested):

coriolis/osmorphing/manager.py
```diff
@@ -239,6 +239,10 @@ def morph_image(origin_provider, destination_provider, connection_info,
         LOG.info("Post packages uninstall")
         export_os_morphing_tools.post_packages_uninstall(packages_remove)
 
+    LOG.info("Checking for packages already installed")
+    packages_add = import_os_morphing_tools.check_installed_packages(
```
Dany9966, line 243: You can replace this with calling `get_installed_packages`, and then only install the set difference

coriolis/osmorphing/redhat.py
```diff
@@ -209,15 +209,14 @@ def set_net_config(self, nics_info, dhcp):
                                                     mac_addresses)
         self._add_net_udev_rules(net_ifaces_info)
 
-    def _has_package_installed(self, package_name):
-        cmd = 'rpm -q %s' % ("".join(package_name))
+    def get_installed_packages(self):
+        cmd = 'rpm -qa'
```
Dany9966, line 213: You can finetune this command also to only get the list of package names:
```
rpm -qa --qf '%{NAME}\n'
```

coriolis/osmorphing/redhat.py
```diff
@@ -295,7 +294,7 @@ def pre_packages_install(self, package_names):
         super(BaseRedHatMorphingTools, self).pre_packages_install(
             package_names)
         self._yum_clean_all()
-        if not self._has_package_installed('grubby'):
+        if not any('grubby' in package for package in self.installed_packages):
```
Dany9966, line 297: As mentioned above, you can just check if 'grubby` is in `self.installed_packages`. Do please check the package name string as a whole, as this could have buggy results. Let's say your list has a package called 'grubby-somethhing', but you need 'grubby' package instead, this code would skip that install.

coriolis/osmorphing/suse.py
```diff
@@ -65,6 +65,15 @@ def set_net_config(self, nics_info, dhcp):
         # TODO(alexpilotti): add networking support
         pass
 
+    def get_installed_packages(self):
+        cmd = 'rpm -qa'
```
Dany9966, line 69: Make this the same as redhat please.

Dany9966 (changes requested):

coriolis/osmorphing/redhat.py
```diff
@@ -209,15 +209,17 @@ def set_net_config(self, nics_info, dhcp):
                                                     mac_addresses)
         self._add_net_udev_rules(net_ifaces_info)
 
-    def _has_package_installed(self, package_name):
-        cmd = 'rpm -q %s' % ("".join(package_name))
+    def get_installed_packages(self):
+        cmd = 'rpm -qa --qf \'%{NAME}\n\''
```
Dany9966, line 213: This is most probably going to fail. If it does, please try:
```python
cmd = 'rpm -qa --qf "%{NAME}\\n"'
```

Dany9966 (changes requested):

coriolis/osmorphing/debian.py
```diff
@@ -122,6 +122,17 @@ def set_net_config(self, nics_info, dhcp):
             cfg_name = "%s/coriolis_netplan.yaml" % netplan_base
             self._write_file_sudo(cfg_name, new_cfg)
 
+    def get_installed_packages(self):
+        cmd = "dpkg-query -f '${binary:Package}\n' -W"
+        try:
+            apt_list_output = self._exec_cmd_chroot(
```
Dany9966, line 128: Also, could you please check whether you could do the following:
```python
apt_list_output = self._exec_cmd_chroot(cmd).decode().splitlines()
```
This way, you only call for `decode()` once.

coriolis/osmorphing/manager.py
```diff
@@ -239,6 +239,13 @@ def morph_image(origin_provider, destination_provider, connection_info,
         LOG.info("Post packages uninstall")
         export_os_morphing_tools.post_packages_uninstall(packages_remove)
 
+    LOG.info("Checking for packages already installed")
+    import_os_morphing_tools.get_installed_packages()
+    LOG.info("Packages already installed: %s" %
```
Dany9966, line 244: This could be a big list, please remove it.

coriolis/osmorphing/redhat.py
```diff
@@ -209,15 +209,17 @@ def set_net_config(self, nics_info, dhcp):
...
 
-    def _has_package_installed(self, package_name):
-        cmd = 'rpm -q %s' % ("".join(package_name))
+    def get_installed_packages(self):
+        cmd = 'rpm -qa --qf "%{NAME}\n"'
         try:
-            self._exec_cmd_chroot(cmd)
-            return True
+            apt_list_output = self._exec_cmd_chroot(
+                cmd).splitlines()
```
Dany9966, line 216: Same as above about the `decode()`

coriolis/osmorphing/suse.py
```diff
@@ -65,6 +65,18 @@ def set_net_config(self, nics_info, dhcp):
         # TODO(alexpilotti): add networking support
         pass
 
+    def get_installed_packages(self):
+        cmd = 'rpm -qa --qf \'%{NAME}\n\''
+        try:
+            apt_list_output = self._exec_cmd_chroot(
+                cmd).splitlines()
```
Dany9966, line 72: Same as above about the `decode()`

Dany9966 (changes requested):

Dany9966 (commented):
LGTM. If your testing also looks good, please go ahead and update the unit tests also.

Dany9966 (approved):
LGTM! Thank you!

PR #340, opened 2024-09-16
`OSMount` - Fix handling of non-existent devices

mihaelabalutoiu:
This PR implements the following updates:
* Added a check to verify the existence of the `device` before attempting to `mount` it;
* Included a `unit test` for this scenario;


Dany9966 (approved):
LGTM!

PR #341, opened 2024-09-24
Fix Windows static IP configuration

Dany9966:
Improves the Windows static IP migration configuration by skipping detected dynamic IP addresses instead of directly raising an error.

### DIFF
--- a/coriolis/osmorphing/windows.py
+++ b/coriolis/osmorphing/windows.py
@@ -17,7 +17,6 @@ from coriolis.osmorphing import base
 from coriolis.osmorphing.osdetect import windows as windows_osdetect
 from coriolis import utils
 
-
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


coriolis/tests/osmorphing/test_windows.py
```diff
@@ -735,27 +735,32 @@ def test_compile_static_ip_conf_from_registry_no_static_ip(self):
...
-    def test_check_ips_info_different_ips(self):
-        nics_info = [
-            {'ip_addresses': [
-                mock.sentinel.ip_address, mock.sentinel.ip_address2]}
-        ]
-        ips_info = [
-            {'ip_address': mock.sentinel.ip_address}
-        ]
+            # no IP addresses on NIC
+            {"ip_addresses": [], "mac_address": "00:50:56:92:91:42"},
```
panigrap, line 744: The values, "00:50:56:92:91:" , ""fe81::728a:688:1a92:baec" .. etc .. can be defined as constants and can be referenced here. As we are using them multiple times.

gluka-CBS:
tested successfully

PR #352, opened 2025-03-24
Fix issue 'Permission denied' error due to repo permissions on redhat linux

fabi200123:
- Fixes the 'Permission denied' error based on the read permission of the `/etc/yum.repos.d/`
- Sends a warning in case of repository not being parsed due to an error
- Updates the redhat tests to match the call of the new method

### DIFF
--- a/coriolis/osmorphing/redhat.py
+++ b/coriolis/osmorphing/redhat.py
@@ -270,7 +270,12 @@ class BaseRedHatMorphingTools(base.BaseLinuxOSMorphingTools):
         installed_repos = []
         for file in repofiles:
             path = os.path.join(reposdir_path, file)
-            content = self._read_file(path).decode()
+            try:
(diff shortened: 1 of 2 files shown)


coriolis/osmorphing/redhat.py
```diff
@@ -270,7 +270,7 @@ def _find_yum_repos(self, repos_to_enable=[]):
         installed_repos = []
         for file in repofiles:
             path = os.path.join(reposdir_path, file)
-            content = self._read_file(path).decode()
+            content = self._read_file_sudo(path).decode()
```
Dany9966, line 273: Can you also please skip the file in case an error occurs?

fabi200123: Updated the code now to send a `LOG.warn()` and to continue in case an error occurs while trying to read the repo file

PR #359, opened 2025-07-07
instance user_scripts fix on deployments

BogdanEmilian:
This PR addresses an issue where instance user scripts added during deployment were not actually processed, which caused them to be unavailable during OS morphing.

To fix this, the _normalize_user_scripts logic was refactored and moved from coriolis.api.v1.utils to the ConductorServerEndpoint class, where it is now used only during the deployment stage of the transfer.

Related unit tests have also been refactored and moved into test_server.py.

### DIFF
--- a/coriolis/api/v1/deployments.py
+++ b/coriolis/api/v1/deployments.py
@@ -80,9 +80,6 @@ class DeploymentsController(api_wsgi.Controller):
             'instance_osmorphing_minion_pool_mappings', {})
         user_scripts = deployment.get('user_scripts', {})
         api_utils.validate_user_scripts(user_scripts)
-        user_scripts = api_utils.normalize_user_scripts(
-            user_scripts, deployment.get("instances", []))
-
         return (
             transfer_id, force, clone_disks, skip_os_morphing,
             instance_osmorphing_minion_pool_mappings,
--- a/coriolis/conductor/rpc/server.py
+++ b/coriolis/conductor/rpc/server.py
@@ -1664,6 +1664,21 @@ class ConductorServerEndpoint(object):
             db_api.set_action_last_execution_status(
                 ctxt, deployment_id, error_status)
 
+    def _normalize_user_scripts(self, user_scripts, instances):
+        """ Removes instance user_scripts if said instance is not one of the
+            selected instances for the replica/migration """
+        if user_scripts is None:
+            user_scripts = {}
+        for instance in list(user_scripts.get('instances', {}).keys()):
+            if instance not in instances:
+                LOG.warn("Removing provided instance '%s' from user_scripts "
+                         "body because it's not included in one of the "
+                         "selected instances for this replica/migration: %s",
+                         instance, instances)
+                user_scripts['instances'].pop(instance, None)
+
+        return user_scripts
+
     @transfer_synchronized
     def deploy_transfer_instances(
             self, ctxt, transfer_id, force=False, wait_for_execution=None,
@@ -1683,6 +1698,7 @@ class ConductorServerEndpoint(object):
             init_status = constants.EXECUTION_STATUS_PENDING
 
         instances = transfer.instances
+        user_scripts = self._normalize_user_scripts(user_scripts, instances)
 
         deployment = models.Deployment()
         deployment.id = str(uuid.uuid4())
--- a/coriolis/tests/conductor/rpc/test_server.py
+++ b/coriolis/tests/conductor/rpc/test_server.py
@@ -2023,6 +2023,31 @@ class ConductorServerEndpointTestCase(test_base.CoriolisBaseTestCase):
             to_dict=False
         )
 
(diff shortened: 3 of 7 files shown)


coriolis/api/v1/deployments.py
```diff
@@ -80,9 +81,9 @@ def _validate_deployment_input(self, context, body):
             'instance_osmorphing_minion_pool_mappings', {})
         user_scripts = deployment.get('user_scripts', {})
         api_utils.validate_user_scripts(user_scripts)
-        user_scripts = api_utils.normalize_user_scripts(
-            user_scripts, deployment.get("instances", []))
-
+        import remote_pdb; remote_pdb.RemotePdb('10.8.254.30', 4444).set_trace()
+        user_scripts = ConductorServerEndpoint._normalize_user_scripts(
```
Dany9966, line 85: The reason for this fix was to only run this normalization in the conductor itself, not make the API layer run it from the Conductor layer. (that's why I asked for it to be moved into the conductor)
Please remove any kind of user script normalization from the API layer.

Dany9966 (changes requested):

coriolis/conductor/rpc/server.py
```diff
@@ -1664,6 +1664,21 @@ def report_deployer_failure(
...
+        """ Removes instance user_scripts if said instance is not one of the
+            selected instances for the replica/migration """
+        if user_scripts is None:
+            user_scripts = {}
+        for instance in list(user_scripts.get('instances', {}).keys()):
+            if instance not in instances:
+                LOG.warn("Removing provided instance '%s' from user_scripts body "
+                         "because it's not included in one of the selected "
+                         "instances for this replica/migration: %s",
+                        instance, instances)
```
Dany9966, line 1677: This will most probably come up if the action runs, but please fix the indentation here:
```suggestion
                LOG.warn("Removing provided instance '%s' from user_scripts body "
                         "because it's not included in one of the selected "
                         "instances for this replica/migration: %s",
                         instance, instances)
```

Dany9966 (commented):
LGTM, aside from the one NIT comment which will either way come up in when the action runs the linter.

coriolis/tests/conductor/rpc/test_server.py
```diff
@@ -22,6 +22,7 @@
 from coriolis.tests import testutils
 from coriolis import utils
 from coriolis.worker.rpc import client as rpc_worker_client
+from coriolis.conductor.rpc.server import ConductorServerEndpoint
```
Dany9966, line 25: This is already imported above, as `server`. Please refer to the class as `server.ConductorServerEndpoint`. 
We never directly import classes, only whole modules. We generally follow openstack code style guidelines, so please also follow them: https://docs.openstack.org/hacking/latest/user/hacking.html

coriolis/tests/conductor/rpc/test_server.py
```diff
@@ -2023,6 +2024,32 @@ def test__get_transfer_not_found(self, mock_get_transfer):
...
+        instances = ["mock_instance_2", "mock_instance_3"]
+
+        expected_result = {
+            'instances': {
+                "mock_instance_2": "mock_value_2"
+            }
+        }
+
+        with self.assertLogs('coriolis.conductor.rpc.server', level='WARNING'):
+            endpoint = ConductorServerEndpoint()
```
Dany9966, line 2043: You can also use the test class' `self.server`.

coriolis/tests/conductor/rpc/test_server.py
```diff
@@ -2253,7 +2285,7 @@ def test_deploy_transfer_instances(
                 trust_id=None)
 
         self.assertEqual(mock_get_deployment.return_value, result)
-        self.assertEqual(mock.sentinel.user_scripts, deployment.user_scripts)
+        self.assertEqual({'instances': {}}, deployment.user_scripts)
```
Dany9966, line 2288: Below, please also add the following:
`mock_normalize_user_scripts.assert_called_once_with(mock.sentinel.user_scripts, transfer_mock.instances)`

Dany9966 (changes requested):

coriolis/tests/conductor/rpc/test_server.py
```diff
@@ -2023,6 +2023,32 @@ def test__get_transfer_not_found(self, mock_get_transfer):
...
+
+        expected_result = {
+            'instances': {
+                "mock_instance_2": "mock_value_2"
+            }
+        }
+
+        with self.assertLogs('coriolis.conductor.rpc.server', level='WARNING'):
+            endpoint = self.server
+            result = endpoint._normalize_user_scripts(user_scripts, instances)
```
Dany9966, line 2043: NIT: You could've called for `self.server._normalize_user_scripts` directly.

Dany9966 (commented):
LGTM, apart from the nits. Please fix those and also squash all the commits into a single one.

PR #364, opened 2025-08-13
[WIP] Refactor cloud init config

Cristi1324:


coriolis/osmorphing/base.py
```diff
@@ -427,6 +428,49 @@ def _configure_cloud_init_user_retention(self):
...
+        # self._exec_cmd_chroot("systemctl enable cloud-init")
+
+    def _reset_cloud_init_run(self):
+        self._exec_cmd_chroot("cloud-init clean --logs")
+
+    def _get_default_cloud_user(self):
+        cloud_cfg_path = 'etc/cloud/cloud.cfg'
+        if not self._test_path(cloud_cfg_path):
+            return DEFAULT_CLOUD_USER
+        cloud_cfg_content = self._read_file(cloud_cfg_path)
```
Dany9966, line 452: Please use `_read_file_sudo` here.

coriolis/osmorphing/base.py
```diff
@@ -427,6 +428,49 @@ def _configure_cloud_init_user_retention(self):
...
+        if self._test_path(installer_config_path):
+            self._event_manager.progress_update(
+                "Disabling Ubuntu installer-generated cloud-config")
+            self._exec_cmd_chroot(
+                f"mv {installer_config_path} {installer_config_path}.bak")
+
+    def _ensure_cloud_init_enabled(self):
+        disabler_file = "/etc/cloud/cloud-init.disabled"
+        if self._test_path(disabler_file):
+            self._exec_cmd_chroot(f"rm {disabler_file}")
```
Dany9966, line 442: This looks incomplete. There's a couple of more methods of disabling cloud-init, we should cover for those too.

coriolis/osmorphing/base.py
```diff
@@ -427,6 +428,49 @@ def _configure_cloud_init_user_retention(self):
...
+
+    def _configure_cloud_init(self):
+        if "cloud-init" not in self.get_packages()[0]:
+            return
+
+        self._disable_installer_cloud_config()
+        self._ensure_cloud_init_enabled()
+        self._reset_cloud_init_run()
+        if self._osmorphing_parameters.get('retain_user_credentials', False):
+            self._configure_cloud_init_user_retention()
```
Dany9966, line 470: This method overwrites cloud.cfg. I'd like to add those mods into a separate file called `99_coriolis.cfg` inside `cloud.cfg.d`. Save them inside a `cloud_cfg_mods` and only at the end of this `_configure_cloud_init` method you'd write that file. 
Also, here is what the `set_dhcp` mods should also be included.

Dany9966 (commented):
Mostly LGTM, just a couple of nits

Dany9966:
LGTM. Please update tests as well.

coriolis/osmorphing/base.py
```diff
@@ -393,39 +394,78 @@ def _set_selinux_autorelabel(self):
...
+            exception.CoriolisException("Could not find %s" % cloud_cfgs_dir)
         self._event_manager.progress_update(
-            "Reconfiguring cloud-init to retain original user credentials")
-        for cloud_cfg_path in cloud_cfg_paths:
-            if not self._test_path(cloud_cfg_path):
-                LOG.warn("Could not find %s. Skipping reconfiguration." % (
-                    cloud_cfg_path))
-                continue
+            "Configuring cloud-init additional options")
+        new_cloud_cfg = yaml.dump(cloud_cfg)
```
Dany9966, line 404: Please always use safe loaders and dumpers when using `yaml` module.
I.E.: `yaml.load(content, Loader=yaml.SafeLoader)`

```suggestion
        new_cloud_cfg = yaml.dump(cloud_cfg. Dumper=yaml.SafeDumper)
```

coriolis/osmorphing/base.py
```diff
@@ -393,39 +394,78 @@ def _set_selinux_autorelabel(self):
...
+        if not self._test_path(cloud_cfgs_dir):
+            exception.CoriolisException("Could not find %s" % cloud_cfgs_dir)
         self._event_manager.progress_update(
-            "Reconfiguring cloud-init to retain original user credentials")
-        for cloud_cfg_path in cloud_cfg_paths:
-            if not self._test_path(cloud_cfg_path):
-                LOG.warn("Could not find %s. Skipping reconfiguration." % (
-                    cloud_cfg_path))
-                continue
+            "Configuring cloud-init additional options")
```
Dany9966, line 403: NIT: I think this sounds better
```suggestion
            "Customizing cloud-init configuration")
```

coriolis/osmorphing/base.py
```diff
@@ -393,39 +394,78 @@ def _set_selinux_autorelabel(self):
...
+        cloud_config_path = "/etc/cloud/cloud.cfg.d/99_coriolis.cfg"
         cloud_cfgs_dir = "/etc/cloud/cloud.cfg.d"
-        if self._test_path(cloud_cfgs_dir):
-            for path in self._exec_cmd_chroot(
-                    'ls -1 %s' % cloud_cfgs_dir).decode().splitlines():
-                if path.endswith('.cfg'):
-                    cloud_cfg_paths.append("%s/%s" % (cloud_cfgs_dir, path))
-
+        if not self._test_path(cloud_cfgs_dir):
+            exception.CoriolisException("Could not find %s" % cloud_cfgs_dir)
```
Dany9966, line 401: I think it'd be better to just create this directory instead of straight out erroring out. Cloud-init will look for this location regardless, it's not customizable afaik. (if it is, find a way to locate it)
At least execute: `mkdir -p /etc/cloud/cloud.cfg.d`

coriolis/osmorphing/base.py
```diff
@@ -393,39 +394,78 @@ def _set_selinux_autorelabel(self):
...
-                continue
+            "Configuring cloud-init additional options")
+        new_cloud_cfg = yaml.dump(cloud_cfg)
+        self._write_file_sudo(cloud_config_path, new_cloud_cfg)
+
+    def _disable_installer_cloud_config(self):
+        installer_config_path = "/etc/cloud/cloud.cfg.d/99-installer.cfg"
+        if self._test_path(installer_config_path):
+            self._event_manager.progress_update(
+                "Disabling Ubuntu installer-generated cloud-config")
```
Dany9966, line 411: Since this is in base now, I think you should drop the `Ubuntu`

coriolis/osmorphing/base.py
```diff
@@ -393,39 +394,78 @@ def _set_selinux_autorelabel(self):
...
+
+    def _ensure_cloud_init_not_disabled(self):
+        disabler_file = "/etc/cloud/cloud-init.disabled"
+        system_conf_disabler = "/etc/systemd/system.conf"
+        grub_conf_disabler = "/etc/default/grub"
+        if self._test_path(disabler_file):
+            self._exec_cmd_chroot(f"rm {disabler_file}")
+        if self._test_path(system_conf_disabler):
+            self._exec_cmd_chroot(
+                "sed -i '/cloud-init=disabled/d' %s" % system_conf_disabler)
```
Dany9966, line 423: I'm not sure this will work as you expect it, it could just leave an empty variable and maybe even bork the system.conf file. If you haven't already, please do a test and see if it gets enabled and systemd doesn't just fall on its head. 
Let me know if you already tested this and it works.

coriolis/osmorphing/base.py
```diff
@@ -393,39 +394,78 @@ def _set_selinux_autorelabel(self):
...
+        if self._test_path(grub_conf_disabler):
+            self._exec_cmd_chroot(
+                "sed -i '/cloud-init=disabled/d' %s" % grub_conf_disabler)
```
Dany9966, line 426: This I know for sure will not work, because any grub update needs an `update-grub` or equivalent (redhat-derivates have a different command). 

In some cases, this grub update may not happen, so we need to also run the grub update command here.

I think for both cases (grub and system.conf), you'll need to first `grep` and then do sed and/or update commands.

coriolis/osmorphing/base.py
```diff
@@ -393,39 +394,78 @@ def _set_selinux_autorelabel(self):
...
+            cloud_cfg_user_retention = {
+                'disable_root': False,
+                'ssh_pwauth': True,
+                'users': None
+            }
+            cloud_cfg_mods.update(cloud_cfg_user_retention)
+        else:
+            self._create_cloudinit_user()
+
+        if not self._osmorphing_parameters.get('set_dhcp', False):
```
Dany9966, line 464: Please set `set_dhcp`'s default to True. Otherwise, in providers where we do not support this option, it will always disable cloud-init networking. 
```suggestion
        if not self._osmorphing_parameters.get('set_dhcp', True):
```

coriolis/osmorphing/redhat.py
```diff
@@ -265,6 +265,13 @@ def pre_packages_install(self, package_names):
         else:
             LOG.debug("Skipping package 'grubby' as it's already installed")
 
+    def post_packages_install(self, package_names):
+        self._configure_cloud_init()
+        self._run_dracut()
+        super(BaseRedHatMorphingTools, self).post_packages_install(
+            package_names
+        )
```
Dany9966, line 273: NIT:
```suggestion
            package_names)
```

coriolis/osmorphing/suse.py
```diff
@@ -127,6 +127,16 @@ def _has_systemd(self):
...
             return False
 
+    def _configure_cloud_init(self):
+        super(BaseSUSEMorphingTools, self)._configure_cloud_init()
+        if self._has_systemd():
+            self._enable_systemd_service("cloud-init")
+
+    def post_packages_install(self, package_names):
+        self._run_dracut()
+        self._configure_cloud_init()
```
Dany9966, line 137: NIT for consistency's sake:
```suggestion
        self._configure_cloud_init()
        self._run_dracut()
```

coriolis/osmorphing/base.py
```diff
@@ -393,39 +394,79 @@ def _set_selinux_autorelabel(self):
         except Exception as err:
             LOG.warning("Failed to set autorelabel: %r" % err)
 
-    def _configure_cloud_init_user_retention(self):
-        cloud_cfg_paths = ["/etc/cloud/cloud.cfg"]
+    def _write_cloud_init_mods_config(self, cloud_cfg):
+        cloud_config_path = "/etc/cloud/cloud.cfg.d/99_coriolis.cfg"
         cloud_cfgs_dir = "/etc/cloud/cloud.cfg.d"
```
Dany9966: Please use this everywhere instead of hardcoding it. 

```python
cloud_cfgs_dir = "/etc/cloud/cloud.cfg.d"
cloud_config_path = f"{cloud_cfgs_dir}/99_coriolis.cfg"
# ...
self._exec_cmd_chroot(f"mkdir -p {cloud_cfgs_dir}")
```

Dany9966 (approved):

PR #371, opened 2025-09-30
Uninstall vmware drivers

Cristi1324:


coriolis/osmorphing/windows.py
```diff
@@ -26,7 +26,9 @@
 SERVICE_START_MANUAL = 3
 SERVICE_START_DISABLED = 4
 
+SERVICES_PATH_FORMAT = "HKLM:\\%s\\ControlSet001\\Services"
```
Dany9966: Is this constant used anywhere? If not, please remove it.

PR #372, opened 2025-10-03
Enhance `OSMorphing` tools with `latest` flag support and improved version detection

mihaelabalutoiu:
This PR includes the following changes:

- Fix `CentOS` version detection for multi-digit releases;
- Add `NetworkManager` support for `CentOS` Stream 9+ `network configuration`;
- Implement `latest` flag system for OSMorphing tools;

### DIFF
--- a/coriolis/osmorphing/base.py
+++ b/coriolis/osmorphing/base.py
@@ -127,6 +127,8 @@ class BaseLinuxOSMorphingTools(BaseOSMorphingTools):
 
     _packages = {}
 
+    latest = False
+
     def __init__(self, conn, os_root_dir, os_root_dev, hypervisor,
                  event_manager, detected_os_info, osmorphing_parameters,
                  operation_timeout=None):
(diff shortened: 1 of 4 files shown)


coriolis/osmorphing/redhat.py
```diff
@@ -160,10 +202,25 @@ def _comment_keys_from_ifcfg_files(
...
+    def _get_os_version(self):
+        try:
+            version_str = self._detected_os_info.get('release_version', '0')
+            major_version = int(version_str.split('.')[0])
+            return major_version
+        except (ValueError, AttributeError):
+            LOG.warning(
+                "Could not parse OS version from detected_os_info: %s",
+                self._detected_os_info)
+            return
```
Dany9966: Please return 0 here, as it's expecting to be an integer value when using this method.

coriolis/osmorphing/redhat.py
```diff
@@ -160,10 +202,25 @@ def _comment_keys_from_ifcfg_files(
...
+
     def set_net_config(self, nics_info, dhcp):
         if dhcp:
             self.disable_predictable_nic_names()
-            self._write_nic_configs(nics_info)
+            os_version = self._get_os_version()
```
Dany9966: If you return None here, the following `if` will raise an exception.

Dany9966 (changes requested):

coriolis/osmorphing/base.py
```diff
@@ -156,6 +158,12 @@ def _version_supported_util(cls, version, minimum, maximum=None):
                 "Non-string version provided: %s (type %s)" % (
                     version, type(version)))
 
+        if cls.latest:
```
Dany9966: I think we should move this block before `if maximum:`, otherwise any old version will pickup the latest one.

Dany9966:
This resulted in a way too complicated design for what it tries to do. We'll make proper refactorings instead. Closing this now.

PR #383, opened 2026-01-20
Update enable_repo from osmorphing for all redhat based OSes

fabi200123:
This PR updates the logic so that during osmorphing, we'll no longer accept repository name substrings. The list of repository names defined by the providers must match exactly, otherwise a warning will be logged and the osmorphing workflow will continue without enabling the given repository.
 
If the instance uses custom repository names, the users may provide osmorphing scripts to enable necessary repositories.

### DIFF
--- a/coriolis/osmorphing/oracle.py
+++ b/coriolis/osmorphing/oracle.py
@@ -1,13 +1,17 @@
 # Copyright 2016 Cloudbase Solutions Srl
 # All Rights Reserved.
 
-import uuid
+from oslo_log import log as logging
 
+from coriolis import exception
 from coriolis.osmorphing.osdetect import oracle as oracle_detect
 from coriolis.osmorphing import redhat
+from coriolis import utils
 
 
 ORACLE_DISTRO_IDENTIFIER = oracle_detect.ORACLE_DISTRO_IDENTIFIER
+
+LOG = logging.getLogger(__name__)
 
 
 class BaseOracleMorphingTools(redhat.BaseRedHatMorphingTools):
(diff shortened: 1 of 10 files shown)


coriolis/osmorphing/redhat.py
```diff
@@ -179,11 +179,9 @@ def get_installed_packages(self):
             LOG.trace(utils.get_exception_details())
             self.installed_packages = []
 
-    def _yum_install(self, package_names, enable_repos=[]):
+    def _yum_install(self, package_names):
         try:
-            yum_cmd = 'yum install %s -y%s' % (
-                " ".join(package_names),
-                "".join([" --enablerepo=%s" % r for r in enable_repos]))
```
Dany9966: Please don't remove this flag from the command, as we may need it in some cases. We want to only remove the `_find_yum_repos` logic, which always runs. But the flag itself may prove itself useful on special cases (we'll have to look for what those special cases are in the providers).

coriolis/osmorphing/oracle.py
```diff
@@ -20,34 +23,18 @@ def check_os_supported(cls, detected_os_info):
...
+        for repo in repo_names:
+            cmd = 'yum-config-manager --enable %s' % repo
+            try:
+                self._exec_cmd_chroot(cmd)
+                LOG.info("Enabled repository '%s' using yum-config-manager",
+                         repo)
+            except exception.CoriolisException as err:
+                LOG.warning(
+                    "Failed to enable repository '%s' with: %s",
+                    repo, err)
```
Dany9966: Same as above

Dany9966 (approved):

Dany9966 (approved):

Dany9966 (approved):

Dany9966:
Can you please rebase this to latest code?

petrutlucian94 (approved):

PR #397, opened 2026-03-10
Add VM inventory `CSV export` support

mihaelabalutoiu:
This PR includes the following changes:
* Adds a `PROVIDER_TYPE_ENDPOINT_INVENTORY_EXPORT` capability type that
supports exporting VM inventory as CSV;
* Unit tests


coriolis/api/v1/endpoint_inventory.py
```diff
@@ -0,0 +1,48 @@
...
+            env = utils.decode_base64_param(env, is_json=True)
+        else:
+            env = {}
+
+        csv_content = self._endpoint_resources_api.get_endpoint_inventory_csv(
+            context, endpoint_id, env)
+
+        response = webob.Response()
+        response.status_int = 200
+        response.content_type = 'text/csv'
```
Dany9966: I think this stems from a greater issue, and that is that our wsgi module doesn't support csv type responses: https://github.com/cloudbase/coriolis/blob/master/coriolis/api/wsgi.py#L282

Please look into handling this by adapting the existing wsgi controllers and serializers, instead of manually setting and returning the `Response` object .

Dany9966 (changes requested):

coriolis/api-refs/source/endpoint.inc
```diff
@@ -325,6 +325,64 @@ Response
   .. literalinclude:: ../api_samples/endpoint/endpoint-instance-show-resp.json
      :language: javascript
 
+Get Endpoint VM Inventory
+=========================
+
+.. rest_method:: GET /endpoints/{endpoint_id}/inventory
+
+Exports the full VM inventory of an endpoint as a CSV file.
```
Dany9966:
```suggestion
Exports the full VM inventory of an endpoint in CSV format. 
```

coriolis/api-refs/source/endpoint.inc
```diff
@@ -325,6 +325,64 @@ Response
...
+Get Endpoint VM Inventory
+=========================
+
+.. rest_method:: GET /endpoints/{endpoint_id}/inventory
+
+Exports the full VM inventory of an endpoint as a CSV file. 
+
+**Preconditions**
+
+The endpoint must exist and the platform must support inventory export.
```
Dany9966:
```suggestion
The endpoint must exist and the provider must support inventory export.
```
The platform itself doesn't do the export, the Coriolis provider does :D

coriolis/api-refs/source/endpoint.inc
```diff
@@ -325,6 +325,64 @@ Response
...
+- **num_cpu** – Number of vCPUs.
+- **memory_mb** – Memory in megabytes.
+- **nested_virtualization** – Whether nested virtualisation is enabled.
+- **dynamic_memory_enabled** – Whether dynamic memory is enabled.
+- **secure_boot** – Whether EFI Secure Boot is enabled.
+- **os_type** – Coriolis-normalised OS type (e.g. ``linux``, ``windows``).
+- **disks** – Semicolon-separated list of disk sizes (e.g. ``100GB; 50GB``).
+- **nics** – Semicolon-separated list of NIC entries, each formatted as
+  ``label|mac[|network[|ip1,ip2,...]]``.
+- **boot_order** – Semicolon-separated boot device list. May be empty if the
```
Dany9966: I don't think any of the platforms return boot_order, or whether we should even care about it. I think we should remove this one from the CSV entirely.

coriolis/api-refs/source/endpoint.inc
```diff
@@ -325,6 +325,64 @@ Response
...
+- **vm_name** – Display name of the VM.
+- **guest_id** – Raw VMware guest OS identifier (e.g. ``ubuntu64Guest``).
+- **firmware** – Firmware type (``bios`` or ``efi``).
+- **num_cpu** – Number of vCPUs.
+- **memory_mb** – Memory in megabytes.
+- **nested_virtualization** – Whether nested virtualisation is enabled.
+- **dynamic_memory_enabled** – Whether dynamic memory is enabled.
+- **secure_boot** – Whether EFI Secure Boot is enabled.
+- **os_type** – Coriolis-normalised OS type (e.g. ``linux``, ``windows``).
+- **disks** – Semicolon-separated list of disk sizes (e.g. ``100GB; 50GB``).
```
Dany9966: Also, I didn't catch this on the last review, but only the total size of the disks was required, not the size of every disk the VM has. Can you please adapt it?

coriolis/api-refs/source/endpoint.inc
```diff
@@ -325,6 +325,64 @@ Response
...
+
+.. rest_parameters:: parameters.yaml
+
+   - endpoint_id : endpoint_id_path
+   - env : inventory_source_env
+
+Response
+--------
+
+The response body is a UTF-8 encoded CSV file with the following columns:
```
Dany9966:
```suggestion
The response body is a UTF-8 encoded CSV file with various columns returned.
Each provider may return more specific fields (i.e. parentVapp when using VMWare).
The example below shows the values returned by the VMWare provider.
```

coriolis/api/v1/endpoint_inventory.py
```diff
@@ -0,0 +1,43 @@
...
+        env = req.GET.get("env")
+        if env is not None:
+            env = utils.decode_base64_param(env, is_json=True)
+        else:
+            env = {}
+
+        csv_content = self._endpoint_resources_api.get_endpoint_inventory_csv(
+            context, endpoint_id, env)
+
+        req.environ['coriolis.best_content_type'] = 'text/csv'
```
Dany9966: `best_match_content_type` should already be able to set this in `environ`. You shouldn't need to set this manually.

coriolis/api/wsgi.py
```diff
@@ -426,6 +428,16 @@ def default(self, data):
...
 
 
+class CSVSerializer(DictSerializer):
+    """Serializer for CSV responses. Expects pre-formatted CSV string."""
+
+    def default(self, data):
+        if isinstance(data, str):
+            return data
+        else:
+            return str(data)
```
Dany9966: You can directly return `str(data)` here, no need to check if the data is already of str type.

coriolis/api/v1/endpoint_inventory.py
```diff
@@ -0,0 +1,43 @@
...
+        resp = api_wsgi.ResponseObject(csv_content)
+        resp['content-disposition'] = (
```
Dany9966: This can also be handled inside of `ResponseObject`.
Also, do we have to necessarily set this header? Isn't file save already handled by CLI and UI?

coriolis/api/wsgi.py
```diff
@@ -922,6 +935,11 @@ def _process_stack(self, request, action, action_args,
             except Fault as ex:
                 response = ex
 
+            # Allow the controller to override content-type negotiation
+            # (e.g. for CSV downloads) by updating coriolis.best_content_type
+            # in the request environ before returning.
+            accept = request.environ.get('coriolis.best_content_type', accept)
```
Dany9966: This shouldn't be needed. Please review how you set the content type, and how `best_match_content_type` method works, you may have a bug :D

Dany9966 (changes requested):

coriolis/api/v1/endpoint_inventory.py
mihaelabalutoiu: You have right the CLI and UI handle file saving themselves. Sorry about that

coriolis/api-refs/source/endpoint.inc
mihaelabalutoiu: Yes, I updated disks to return the total provisioned size instead of per-disk sizes.

mihaelabalutoiu (commented):
I updated the code, thank you a lot

Dany9966 (approved):

coriolis/providers/base.py
```diff
@@ -1385,6 +1385,27 @@ def healthcheck_minion(
...
+    required at the provider level.
+    """
+
+    @abc.abstractmethod
+    def export_instance_inventory(
+            self, ctxt, connection_info, source_environment):
+        """Export the full VM inventory as a CSV-formatted string.
+
+        Returns a standards-compliant CSV string with a header row and one
+        row per VM, sorted deterministically by VM ID.
```
Dany9966: Please also document each param of the method, and also what's expected to be returned.

Dany9966 (changes requested):

coriolis/providers/base.py
mihaelabalutoiu: Done, thank you!

PR #403, opened 2026-03-25
Add clustered migration sync for shared disks (SYNCING barrier)

fabi200123:
This PR adds core support for the shared disks clustered transfers:
- Adds a new `SYNCING` task status and a `TASK_TYPES_TO_SYNC` list (`GET_INSTANCE_INFO`, `SHUTDOWN_INSTANCE`). (These tasks wait for their peers across all instances of a clustered transfer before completing)
- Adds a `clustered` flag on `BaseTransferAction`, set automatically when a transfer has more than one instance.
- After all `GET_INSTANCE_INFO` tasks sync, the conductor assigns an `owner` (new property in the VM export info schema) to every disk in each instance's `export_info` (first instance reporting a disk id owns it).
- `REPLICATE_DISKS` now passes the full `volumes_info` to the source provider (skipping shared disks is handled provider-side).
- If a synced task fails, peer tasks waiting in `SYNCING` are failed too (cancellation handles `SYNCING` tasks like pending ones).


coriolis/tasks/replica_tasks.py
```diff
@@ -244,12 +244,35 @@ def _run(self, ctxt, instance, origin, destination, task_info,
...
+                volumes_to_replicate, incremental)
+            schemas.validate_value(
+                replicated_volumes, schemas.CORIOLIS_VOLUMES_INFO_SCHEMA)
+        else:
+            LOG.info(
+                "No disks marked for replication for instance '%s'. "
+                "Using pre-provisioned volumes_info.", instance)
+            replicated_volumes = []
+
+        merged_by_disk_id = {
```
petrutlucian94: Looks a bit odd, why not just `volumes_info = pre_replicated_volumes + replicated_volumes`

fabi200123: We might end up adding the same disk twice if the same disk is in both pre_replicated_volumes and replicated_volumes

petrutlucian94: From what I see, those are two distinct lists:

```
        volumes_to_replicate = [
            vol for vol in volumes_info
            if vol.get(constants.VOLUME_INFO_REPLICATE_DISK_DATA, True)]
        pre_replicated_volumes = [
            vol for vol in volumes_info
            if not vol.get(constants.VOLUME_INFO_REPLICATE_DISK_DATA, True)]
```

fabi200123: Updated it to ```volumes_info = pre_replicated_volumes + replicated_volumes```

coriolis/conductor/rpc/client.py
```diff
@@ -169,7 +169,8 @@ def create_instances_transfer(self, ctxt,
                                   source_environment, destination_environment,
                                   instances, network_map, storage_mappings,
                                   notes=None, user_scripts=None,
-                                  clone_disks=True, skip_os_morphing=False):
+                                  clone_disks=True, skip_os_morphing=False,
+                                  clustered=None):
```
Dany9966: Is this arg really needed here? You can just set it server-side.

coriolis/conductor/rpc/server.py
```diff
@@ -2536,6 +2542,99 @@ def _advance_execution_state(
...
+            if raw_value is None:
+                return None
+            normalized = str(raw_value).strip()
+            if not normalized:
+                return None
+            return normalized.lower()
+
+        def _get_disk_identity(disk_info):
+            disk_info = disk_info or {}
+            for key in ("native_id", "uuid", "path", "name", "disk_id", "id"):
```
Dany9966: I think we should much rather standardize how volumes_info looks like (via schema), if it's not defined already. `disk_id` should always be the source disk identifier, while `shareable` should always be the prop you're looking for, therefore the provider should make these readily available to core, core shouldn't have logic on top of it just to identify which of the disks are shareable and how to identify them.

All of them should already return a disk_id as the source disk ID, but please also double-check. Anyway, it'd be best to standardize/enforce this.

coriolis/conductor/rpc/server.py
```diff
@@ -2536,6 +2542,99 @@ def _advance_execution_state(
...
+                ident = _normalize_identity(disk_info.get(key))
+                if ident:
+                    if key in ("disk_id", "id") and ident.isdigit():
```
Dany9966: Is this check done because the vmware returns device keys instead of actual IDs? If that's the case, please handle this provider-side, so it returns first class disk IDs instead.

coriolis/conductor/rpc/server.py
```diff
@@ -2536,6 +2542,99 @@ def _advance_execution_state(
...
+            task_by_instance = {
+                t.instance: t for t in execution.tasks
+                if t.task_type in constants.TASK_TYPES_REQUIRING_CLUSTER_SYNC}
+            for ident in instance_identities:
+                owner = owners.get(ident)
+                if not owner or owner == instance_id:
+                    continue
+                owner_task = task_by_instance.get(owner)
+                if (
+                        owner_task
```
Dany9966: NIT: Can you please fix this ugliness? :D There's an extra newline above that bothers me.

coriolis/conductor/rpc/server.py
```diff
@@ -2760,6 +2868,174 @@ def _start_task(task):
...
 
+    def _release_cluster_synced_replicate_tasks(
+            self, ctxt, execution, action_id, completed_task):
+        """Unblocks SYNCING replicate tasks after owner completion.
+
+        This is a task-level barrier used for clustered transfers with
+        shared disks. Once the owner replicate task completes, any
+        waiting (SYNCING) replicate tasks are moved back to SCHEDULED
+        so they can continue their normal flow.
+        """
```
Dany9966: So as far as I understand this, this will basically block the rest of the cluster while the main one gets transferred. I mean this is fine for shared disks, but what I proposed initially was to have them all running in parallel once the syncing tasks (which was not meant to be REPLICATE_DISKS btw) are done. You can add SYNCING on this task alone while waiting for the rest of the tasks to complete (tasks like DEPLOY_REPLICA_DISKS, SHUTDOWN_INSTNANCE), that's also feasible, but please set the volumes_info accordingly, not block all of them.

What I originally envisioned was something like this:
instance1 has: root_disk1, shared_disk1;
instance2 has root_diks2, shared_disk1.

What's the point of blocking instance2 while instance1 is replicating? When you can have the following:
instance1 replicates root_disk1, shared_disk1;
instance2 replicates root_disk2, skips shared_disk1, based on the volumes_info that you can set up before launching the replicate_disks task.

coriolis/conductor/rpc/server.py
```diff
@@ -2760,6 +2868,174 @@ def _start_task(task):
...
+
+        def _get_disk_identity(disk_info):
+            disk_info = disk_info or {}
+            for key in ("native_id", "uuid", "path", "name", "disk_id", "id"):
+                ident = _normalize_identity(disk_info.get(key))
+                if ident:
+                    if key in ("disk_id", "id") and ident.isdigit():
+                        continue
+                    return ident
+            return None
```
Dany9966: Did we really have to have these duplicated?

coriolis/conductor/rpc/server.py
Dany9966: Please let me know if there's anything preventing us from doing a parallel sync, with shared_disks being transferred by only one of the instances.

Dany9966 (changes requested):

coriolis/conductor/rpc/server.py
```diff
@@ -2091,7 +2094,8 @@ def _cancel_tasks_execution(
                         "of the parent tasks execution."))
             elif task.status in (
                     constants.TASK_STATUS_PENDING,
-                    constants.TASK_STATUS_STARTING):
+                    constants.TASK_STATUS_STARTING,
+                    constants.TASK_STATUS_SYNCING):
```
Dany9966, line 2098: Wasn't this already handling whatever `_abort_peer_sync_barrier_tasks_on_error` does? If one of the tasks gets an error, the rest of the tasks get cancelled, even if the error is from another instance.

coriolis/utils.py
```diff
@@ -196,6 +196,54 @@ def _exec_retry(*args, **kwargs):
...
 
+def normalized_volume_disk_path_key(disk_id):
+    """Lowercase/stripped path/string key for comparing disk identifiers."""
+    if disk_id is None:
+        return None
+    s = str(disk_id).strip().lower()
+    return s if s else None
+
+
+def cluster_disk_identity(disk_id_or_obj):
```
Dany9966, line 207: If this is just some kind of workaround for the fact that we use disk_keys instead of IDs, I think you should fix this provider-side. The provider should always return a good identifier as the `disk_id` (one that's the same amongst clustered instances). In this case, for VMWare, you should use the disk's FCD ID instead of VM device key (which is arbitrary).

coriolis/conductor/rpc/server.py
```diff
@@ -2782,6 +2810,357 @@ def _update_volumes_info_for_deployment_parent_transfer(
...
+            sorted_volumes = sorted(
+                volumes_info,
+                key=lambda v: (
+                    0
+                    if (
+                        v.get("shareable")
+                        or not v.get(
+                            constants.VOLUME_INFO_REPLICATE_DISK_DATA, True))
+                    else 1))
+            if sorted_volumes != volumes_info:
```
Dany9966, line 3030: Why are volumes re-ordered?

coriolis/conductor/rpc/server.py
```diff
@@ -2782,6 +2810,357 @@ def _update_volumes_info_for_deployment_parent_transfer(
...
+                owner_id = owners.get(ident)
+                if not owner_id or owner_id == instance_id:
+                    continue
+                owner_disk_id = instance_disk_maps.get(owner_id, {}).get(ident)
+                if not owner_disk_id:
+                    continue
+                owner_vol = owner_volumes_by_disk.get(owner_id, {}).get(ident)
+                if not owner_vol:
+                    continue
+                if "change_id" in owner_vol and (
```
Dany9966, line 3079: 1. `change_id` key is not the only incremental key we use/might use;
2. should we update this key for all of the instances? As long as we keep the same owner or `replicate_disk_data` the same between syncs, I don't think so. It's not like users are able to remove the "owner" from the clustered migration anyway, that stays the same, always.

coriolis/schemas/disk_sync_resources_info_schema.json
```diff
@@ -18,7 +18,7 @@
             "description": "String device path (ex: /dev/sdc) from within the temporary minion VM where the disk was attached."
           }
         },
-        "required": ["disk_id", "volume_dev"],
+        "required": ["disk_id"],
```
Dany9966, line 21: Why was `volume_dev` removed?
The `replicate_disks` heavily relies on it, even if there's no shared disks (the most common case).
I think we'd be better off just setting an empty `volume_dev` for the volumes we don't actually transfer (non owners). It's better to enforce it for the most common case instead of removing it entirely imo.

coriolis/tasks/replica_tasks.py
```diff
@@ -572,6 +587,10 @@ def _run(self, ctxt, instance, origin, destination, task_info,
         connection_info = base.get_connection_info(ctxt, destination)
 
         volumes_info = _get_volumes_info(task_info)
+        # NOTE: parallel minion attach may run before conductor merges
+        # shareable.
+        utils.apply_export_disk_shareable_metadata_to_volumes_info(
```
Dany9966, line 592: Is this really necessary? 
How can the minion make attachments if the deploy replica disks tasks are synced?

coriolis/conductor/rpc/server.py
fabi200123: The idea here is to differentiate between an error in the first task, after which this one awaits

coriolis/conductor/rpc/server.py
```diff
@@ -786,6 +791,42 @@ def _begin_tasks(
...
+        if len(transfer.instances) < 2:
+            return
+        deploy_id_by_instance = {
+            t.instance: t.id for t in execution.tasks
+            if t.task_type == constants.TASK_TYPE_DEPLOY_TRANSFER_DISKS}
+        for t in execution.tasks:
+            if t.task_type != constants.TASK_TYPE_REPLICATE_DISKS:
+                continue
+            seen = set()
+            new_deps = []
```
Dany9966: You didn't need to use both a set and a list here. You can start off with a set, add values to it, and then convert it to a list. The order of the dependent tasks don't matter either way.

coriolis/conductor/rpc/server.py
```diff
@@ -786,6 +791,42 @@ def _begin_tasks(
...
 
+    @staticmethod
+    def _apply_clustered_replicate_peer_deploy_dependencies(
+        transfer, execution
+    ):
+        """Order REPLICATE_DISKS after every peer's DEPLOY_TRANSFER_DISKS.
+
+        For clustered transfers, deploy populates volumes_info (including
+        replicate_disk_data for shared disks). Without this, a waiter's
+        REPLICATE_DISKS can start before another instance's deploy finishes,
```
Dany9966: NIT (so it's clearer about what deploy this is about): 
```suggestion
        REPLICATE_DISKS can start before another instance's disks deploy finishes,
```

coriolis/conductor/rpc/server.py
```diff
@@ -813,108 +854,85 @@ def _check_task_cls_param_requirements(task, instance_task_info_keys):
...
-                task_info_keys = task_info_keys.union(set(
-                    modified_fields_by_queued_tasks.keys()))
-                LOG.debug(
-                    "Successfully processed following tasks for instance '%s' "
-                    "for execution %s (type '%s') for any state conflict "
-                    "checks: %s", instance, execution.id, execution.type,
-                    [(t.id, t.task_type) for t in queued_tasks])
+        all_tasks_by_id = {t.id: t for t in execution.tasks}
+        task_info_by_instance = {
+            inst: set((initial_task_info.get(inst) or {}).keys())
```
Dany9966: Above, you already set initial_task_info to `{}` if it's `None`. There's also no point in making this a set, a dict cannot contain duplicate keys by design.
You can just set this to:

```suggestion
            inst: initial_task_info.get(inst, {}).keys()
```

coriolis/conductor/rpc/server.py
```diff
@@ -813,108 +854,85 @@ def _check_task_cls_param_requirements(task, instance_task_info_keys):
...
+            for inst in all_instances_in_tasks}
+
+        tasks_to_process = {t.id: t for t in execution.tasks}
+        if not tasks_to_process:
+            return
+        processed_tasks = {}
```
Dany9966: This also could've easily been just a list or set of task IDs, no need to make it a dict.

coriolis/conductor/rpc/server.py
```diff
@@ -943,6 +961,11 @@ def execute_transfer_tasks(self, ctxt, transfer_id, shutdown_instances,
         dest_env['network_map'] = transfer.network_map
         dest_env['storage_mappings'] = transfer.storage_mappings
 
+        # Serialized DEPLOY for clustered transfers so the first instance
+        # (owner) materializes libvirt volumes before other instances run
+        # deploy_replica_disks; _setup can then copy pool/volume to waiters.
```
Dany9966: No, core shouldn't take into account platform specific stuff. Please re-phrase this comment.

coriolis/conductor/rpc/server.py
```diff
@@ -1025,9 +1048,16 @@ def execute_transfer_tasks(self, ctxt, transfer_id, shutdown_instances,
                 disk_deployment_depends_on.append(
                     validate_transfer_destination_inputs_task.id)
 
+            disk_deploy_deps = list(disk_deployment_depends_on)
```
Dany9966: Since `disk_deployment_depends_on` is already defined as a list and never gets changed into something else (there just some appends), what's the point of converting this to a list again?

coriolis/conductor/rpc/server.py
```diff
@@ -2819,6 +2882,323 @@ def _update_volumes_info_for_deployment_parent_transfer(
...
 
+    def _clustered_shared_disk_export_maps(self, action):
+        """Build per-instance export disk maps and shared cluster identities.
+
+        Returns ``(instance_disk_maps, shared_identities)`` where each
+        ``instance_disk_maps[instance_id]`` maps ``cluster_disk_identity``
+        to that disk's export ``id`` from ``export_info``.
+        """
+        identity_counts = {}
+        explicitly_shareable = {}
```
Dany9966: You shouldn't check explicit `shareable` flag and count it. What if there's a disk on one of the instances that has shareable enabled but it's not actually attached to other instances? That shouldn't be counted as a clustered disk. 
The fact that the source provider has to handle the transfer if the shareable is enabled should be, obviously, handled in said provider, not here.
Please remove this var along with whatever logic is related to it. Only count actual clustered disks amongst selected transfer instances.

coriolis/conductor/rpc/server.py
```diff
@@ -2819,6 +2882,323 @@ def _update_volumes_info_for_deployment_parent_transfer(
...
+        instance_disk_maps = {}
+        for instance_id in action.instances:
+            export_info = action.info.get(instance_id, {}).get(
+                "export_info", {})
+            disks = export_info.get("devices", {}).get("disks", [])
+            disk_map = {}
+            for disk in disks:
+                ident = utils.cluster_disk_identity(disk or {})
```
Dany9966: disk['id'] should be ok to use here. `disk_id` is something that usually resides in `volumes_info` actually, so it's not useful here.

coriolis/conductor/rpc/server.py
```diff
@@ -2819,6 +2882,323 @@ def _update_volumes_info_for_deployment_parent_transfer(
...
+                if not ident:
+                    continue
+                disk_map[ident] = disk.get("id")
```
Dany9966: This will also therefore be irrelevant. A `disk_id: disk_id` mapping is useless :D

coriolis/conductor/rpc/server.py
```diff
@@ -2819,6 +2882,323 @@ def _update_volumes_info_for_deployment_parent_transfer(
...
+                identity_counts[ident] = identity_counts.get(ident, 0) + 1
```
Dany9966: A `collections.Counter` would be perfect for this: https://docs.python.org/3/library/collections.html#collections.Counter
It's almost the same thing, but a bit more straight-forward and safe to use.
```suggestion
        identity_counts = collections.Counter()
        # ....
                identity_counts[ident] += 1
        # .....
        # Accessing a counter item, even if never added, never throws a KeyError, it just returns 0.
        # Also, that never-added key is never saved in `counter.items()` unless it has at least counted once.
        shared_identities = {ident for ident, count in identity_counts.items() if count > 1}
        # can be the same as
        shared_identities = {ident for ident in identity_counts if identity_counts[ident] > 1}
```

coriolis/conductor/rpc/server.py
```diff
@@ -2819,6 +2882,323 @@ def _update_volumes_info_for_deployment_parent_transfer(
...
+            if not export_info:
+                continue
+            disks = export_info.get("devices", {}).get("disks")
+            if not disks:
+                continue
+            updated = False
+            for disk in disks:
+                ident = utils.cluster_disk_identity(disk or {})
+                if ident in shared_idents and not disk.get("shareable"):
+                    disk["shareable"] = True
```
Dany9966: Do we actually need this "promotion"?
Shouldn't the export provider already properly set this flag? 
Is this like a safeguard in case the provider doesn't flag this, but core identifies that a disk is actually shared?

coriolis/conductor/rpc/server.py
```diff
@@ -2822,6 +2885,323 @@ def _update_volumes_info_for_deployment_parent_transfer(
...
+        Returns True if the barrier was activated (caller should return
+        early), False otherwise.
+        """
+        if task.task_type not in constants.TASK_TYPES_TO_SYNC:
+            return False
+        if not bool(getattr(action, "clustered", False)):
+            return False
+        if task.task_type == constants.TASK_TYPE_DEPLOY_TRANSFER_DISKS:
+            # Deploy tasks are ordered per instance (chained depends_on) so
+            # libvirt can reuse the owner volume. The barrier would
```
Dany9966: We should not mention any provider names here, these functions are plugin-agnostic.

coriolis/conductor/rpc/server.py
```diff
@@ -3138,11 +3518,35 @@ def task_completed(self, ctxt, task_id, task_result):
 
             # NOTE: refresh the execution just in case:
             execution = db_api.get_tasks_execution(ctxt, task.execution_id)
+
+            # Copy owner's libvirt / replicate flags to waiters as soon as the
```
Dany9966: Same as above.

coriolis/conductor/rpc/server.py
```diff
@@ -3138,11 +3518,35 @@ def task_completed(self, ctxt, task_id, task_result):
...
+            # owner finishes DEPLOY, before the next instance's deploy runs.
+            if (
+                    task_result
+                    and task.task_type == (
+                        constants.TASK_TYPE_DEPLOY_TRANSFER_DISKS)
+                    and getattr(action, "clustered", False)):
+                self._setup_shared_disk_volumes_info(ctxt, execution)
```
Dany9966: Do we need this both here and in `_handle_synced_tasks`?

coriolis/tasks/replica_tasks.py
```diff
@@ -244,12 +276,27 @@ def _run(self, ctxt, instance, origin, destination, task_info,
...
+                ctxt, connection_info, source_environment, instance,
+                source_resources, migr_source_conn_info, migr_target_conn_info,
+                volumes_to_replicate, incremental)
+            schemas.validate_value(
+                replicated_volumes, schemas.CORIOLIS_VOLUMES_INFO_SCHEMA)
+        else:
+            LOG.info(
+                "No disks marked for replication for instance '%s'. "
+                "Using pre-provisioned volumes_info.", instance)
+            replicated_volumes = []
```
Dany9966: I'd prefer if we'd have this value initialization before the `if`, so we avoid an eventual "reference before assignment"

coriolis/tasks/replica_tasks.py
```diff
@@ -23,6 +23,38 @@ def _get_volumes_info(task_info):
...
 
 
+def _preserve_replicate_disk_data_from_prior_volumes(
+        prior_volumes_info, new_volumes_info):
+    """Keep replicate_disk_data=False set by the conductor for cluster waiters.
+
+    update_transfer_action_info_for_instance replaces the whole
+    volumes_info list when a task returns. Destination deploy does not
+    round-trip that flag, so shared-disk waiters would lose it and replicate
+    again unless we merge it back per disk_id.
```
Dany9966: Which providers do this? I don't know of any that rewrite the whole volumes_info in deploy_replica_disks task. The `deploy_replica_instance` will return a different volumes info inside of `instance_deployment_info`, but that isn't used for transfers regardless and doesn't overwrite the original `action.info[instance]['volumes_info']`.

coriolis/utils.py
```diff
@@ -199,6 +199,90 @@ def _exec_retry(*args, **kwargs):
     return _retry_on_error
 
 
+def normalized_volume_disk_path_key(disk_id):
```
Dany9966: Not sure this is needed.

Dany9966 (changes requested):

coriolis/conductor/rpc/server.py
fabi200123: `.keys()` creates a `dict_keys`, which breaks the below `task_info_by_instance[_task.instance].add(new_field)`

coriolis/conductor/rpc/server.py
```diff
@@ -2827,6 +2889,284 @@ def _update_volumes_info_for_deployment_parent_transfer(
...
+                continue
+
+            volumes_info = list(
+                action.info.get(instance_id, {}).get("volumes_info", []) or [])
+
+            updated = False
+            for disk_id in shared_disk_ids:
+                ident = utils.cluster_disk_identity(disk_id)
+                owner_id = owners.get(ident)
+                if owner_id == instance_id or not owner_id:
```
Dany9966: Shouldn't it be guaranteed that we get an owner_id? I feel like this should error out if there's no owner returned.

coriolis/conductor/rpc/server.py
```diff
@@ -2827,6 +2889,284 @@ def _update_volumes_info_for_deployment_parent_transfer(
...
+                inherited["shareable"] = True
+                inherited["volume_dev"] = ""
+
+                target_vol = None
+                for vol in volumes_info:
+                    if utils.cluster_disk_identity(vol) == ident:
+                        target_vol = vol
+                        break
+                if target_vol is None:
+                    norm_wid = utils.cluster_disk_identity(
```
Dany9966: Why not use `ident` instead of assigning a new `norm_wid`? It's the same value.

coriolis/conductor/rpc/server.py
```diff
@@ -2827,6 +2889,284 @@ def _update_volumes_info_for_deployment_parent_transfer(
...
+                        disk_id)
+                    for vol in volumes_info:
```
Dany9966: This for loop is pointless. The same thing is already checked above (since `ident == norm_wid`).

Dany9966 (approved):

coriolis/conductor/rpc/server.py
```diff
@@ -2130,13 +2110,15 @@ def _cancel_tasks_execution(
...
                         "of the parent tasks execution."))
+            # NOTE: SYNCING covers tasks waiting on a clustered
+            # barrier; there is no running worker to cancel, so we
+            # unschedule like PENDING/STARTING. (Peers stuck in
+            # SYNCING when another instance fails are handled in
+            # _abort_peer_sync_barrier_tasks_on_error, not here.)
             elif task.status in (
                     constants.TASK_STATUS_PENDING,
-                    constants.TASK_STATUS_STARTING):
-                # any PENDING/STARTING tasks means that they did not have a
```
petrutlucian94, line 2136: nit: we may want to keep the previous inline comment as well, it explains why it's safe to mark these tasks as unscheduled.

coriolis/conductor/rpc/server.py
```diff
@@ -813,108 +813,85 @@ def _check_task_cls_param_requirements(task, instance_task_info_keys):
...
-                            task, task_info_keys):
-                        if new_field not in modified_fields_by_queued_tasks:
-                            modified_fields_by_queued_tasks[new_field] = [
-                                task]
-                        else:
-                            modified_fields_by_queued_tasks[new_field].append(
-                                task)
-
-                # check if any queued tasks would manipulate the same fields:
-                conflicting_fields = {
```
petrutlucian94, line 887: `_check_execution_tasks_sanity` is probably the most insane sanity check I've seen :). It's quite difficult to follow, but I see that `for instance, instance_tasks in instances_tasks_mapping.items():` was updated so that we're no longer grouping tasks by instance. I guess it's because of clustered migrations were tasks affecting separate instances may have common dependencies.

coriolis/constants.py
```diff
@@ -55,11 +55,13 @@
 TASK_STATUS_CANCELED_FROM_DEADLOCK = "STRANDED_AFTER_DEADLOCK"
 TASK_STATUS_ON_ERROR_ONLY = "EXECUTE_ON_ERROR_ONLY"
 TASK_STATUS_FAILED_TO_SCHEDULE = "FAILED_TO_SCHEDULE"
+TASK_STATUS_SYNCING = "SYNCING"
```
petrutlucian94, line 58: We should add an inline comment describing the "syncing" status, e.g.
```
The "SYNCING" status used by clustered migrations that reach a barrier, waiting for the other cluster migrations to complete.
```

The `SYNCING` name is quite generic too, it doesn't really emphasize the intended behavior. I would've named it `TASK_STATUS_WAITING_FOR_PEER_JOBS` or `TASK_STATUS_CLUSTER_BARRIER`.

coriolis/conductor/rpc/server.py
```diff
@@ -2827,6 +2833,137 @@ def _update_volumes_info_for_deployment_parent_transfer(
...
+
+        db_api.set_task_status(
+            ctxt, task.id, constants.TASK_STATUS_SYNCING)
+
+        peer_tasks = [
+            t for t in execution.tasks
+            if t.task_type == task.task_type and t.id != task.id]
+        all_syncing = all(
+            t.status == constants.TASK_STATUS_SYNCING for t in peer_tasks)
+        if not all_syncing:
```
petrutlucian94, line 2862: There's a small race condition risk: two tasks may reach `TASK_STATUS_SYNCING` at the same time, see that there are pending tasks and hang in `SYNCING` state forever.

petrutlucian94 (commented):
The PR merged while I was still reviewing it, I'll drop some comments anyway.

PR #404, opened 2026-04-01
Document the provider interface

petrutlucian94:
We'll document the Coriolis provider interface, making it much easier to write new providers or extend the existing ones, with or without AI assistance.

At the moment, most of the provider API is undocumented. It's true that we have JSON schemas for some of the dicts, but not all of them are covered. Furthermore, the intended provider behavior isn't always immediately obvious.

Let's take "deploy_replica_target_resources" for example, it's meant to deploy a minion instance and return the minion connection info, if minion instances are needed by this import provider. Somebody writing a provider would have a hard time guessing that. We intend to solve this.

While at it, we'll add type annotations.

Also, we're cleaning up unused deprecated base classes, otherwise new providers may end up implementing these APIs unnecessarily. We'll keep the class definitions without any abstract methods for backwards compatibility, in case there are providers that still inherit them.


coriolis/providers/base.py
```diff
@@ -538,114 +1106,351 @@ def check_update_destination_environment_params(
...
 
     @abc.abstractmethod
     def start_minion(
-            self, ctxt, connection_info, minion_properties):
+        self,
+        ctxt: context.RequestContext,
+        connection_info: dict,
+        minion_properties: dict,
+    ):
+        """Shutdown the minion instance.
```
Dany9966: NIT:
```suggestion
        """Start the minion instance.
```

Dany9966 (commented):
LGTM, aside from that one nit.

coriolis/providers/base.py
petrutlucian94: Whops, thanks.

Dany9966 (approved):

PR #406, opened 2026-04-06
Add post-transfer disk checksum verification

claudiubelu:
After each disk is written, compare the source (replicator) and destination (writer) checksums to catch any data corruption during transfer. The comparison happens while the writer device is still acquired so the checksum job can run.

- `Client.get_disk_checksum`: calls GET `/api/v1/dev/{disk}/checksum`

- `HTTPBackupWriterImpl`:
  - `_create_checksum_job`: calls POST `/api/v2/device/{disk}/checksumJob`
  - `_delete_checksum_job`: calls DELETE `/api/v2/device/{disk}/checksumJob/{id}`
  -` _get_checksum_job_status`: calls GET `/api/v2/device/{disk}/checksumJob/{id}`
  - `get_disk_checksum`: creates checksum job, waits for it to finish, and returns the checksum value and algorithm.

- `Replicator._verify_disk_checksum`: compares both sides, raises on algorithm or value mismatch.

### DIFF
--- a/coriolis/providers/backup_writers.py
+++ b/coriolis/providers/backup_writers.py
@@ -31,6 +31,11 @@ opts = [
     cfg.BoolOpt('compress_transfers',
                 default=True,
                 help='Use compression if possible during disk transfers'),
+    cfg.IntOpt('disk_checksum_timeout',
+               default=3600,
+               help='Maximum number of seconds to wait for a disk checksum '
+                    'job to complete on the backup writer. Larger disks may '
+                    'require a higher value.'),
 ]
 CONF.register_opts(opts)
 _CORIOLIS_HTTP_WRITER_CMD = "coriolis-writer"
@@ -65,6 +70,10 @@ _WRITER_ERR_MAP = {
     14: "ERR_WRITE_MSG_ID",
     15: "ERR_OUT_OF_BOUDS",
 }
+
+_CHECKSUM_JOB_POLL_INTERVAL = 15  # seconds between writer checksum job polls
+_CHECKSUM_JOB_FINISHED = "finished"
+_CHECKSUM_JOB_FAILED = "failed"
 
 
 def _disable_lvm2_lvmetad(ssh):
@@ -192,6 +201,18 @@ class BaseBackupWriterImpl(with_metaclass(abc.ABCMeta)):
     @abc.abstractmethod
     def close(self):
         pass
+
+    def get_disk_checksum(self, algorithm, start_offset=0, end_offset=0):
+        """Returns the destination disk checksum, or None if unsupported.
+
+        :param algorithm: Checksumming algorithm to use.
+        :param start_offset: Checksumming starts from this offset.
+        :param end_offset: Checksumming stops at this offset. If it is 0,
+          the end of the disk is considered instead.
+        :returns: dict with 'checksum' and 'algorithm' keys, or None if
+          unsupported.
+        """
+        return None
 
 
 class BaseBackupWriter(with_metaclass(abc.ABCMeta)):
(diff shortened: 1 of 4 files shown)


coriolis/providers/backup_writers.py
```diff
@@ -735,6 +749,95 @@ def _wait_for_queues(self):
...
+
+        :returns: job ID string.
+        """
+        self._ensure_session()
+        uri = "%s/checksumJob" % self._uri
+        headers = {"X-Client-Token": self._id}
+        body = {
+            "start_offset": 0,
+            "end_offset": 0,
+            "checksum_algorithm": algorithm,
```
Dany9966: start_offset/end_offset should also be configurable. Default to 0. 

There are cases where we add a 1GB extra padding to destination disks, in that case we'll need to specifically pass in an end_offset, so that the destination hasher stops there. The padding was added because in some cases, the source provider can't accurately report the exact size of the source disks (i.e. the in-guest disk is larger than what the platform reports by a small amount).

coriolis/providers/replicator.py
```diff
@@ -845,6 +895,7 @@ def replicate_disks(self, source_volumes_info, backup_writer):
                     total += 1
                     self._event_manager.set_percentage_step(
                         perc_step, total)
+                self._verify_disk_checksum(devName, destination)
```
Dany9966: This shouldn't always be called. Please introduce a switch to this `replicate_disks` method. The provider will control whether or not verification needs to run.

coriolis/resources/bin/coriolis-writer
Dany9966, line 1: This may not be needed anymore: https://github.com/cloudbase/coriolis/pull/407

Dany9966 (changes requested):

coriolis/providers/replicator.py
claudiubelu: Done

coriolis/resources/bin/coriolis-writer
claudiubelu: Done

coriolis/providers/backup_writers.py
claudiubelu: Done

coriolis/providers/backup_writers.py
```diff
@@ -735,6 +748,103 @@ def _wait_for_queues(self):
...
+        """
+        self._wait_for_queues()
+        if self._exception:
+            raise exception.CoriolisException(
+                "Cannot checksum disk '%s', pending write error: %s" % (
+                    self._path, self._exception))
+
+        timeout = CONF.disk_checksum_timeout
+        deadline = time.monotonic() + timeout
+        job_id = self._create_checksum_job(algorithm)
```
Dany9966: Please also pass the start_offeset, end_offset args down the stack :D

coriolis/providers/backup_writers.py
```diff
@@ -193,6 +202,10 @@ def write(self, data):
     def close(self):
         pass
 
+    def get_disk_checksum(self, algorithm):
```
Dany9966: Add the `start_offset/end_offset` kwargs here as well please.

Dany9966 (changes requested):

coriolis/providers/backup_writers.py
claudiubelu: Done

coriolis/providers/backup_writers.py
claudiubelu: Done

Dany9966 (approved):

PR #410, opened 2026-04-09
Skips retrying on "command not found" errors

claudiubelu:
`exec_ssh_cmd` will retry the given cmd in case an error occurs. But if the command does not exist, there's no need to retry.

### DIFF
--- a/coriolis/utils.py
+++ b/coriolis/utils.py
@@ -311,7 +311,9 @@ def list_ssh_dir(ssh, remote_path):
         return [f for f in output.splitlines() if f.strip()]
 
 
-@retry_on_error(terminal_exceptions=[exception.MinionMachineCommandTimeout])
+@retry_on_error(terminal_exceptions=[
+    exception.MinionMachineCommandTimeout,
+    exception.SSHCommandNotFoundException])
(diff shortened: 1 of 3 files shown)


coriolis/utils.py
```diff
@@ -335,12 +337,16 @@ def exec_ssh_cmd(ssh, cmd, environment=None, get_pty=False, timeout=None):
...
+        stdout_str = std_out.decode(errors='ignore')
+        stderr_str = std_err.decode(errors='ignore')
+        msg = (
             "Command \"%s\" failed on host '%s' with exit code: %s\n"
             "stdout: %s\nstd_err: %s" %
-            (cmd, remote_str, exit_code,
-             std_out.decode(errors='ignore'),
-             std_err.decode(errors='ignore')))
+            (cmd, remote_str, exit_code, stdout_str, stderr_str))
+        if ("command not found" in stdout_str or
```
Dany9966: I think it might be better to check `exit_code == 127` instead of reliying on the stdout/stderr output. What if something outputs that exact message?

Dany9966 (changes requested):

coriolis/utils.py
claudiubelu: Done

petrutlucian94 (approved):

PR #413, opened 2026-04-22
Adds Coriolis integration tests

claudiubelu:
Introduces the skeleton for integration tests that exercise the full Coriolis migration pipeline without requiring RabbitMQ, Keystone, or Barbican.

- `eventlet.monkey_patch()` needs to be called (also called in `coriolis/cmd/__init__.py`).
- use `fake://` `messaging_transport_url`, effectively using an in-process fake messaging queue, instead of relying on an external service.
- use a no-auth middleware, removing the need for Keystone.
- use a test APIRouter, which won't have the /project_id/ prefix.
- use an in-process worker server, which will not spawn subprocesses. This is needed; otherwise Coriolis would hang waiting for workers to reply (the messaging transport url is fake).
- Uses `python-coriolisclient` to send requests to Coriolis.
- Adds a few smoke tests, which lightly exercises the Coriolis API, ensuring the harness works.
- Adds a light integration tests `TestProvider`, which implements the base provider interfaces.
- Adds replica transfer test, deployment test, transfer failure test.
- Adds integration entry in tox (`sudo tox -e integration`)
- Adds the integration tests GitHub Action.


coriolis/tests/integration/providers/test_provider/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+The Coriolis provider factory resolves a single dotted class path and queries
+it for every provider type it implements. A single entry in CONF.providers
+satisfies both the source and destination side of a replica execution.
+"""
+
+from coriolis.tests.integration.providers.test_provider import exp
+from coriolis.tests.integration.providers.test_provider import imp
+
+
+class TestProvider(exp.TestExportProvider, imp.TestImportProvider):
```
petrutlucian94: Is there a reason why we need a class that inherits both the export and import provider? It might be less confusing and safer to keep them separate.

You can use a base class for common code, if necessary.

coriolis/tests/integration/providers/test_provider/exp.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def __init__(self, event_handler):
+        self._event_handler = event_handler
+
+    def _event_manager(self):
+        return events.EventManager(self._event_handler)
+
+    def _make_replicator(self, pkey_path, event_mgr, volumes_info, repl_state):
+        pkey = paramiko.RSAKey.from_private_key_file(pkey_path)
+        conn_info = {
```
petrutlucian94, line 53: So we have a test provider that points to localhost instead of a minion vm. This is very likely to pollute the test environment.

Since we're planning to use a container or vm in a subsequent PR, I won't block it.

coriolis/tests/integration/providers/test_provider/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+    """Single provider class that acts as both source and destination.
+
+    Inherits all abstract method implementations from TestExportProvider
+    and TestImportProvider. The MRO ensures that each method resolves to the
+    correct side's implementation without any overrides needed here.
+    """
+
+    platform = "foo"
```
petrutlucian94: Let's call it "mock-platform" instead of "foo". It will be less confusing if it gets logged somewhere.

coriolis/tests/integration/providers/test_provider/exp.py
```diff
@@ -0,0 +1,217 @@
...
+            "type": "object",
+            "properties": {
+                "device": {"type": "string"},
+                "pkey_path": {"type": "string"},
+            },
+            "required": ["device", "pkey_path"],
+        }
+
+    def validate_connection(self, ctxt, connection_info):
+        device = connection_info["device"]
```
petrutlucian94: "device" is a bit too generic, how about "block_device_path"?

also, what if we'll want to test multiple block devices?

coriolis/tests/integration/providers/test_provider/exp.py
```diff
@@ -0,0 +1,217 @@
...
+            },
+        }
+
+    def delete_replica_source_resources(
+            self, ctxt, connection_info, source_environment,
+            migr_resources_dict):
+        pkey_path = connection_info.get("pkey_path")
+        if not pkey_path:
+            return
+        replicator = self._make_replicator(
```
petrutlucian94, line 150: Let's add `TODO: use a container` to at least state our plan.

coriolis/tests/integration/providers/test_provider/exp.py
```diff
@@ -0,0 +1,217 @@
...
+
+    # BaseReplicaExportValidationProvider
+
+    def validate_replica_export_input(
+            self, ctxt, connection_info, instance_name, source_environment):
+        return {}
+
+
+# Helpers
+def _block_device_size(device):
```
petrutlucian94: nit: let's call it '_get_block_device_size' for clarity.

coriolis/tests/integration/harness.py
```diff
@@ -0,0 +1,557 @@
+# Copyright 2026 Cloudbase Solutions Srl
+# All Rights Reserved.
+
+"""
+Base test harness for Coriolis integration tests.
```
petrutlucian94, line 5: We should have a readme section describing the integration test workflow and prerequisites.

It can be part of the top level readme or a separate readme under `coriolis/tests/integration`.

coriolis/tests/integration/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2026 Cloudbase Solutions Srl
+# All Rights Reserved.
+
+import eventlet
```
petrutlucian94: Eventlet is about to be removed, let's catch import errors and skip monkey patching if it's missing.

coriolis/tests/integration/harness.py
```diff
@@ -0,0 +1,557 @@
...
+                pass
+
+        shutil.rmtree(self.workdir, True)
+        try:
+            test_utils.destroy_scsi_debug()
+        except Exception:
+            pass
+
+
+class CoriolisIntegrationTestBase(test_base.CoriolisBaseTestCase):
```
petrutlucian94: I'd move the base classes to a separate module, e.g. base or test_base.

It would improve readability. Right now, people will need to go through all that harness code to find what the test base class provides.

coriolis/tests/integration/harness.py
```diff
@@ -0,0 +1,557 @@
...
+        sqlalchemy_api._facade = None
+        rpc_module._TRANSPORT = None
+
+        engine = db_api.get_engine()
+        db_migration.db_sync(engine)
+
+        # SQLite does not support bulk UPDATE ... FROM ... for
+        # joined-table-inheritance models; replace the production function
+        # with a per-object alternative for the lifetime of this process.
+        db_api._delete_transfer_action = _sqlite_delete_transfer_action
```
petrutlucian94, line 234: We're using quite a few hacks in order to run the Coriolis services manually, without an amqp broker or mysql database (using sqlite). It also skips the standard service launch commands, which would go untested.

On the other hand, it allows us to run the integration tests with a minimal configuration.

I'd say it's a reasonable compromise.

coriolis/tests/integration/harness.py
claudiubelu: Done, added `coriolis/tests/integration/README.md`.

coriolis/tests/integration/harness.py
claudiubelu: Right. Running with other services might be interesting as well, unintended interactions / issues could be caught doing so. Maybe some time in the future we'd set those up as well. But for the time being, these are sufficient, it allows us to test the project internals.

Though, the sqlite issue could eventually be addressed, if we'd want to add support for that, but it's beyond the scope of this PR.

coriolis/tests/integration/harness.py
claudiubelu: Done

coriolis/tests/integration/__init__.py
claudiubelu: Done

coriolis/tests/integration/providers/test_provider/__init__.py
claudiubelu: Done. Since it's split, it's now `test-src` and `test-dest`.

coriolis/tests/integration/providers/test_provider/__init__.py
claudiubelu: Done

coriolis/tests/integration/providers/test_provider/exp.py
claudiubelu: Done

coriolis/tests/integration/providers/test_provider/exp.py
claudiubelu: Done

coriolis/tests/integration/providers/test_provider/exp.py
claudiubelu: Updated to `block_device_path`. Multiple block devices will be interesting to also test. Will add it to the list. But for now, we can leave it as is.

petrutlucian94 (approved):
LGTM, thanks!

PR #414, opened 2026-04-24
Drop eventlet

petrutlucian94:
Eventlet is now deprecated, about to become unsupported.

Just like OpenStack, we're going to switch from Eventlet to native threading.

* https://wiki.openstack.org/wiki/Eventlet-removal
* https://removal.eventlet.org/guide/openstack/

This PR will:

* drop the eventlet dependency
* stop monkey-patching standard libraries
* spawn native threads instead of eventlet coroutines
	* added a few TODOs to merge some of them since threads are more heavyweight compared to coroutines
	* asyncio may be applicable in some cases but only if we can guarantee that the callback doesn't perform any blocking operations
* use native coordination and communication primitives such as semaphores, locks or queues
* instead of eventlet.kill, we use stop variables or events to notify the threads that the work is done
* tell oslo.messaging and oslo.service to use threading instead of eventlet
* update coriolis-api to use cheroot (same as Ironic) instead of the eventlet based wsgi.
  * Note that we still allow using an external web server such as Apache, in which case the Coriolis wsgi app will be loaded insted of calling "coriolis-api".

The integration tests that merged in the meantime also need a few changes:

* switch from sqlite to a mysql container since sqlite doesn't allow concurrent writes and there are a few hacks to accommodate it.


Dany9966:
LGTM so far!

coriolis/providers/backup_writers.py
```diff
@@ -660,8 +689,13 @@ def _ensure_session(self):
             return
 
     def _compressor(self):
-        while True:
-            payload = self._comp_q.get()
+        LOG.debug("Backup comperssor started.")
```
Dany9966: NIT:
```suggestion
        LOG.debug("Backup compressor started.")
```

coriolis/providers/backup_writers.py
```diff
@@ -674,17 +708,23 @@ def _compressor(self):
                     if compressed:
                         send_payload["encoding"] = 'gzip'
                 except BaseException as err:
-                    LOG.exception(err)
+                    LOG.exception("Backup compressor failure.")
```
Dany9966, line 711: Should we hide the original error here?

petrutlucian94: LOG.exception will print the entire trace, including the original exception message.

I don't think we need to hide it, the trace can be useful.

coriolis/tests/integration/harness.py
```diff
@@ -243,31 +245,46 @@ def __init__(self):
...
+            [
+                "docker",
+                "run",
+                "-d",
+                "--name",
+                self._mysql_container_name,
+                "-e",
+                f"MYSQL_ROOT_PASSWORD={self._mysql_password}",
+                "-e", f"MYSQL_DATABASE={self._mysql_database}",
+                "-p", "3306:3306",
```
claudiubelu: nit: could we map a different host port, something like `13306` instead? If I'm running a local mysql server, I won't be able to run these tests.

petrutlucian94: Makes sense, sure!

coriolis/tests/integration/harness.py
```diff
@@ -278,7 +278,7 @@ def _start_db_container(self):
                 "-e",
                 f"MYSQL_ROOT_PASSWORD={self._mysql_password}",
                 "-e", f"MYSQL_DATABASE={self._mysql_database}",
-                "-p", "3306:3306",
+                "-p", "13306:13306",
```
claudiubelu: it should be `13306:3306`

petrutlucian94: ouch, thanks.

claudiubelu (approved):

PR #416, opened 2026-04-27
Preserve `interface name` and `NetworkAddress` on Windows migrations

mihaelabalutoiu:
When `set_dhcp` is false, extract `DriverDesc` and `NetworkAddress` from the offline registry and use `Rename-NetAdapter` to restore the source adapter name on the destination.

### DIFF
--- a/coriolis/osmorphing/windows.py
+++ b/coriolis/osmorphing/windows.py
@@ -77,6 +77,9 @@ REQUIRED_DETECTED_WINDOWS_OS_FIELDS = [
 INTERFACES_PATH_FORMAT = (
     "HKLM:\\%s\\ControlSet001\\Services\\Tcpip\\Parameters\\Interfaces")
 
+DEVICE_CLASS_BASE_PATH_FORMAT = (
+    "HKLM:\\%s\\ControlSet001\\Control\\Network")
+
 STATIC_IP_SCRIPT_TEMPLATE = """
 $ErrorActionPreference = "Stop"
 
@@ -175,6 +178,16 @@ function Invoke-Main {
         Set-DnsClientServerAddress -InterfaceIndex $interface.IfIndex -ResetServerAddresses
 
         Set-IPAddresses $interface $nic.ip_addresses $ips_info
+
+        if ($nic.PSObject.Properties['interface_name'] -and $nic.interface_name) {
+            $conflictAdapter = Get-NetAdapter -Name $nic.interface_name -ErrorAction SilentlyContinue
+            if ($conflictAdapter -and $conflictAdapter.IfIndex -ne $interface.IfIndex) {
+                Write-Host "Removing conflicting adapter '$($conflictAdapter.Name)' to free up name '$($nic.interface_name)'"
+                Remove-NetAdapter -Name $conflictAdapter.Name -Confirm:$false
+            }
+            Rename-NetAdapter -Name $interface.Name -NewName $nic.interface_name
+            Write-Host "Renamed network adapter '$($interface.Name)' to '$($nic.interface_name)'"
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


coriolis/osmorphing/windows.py
```diff
@@ -77,6 +77,9 @@
 INTERFACES_PATH_FORMAT = (
     "HKLM:\\%s\\ControlSet001\\Services\\Tcpip\\Parameters\\Interfaces")
 
+NET_ADAPTER_CLASS_BASE_PATH_FORMAT = (
```
Dany9966: This should actually be `DEVICE_CLASS_BASE_PATH_FORMAT`. Please rename it.

coriolis/osmorphing/windows.py
```diff
@@ -175,6 +178,16 @@
         Set-DnsClientServerAddress -InterfaceIndex $interface.IfIndex -ResetServerAddresses
 
         Set-IPAddresses $interface $nic.ip_addresses $ips_info
+
+        if ($nic.PSObject.Properties['interface_name'] -and $nic.interface_name) {
+            $conflictAdapter = Get-NetAdapter -Name $nic.interface_name -ErrorAction SilentlyContinue
+            if ($conflictAdapter -and $conflictAdapter.IfIndex -ne $interface.IfIndex) {
+                Write-Host "Cannot rename adapter '$($interface.Name)' to '$($nic.interface_name)': name already in use by another adapter"
```
Dany9966: This will almost always fail. We have to also delete the original NIC after we successfully gather all the information about it.
Deleting the source NIC will also help with Windows warnings like "Static IP conflict" that appears on the UI when the same static IP is set on the destination NIC.

coriolis/osmorphing/windows.py
```diff
@@ -599,12 +612,46 @@ def _install_cloudbase_init(self, download_url,
 
         return cloudbaseinit_base_dir
 
+    def _get_guid_to_adapter_info_map(self, key_name):
+        class_base_path = NET_ADAPTER_CLASS_BASE_PATH_FORMAT % key_name
+        ps_command = (
+            "$ErrorActionPreference = 'SilentlyContinue'; "
+            "$classBase = '%(base)s'; "
+            "$netClassGuid = (Get-ChildItem $classBase "
+            "-ErrorAction SilentlyContinue | "
```
Dany9966: Do we have to pass the error action twice?

coriolis/osmorphing/windows.py
```diff
@@ -599,12 +612,46 @@ def _install_cloudbase_init(self, download_url,
...
+            "ForEach-Object { $p = Get-ItemProperty $_.PSPath "
+            "-ErrorAction SilentlyContinue; "
+            "if ($p.Class -eq 'Net') { $_.PSChildName } }); "
```
Dany9966: Won't this simply print out the child GUID? Do we need this here?

coriolis/osmorphing/windows.py
```diff
@@ -599,12 +612,46 @@ def _install_cloudbase_init(self, download_url,
...
+            "Get-ChildItem -Path \"$classBase\\$netClassGuid\" "
+            "-ErrorAction SilentlyContinue | "
+            "ForEach-Object { $p = Get-ItemProperty -Path $_.PSPath "
+            "-ErrorAction SilentlyContinue; "
+            "if ($p.NetCfgInstanceId -and $p.DriverDesc) "
+            "{ \"$($p.NetCfgInstanceId)|$($p.DriverDesc)|"
+            "$($p.NetworkAddress)\" } }; "
+            "exit 0" % {'base': class_base_path})
+        result = self._conn.exec_ps_command(ps_command)
```
Dany9966, line 628: Can you please log this result output?

coriolis/osmorphing/windows.py
Dany9966: Also, this apperently is the wrong path.
Apologies for that, I've mistaken the Interface name with the interface description. We're looking for interface name actually, that's the thing that needs to be kept.
You will find it in: 
`HKLM:\SYSTEM\ControlSet001\Control\Network\{SOME_RANDOM_WINDOWS_GUID}\{NetGUID}\Connection`
Key `Name`

Dany9966: We'll look into finding the MAC address somewhere else as well, it seems like that's not the relevant location either.

Dany9966 (changes requested):

Dany9966:
We took a different approach in-provider for this. We can close this now.

PR #417, opened 2026-04-28
integration: Adds Docker container usage for minions

claudiubelu:
Adds docker-related utils. Updates the test provider to create docker container minions instead of using `127.0.0.1` as a minion.

Adds data-minion Dockerfile. The container image will have `openssh-server` (Coriolis will SSH into it) and `systemd` installed in it
(Coriolis will set up `systemd` units for the replicator and the writer)

The integration tests now require a container image for minions to run.

### DIFF
--- a/coriolis/tests/integration/base.py
+++ b/coriolis/tests/integration/base.py
@@ -13,6 +13,7 @@ Subclasses must be run as root.
 """
 
 import os
+import subprocess
 import time
 import unittest
 from unittest import mock
@@ -32,11 +33,6 @@ from coriolis.tests import test_base
 
 CONF = cfg.CONF
 LOG = logging.getLogger(__name__)
-
-# Path to the SSH private key used to connect to the (local) provider.
-# Override via the CORIOLIS_TEST_SSH_KEY_PATH environment variable.
-_TEST_SSH_KEY_PATH = os.environ.get(
-    'CORIOLIS_TEST_SSH_KEY_PATH', '/root/.ssh/id_rsa')
 
 
 class CoriolisIntegrationTestBase(test_base.CoriolisBaseTestCase):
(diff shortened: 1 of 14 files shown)


coriolis/tests/integration/base.py
```diff
@@ -109,6 +110,29 @@ def f(*args, **kwargs):
 
 
 class ReplicaIntegrationTestBase(CoriolisIntegrationTestBase):
+
+    @classmethod
+    def setUpClass(cls):
+        if not os.path.exists(_TEST_SSH_KEY_PATH):
+            raise unittest.SkipTest(
```
petrutlucian94: Can we generate a set of keys as part of the test?

coriolis/tests/integration/base.py
```diff
@@ -154,6 +178,7 @@ def setUp(self):
         repl = "coriolis.providers.replicator.Replicator"
         for prop in [
             "coriolis.providers.backup_writers._disable_lvm2_lvmetad",
+            "coriolis.providers.backup_writers._disable_lvm_metad_udev_rule",
```
petrutlucian94: Why is this mock necessary?

petrutlucian94 (approved):
Looks good overall.

coriolis/tests/integration/base.py
claudiubelu: Done

coriolis/tests/integration/base.py
claudiubelu: It's not, but there are a few SSH commands being executed, fail (and generate visible noise in the logs), and we continue executing anyways.

Anyways, removed.

petrutlucian94: It may be worth revisiting the minion initialization.

PR #418, opened 2026-04-28
Remove cloud_etc_hosts module when DHCP is disabled

Cristi1324:


coriolis/osmorphing/base.py
```diff
@@ -429,6 +429,20 @@ def _ensure_cloud_init_not_disabled(self):
     def _reset_cloud_init_run(self):
         self._exec_cmd_chroot("cloud-init clean --logs")
 
+    def _disable_cloud_manage_etc_hosts(self):
```
Dany9966: I think you should also add the list of cloud modules without the `update_etc_hosts` into cloud_cfg_mods instead of directly modifying the main file.

Dany9966 (changes requested):

Dany9966 (approved):

PR #420, opened 2026-04-29
Adds exception subclasses for checksum verification

claudiubelu:


petrutlucian94 (approved):
I guess this will be used by the integration tests. Looks good.

PR #421, opened 2026-04-29
Fix a flaky issue with task percentage not stored in DB

fabi200123:
It seems like there is a flaky issue with the percentage for Replicate Disks.
The final steps for replication are not stored in the DB and the UI doesn't display the 100% for disk replication. The worker reaches the 100%, but the 100% is never stored in the DB.

This issue seems to come from the `self._cast` from `update_task_progress_update`.

NOTE: Changing it to `_call` seems to fix the issue, but the whole issue is hard to replicate.

### DIFF
--- a/coriolis/conductor/rpc/client.py
+++ b/coriolis/conductor/rpc/client.py
@@ -293,8 +293,17 @@ class ConductorClient(rpc.BaseRPCClient):
 
     def update_task_progress_update(
             self, ctxt, task_id, progress_update_index, new_current_step,
-            new_total_steps=None, new_message=None):
-        self._cast(
+            new_total_steps=None, new_message=None, sync=False):
+        # Use _call on the last step to make sure the DB is updated
+        # and the UI is showing the correct progress.
+        # Intermediate steps use cast to avoid blocking on every tick.
+        if sync:
+            return self._call(
+                ctxt, 'update_task_progress_update', task_id=task_id,
+                progress_update_index=progress_update_index,
+                new_current_step=new_current_step,
+                new_total_steps=new_total_steps, new_message=new_message)
+        return self._cast(
             ctxt, 'update_task_progress_update', task_id=task_id,
             progress_update_index=progress_update_index,
             new_current_step=new_current_step,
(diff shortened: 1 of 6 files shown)


coriolis/conductor/rpc/client.py
```diff
@@ -294,7 +294,10 @@ def add_task_progress_update(
     def update_task_progress_update(
             self, ctxt, task_id, progress_update_index, new_current_step,
             new_total_steps=None, new_message=None):
-        self._cast(
+        # Must use synchronous RPC: EventManager advances last_perc only after
```
petrutlucian94: Won't a synchronous RPC call slow down the transfer? There are quite a few progress updates that will now have to wait for the conductor.

fabi200123: That was the Daniel's question as well. 
Asked Emilian to do a test, and there was no significant difference regarding time.

petrutlucian94: "no significant difference" doesn't tell us much. When talking about performance, we need actual numbers. Ideally some profiler data.

The performance impact may be more significant if the conductor is under load, e.g. having 10-20 concurrent migrations.

coriolis/db/api.py
```diff
@@ -990,6 +925,17 @@ def update_task_progress_update(
             "Could not find progress update for task with ID '%s' and "
             "index %s in the DB for updating." % (task_id, update_index))
 
+    # The _call on the last task update might get overwritten by an
+    # out-of-order update with an older current_step value, so we need to
+    # ensure that the progress is not overwritten by an older step.
+    if new_current_step < task_progress_update.current_step:
```
petrutlucian94, line 935: This is better than what we have now but it's still prone to race conditions. We rely on `task_progress_update.current_step`, but it may be incremented by a concurrent DB operation and we'd end up overwriting it, potentially decreasing `current_step`.

The safest approach would be to use a "compare-and-update" sql operation.

I'm ok with merging this PR as it is and add a TODO to use a db-level check.

petrutlucian94: Suggested comment:

```
# Quick progress updates may be processed out of order. We're trying to mitigate
# this by checking the current step.
#
# TODO: the current approach is still prone to race conditions since
# `task_progress_update.current_step` may be out of date. The safest approach
# would be to use a db-level `compare-and-update` sql operation.
```

PR #423, opened 2026-05-05
Add API pagination

petrutlucian94:
At the moment, listing endpoint instances is the only Coriolis API that supports pagination. For this reason, Coriolis clients often retrieve much more database records than needed, leading to poor performance.

We're now adding pagination to other Coriolis APIs:
* transfers
* transfer executions
* deployments

New optional parameters:
* ``sort_key`` - sort key, repeatable. `created_at` and `id` are used by default.
* ``sort_dir`` - sort direction, repeatable. `asc` or `desc` (default).
* ``marker`` - the last seen ID, omitted from the results.
* ``limit`` - the maximum number of records to retrieve.

The pagination will be performed on the db API side, leveraging the "utils.paginate_query" helper from oslo_db.

Example:

```
GET http://server:7667/v1/transfers?marker=a7061715-e56c-470c-a6ac-80bb02f1f198&limit=2&sort_key=id&sort_dir=asc
```

Note that the integration tests depend on the client changes: https://github.com/cloudbase/python-coriolisclient/pull/98

### DIFF
--- a/coriolis/api/common.py
+++ b/coriolis/api/common.py
@@ -10,3 +10,36 @@ def get_paging_params(req):
     if limit is not None:
         limit = utils.parse_int_value(limit)
     return marker, limit
+
+
+def get_sort_params(req,
+                    default_key='created_at',
+                    default_dir='desc'):
+    """Retrieves sort keys/directions parameters.
+
+    Processes the parameters to create a list of sort keys and sort directions
+    that correspond to the 'sort_key' and 'sort_dir' parameter values. These
+    sorting parameters can be specified multiple times in order to generate
+    the list of sort keys and directions.
+
+    The input parameters are not modified.
+
+    :param req: coriolis.api.wsgi.Request object
+    :param default_key: default sort key value, added to the list if no
+                        'sort_key' parameters are supplied
+    :param default_dir: default sort dir value, added to the list if no
+                        'sort_dir' parameters are supplied
+    :returns: list of sort keys, list of sort dirs
+    """
+    params = req.params.copy()
+    sort_keys = []
+    sort_dirs = []
+    while 'sort_key' in params:
+        sort_keys.append(params.pop('sort_key').strip())
+    while 'sort_dir' in params:
+        sort_dirs.append(params.pop('sort_dir').strip())
+    if len(sort_keys) == 0 and default_key:
+        sort_keys.append(default_key)
+    if len(sort_dirs) == 0 and default_dir:
+        sort_dirs.append(default_dir)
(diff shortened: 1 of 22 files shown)


coriolis/api/common.py
```diff
@@ -10,3 +10,36 @@ def get_paging_params(req):
     if limit is not None:
         limit = utils.parse_int_value(limit)
     return marker, limit
+
+
+def get_sort_params(req,
```
petrutlucian94, line 15: Shared by most Openstack services.

coriolis/db/api.py
```diff
@@ -1504,3 +1569,78 @@ def update_minion_pool(context, minion_pool_id, updated_values):
     # the oslo_db library uses this method for both the `created_at` and
     # `updated_at` fields
     setattr(lifecycle, 'updated_at', timeutils.utcnow())
+
+
+def process_sort_params(
```
petrutlucian94, line 1574: Shared by most Openstack services.

coriolis/tests/integration/test_pagination.py
```diff
@@ -0,0 +1,307 @@
...
+        deployment = models.Deployment(
+            **kwargs)
+        db_api.add_deployment(cls._ctx, deployment)
+        cls._cleanup_callbacks.append(
+            lambda: db_api.delete_deployment(cls._admin_ctx, deployment.id))
+        return deployment
+
+    @classmethod
+    def _setup_mocks(cls):
+        # Can't use "addCleanup" for resources created by "setUpClass".
```
claudiubelu: Indeed, but there is `addClassCleanup` (https://docs.python.org/3/library/unittest.html#unittest.TestCase.addClassCleanup).

coriolis/tests/integration/test_pagination.py
```diff
@@ -0,0 +1,307 @@
...
+        # despite not being declared by the model.
+        execution.action = transfer
+        db_api.add_transfer_tasks_execution(cls._admin_ctx, execution)
+        cls._cleanup_callbacks.append(
+            lambda: db_api.delete_transfer_tasks_execution(
+                cls._admin_ctx, execution.id))
+        return execution
+
+    @classmethod
+    def _create_endpoint(
```
claudiubelu: nit: could we name these `_create_*` methods as `_create_db_` instead? We already have `_create_endpoint` and  `_create_transfer` in `CoriolisIntegrationTestBase`, so it can get a bit confusing.

claudiubelu (approved):

PR #426, opened 2026-05-11
integration: Adds Minion pools tests

claudiubelu:
Adds `BaseDestinationMinionPoolProvider` to `TestImportProvider` and implements all abstract methods. The test provider now advertises the `PROVIDER_TYPE_DESTINATION_MINION_POOL` capability.

`TestImportProvider.create_minion()` starts a `coriolis-data-minion` container, waits for SSH, deploys the HTTP backup writer, and returns real connection / writer info.

`start_minion` / `shutdown_minion` / `delete_minion map` to docker start / stop / remove.

`attach_volumes_to_minion` hotplugs block devices into the container's mount
namespace via `nsenter` + `mknod`.

healthcheck_minion opens an SSH connection to verify liveness.

Adds `_InProcessTaskflowRunner` that runs pool task flows in daemon threads  instead of child processes, keeping them inside the shared `fake://`  transport.

Add `_InProcessMinionManagerServerEndpoint` that uses the in-thread runner.

Start the `MinionManagerServerEndpoint` in the harness.

Patch `coriolis.keystone.delete_trust` globally (no real Keystone in tests).

Disable the automatic pool-refresh cron (period=0) to prevent Keystone calls during pool lifecycle tests.

Adds integration tests for minion pool lifecycle. Adds test for allocating / deallocating minion pools.

 Adds transfer, transfer execution, and  deployment tests using minion  pools.

### DIFF
--- a/coriolis/tests/integration/base.py
+++ b/coriolis/tests/integration/base.py
@@ -22,17 +22,32 @@ from coriolisclient import client as coriolis_client
 from keystoneauth1 import session as ks_session
 from keystoneauth1 import token_endpoint
 from oslo_config import cfg
+from oslo_db.sqlalchemy import models
 from oslo_log import log as logging
+import oslo_messaging as messaging
 
 from coriolis import constants
 from coriolis import context
 from coriolis.db import api as db_api
+from coriolis.providers import factory as providers_factory
 from coriolis.tests.integration import harness
 from coriolis.tests.integration import utils as test_utils
 from coriolis.tests import test_base
 
 CONF = cfg.CONF
 LOG = logging.getLogger(__name__)
+
+# Statuses that represent a completed allocation attempt.
+MINION_ALLOCATED_TERMINAL = {
+    constants.MINION_POOL_STATUS_ALLOCATED,
+    constants.MINION_POOL_STATUS_ERROR,
+}
+
+# Statuses that represent a completed deallocation attempt.
+MINION_DEALLOCATED_TERMINAL = {
+    constants.MINION_POOL_STATUS_DEALLOCATED,
+    constants.MINION_POOL_STATUS_ERROR,
+}
 
 
 class CoriolisIntegrationTestBase(test_base.CoriolisBaseTestCase):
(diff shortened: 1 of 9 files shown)


coriolis/tests/integration/base.py
```diff
@@ -55,7 +70,36 @@ def setUpClass(cls):
     def setUp(self):
         super().setUp()
 
-        patcher = mock.patch("psutil.Process.send_signal")
+        to_patch = [
```
petrutlucian94, line 73: Ouch, can't wait to get rid of those mocks once we switch to external services.

coriolis/tests/integration/base.py
```diff
@@ -96,11 +140,78 @@ def _create_transfer(self, src_id, dst_id, instances):
...
+                self._client.minion_pools.deallocate_minion_pool(
+                    pool_id, force=True)
+                self._wait_for_pool(pool_id, MINION_DEALLOCATED_TERMINAL)
+            except Exception:
+                pass
+
+        try:
+            self._client.minion_pools.delete(pool_id)
+        except Exception:
+            pass
```
petrutlucian94: Why are we ignoring pool deletion failures? At the very least I'd log the suppressed exceptions.

coriolis/tests/integration/utils.py
```diff
@@ -111,12 +111,19 @@ def write_test_pattern(device_path, chunk_size=4096):
 
     Returns the pattern bytes so callers can verify the destination later.
     The write is done with ``dd`` so it works on raw block devices.
+
+    Two chunks are written (offsets 0 and chunk_size) so that any residual
```
petrutlucian94: A bit ugly, the caller may simply request a larger chunk size. At the moment the helper makes assumptions about the test logic.

coriolis/tests/integration/base.py
claudiubelu: Fair enough. Not catching those exceptions anymore. If it fails, we'll see it in the fact that the test failed.

coriolis/tests/integration/utils.py
claudiubelu: Done. Reverted, in `base.py` we now call this with `chunk_size=8192`.

petrutlucian94 (approved):

PR #427, opened 2026-05-11
Ubuntu OSMorphing Refactor

Dany9966:
This PR contains the following patches:

1. Fix new cloud-init version service enablement.
    
    New cloud-init version (>26) changed its systemd unit service name,
    therefore the enablement method needs to reflect that.

2. Improve upon Ubuntu OSMorphing tool check
    
    This patch drops OSMorphing support for any non-LTS Ubuntu release.
    This will allow for provider-side refactoring and tool grouping.

3. Replace all deprecated `@abc.abstractclassmethod` declarations
    
    Replaced them with stacked `@classmethod\n@abc.abstractmethod`

### DIFF
--- a/coriolis/osmorphing/ubuntu.py
+++ b/coriolis/osmorphing/ubuntu.py
@@ -24,15 +24,16 @@ class BaseUbuntuMorphingTools(debian.BaseDebianMorphingTools):
                 UBUNTU_DISTRO_IDENTIFIER):
             return False
 
-        lts_releases = [12.04, 14.04, 16.04, 18.04, 20.04]
-        for lts_release in lts_releases:
-            if cls._version_supported_util(
-                    detected_os_info['release_version'],
-                    minimum=lts_release, maximum=lts_release):
-                return True
+        version, subversion = detected_os_info['release_version'].split('.', 1)
+        if int(version) % 2 != 0 or not subversion.startswith("04"):
+            LOG.warning(
+                "Detected Ubuntu release version "
+                f"'{detected_os_info['release_version']}' is not an LTS one. "
+                "Coriolis only supports morphing Ubuntu LTS versions.")
+            return False
 
         return cls._version_supported_util(
-            detected_os_info['release_version'], minimum=22.04)
+            detected_os_info['release_version'], minimum=12.04)
 
     def _set_netplan_ethernet_configs(
             self, nics_info, dhcp=False, iface_name_prefix=None):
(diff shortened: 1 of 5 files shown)


coriolis/osmorphing/ubuntu.py
```diff
@@ -24,15 +24,16 @@ def check_os_supported(cls, detected_os_info):
...
             return False
 
-        lts_releases = [12.04, 14.04, 16.04, 18.04, 20.04]
-        for lts_release in lts_releases:
-            if cls._version_supported_util(
-                    detected_os_info['release_version'],
-                    minimum=lts_release, maximum=lts_release):
-                return True
+        _, subversion = detected_os_info['release_version'].split('.', 1)
+        if not subversion.startswith("04"):
```
petrutlucian94: Not all `*.04` releases are LTS, for example `25.04` is not a LTS release. You may want to check if the major version is an even number.

coriolis/osmorphing/ubuntu.py
```diff
@@ -24,15 +24,16 @@ def check_os_supported(cls, detected_os_info):
...
+            LOG.warning(
+                "Detected Ubuntu release version "
+                f"'{detected_os_info['release_version']}' is not an LTS one. "
+                "Coriolis only supports morphing Ubuntu LTS versions.")
```
petrutlucian94, line 32: So we're completely dropping support for non-LTS Ubuntu releases (at the moment it accepts non-LTS versions newer than 22.04).

That's reasonable, however let's clearly state it in the PR message.

```
This patch makes sure that the base Ubuntu osmorphing tool checks
for the migrated machine's subversion, and makes sure that it's an
LTS subversion, instead of checking each LTS version separately.
```

As it's stated now, it implies that we're rewriting the version check, not that we're completely dropping support for non-LTS releases. Simply saying `This patch drops os-morphing support for non-LTS Ubuntu releases.` would be much clearer.

coriolis/osmorphing/ubuntu.py
Dany9966: Very good catch, thank you!

coriolis/osmorphing/ubuntu.py
Dany9966: We never officially supported non-LTS versions, see the old code, it literally just went through the list of LTS versions (at the time), and otherwise it refused to use any kind of other ubuntu release (if you consider the provider code as well). 

Anyway, I'll clarify the PR and commit messages. Thank you!

petrutlucian94: Interesting. The current code checks if it's a LTS version. If it isn't, then it checks if it's newer than 22.04. Unless I'm mistaken, 22.10 would be reported as "supported.

This made me believe that we had the policy of supporting recent non-LTS releases.

Dany9966: At least at the base level maybe it was an attempt at future-proofing, but the providers had specific supported versions, and they all had tools for LTS versions only.
We're trying to future-proof both the core and provider tools with these additions.

petrutlucian94 (approved):

PR #428, opened 2026-05-12
integration: Fixes the test_scheduled_execution test

claudiubelu:
If the the cron job checker ran for *this* minute, and we create a scheduled transfer for *this* minute, it will be ignored / skipped.

In some cases, we were creating the scheduled transfer in *this* minute, causing the test to fail randomly.


petrutlucian94 (approved):
Reasonable workaround.

PR #430, opened 2026-05-13
integration: Adds OS morphing deployment integration test

claudiubelu:
Replace the stub `deploy_os_morphing_resources` / `delete_os_morphing_resources` with a real implementation that starts a container with `openssh-server`, injects the test public key via bind mount, and returns SSH connection info for the OS morphing pipeline.

Adds OS morphing test, in which we prepare a source disk with an Ubuntu filesystem, based on the Ubuntu container image. The test expects that a package (`jq`) will be present after the OS morphing process.


coriolis/tests/integration/deployments/test_osmorphing.py
```diff
@@ -0,0 +1,42 @@
...
+
+class OsMorphingDeploymentTest(integration_base.ReplicaIntegrationTestBase):
+
+    # NOTE(claudiub): Size must be high enough to contain the tested OS and
+    # any new packages to be added during OS morphing.
+    _SCSI_DEBUG_SIZE_MB = 256
+
+    def setUp(self):
+        super().setUp()
+        test_utils.make_os_device(self._src_device, "ubuntu:24.04")
```
petrutlucian94: nit: I don't particularly like the naming, "os_device" doesn't say much. I'd call it something like  "write_os_image_to_disk".

coriolis/tests/integration/base.py
```diff
@@ -280,6 +281,10 @@ def setUpClass(cls):
                     "Pool did not reach ALLOCATED (got %s)" % pool_obj.status,
                 )
 
+        # (re)init the scsi_debug module.
+        test_utils.destroy_scsi_debug()
```
petrutlucian94, line 285: Are we planning to switch to loopback devices btw? I think we'll need to in order to parallelize the tests.

coriolis/tests/integration/deployments/test_osmorphing.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def test_deployment_with_os_morphing(self):
+        self.assertFalse(
```
petrutlucian94, line 25: This feels a bit brittle. So if `jq` ever gets included in the Ubuntu image, this test will start to fail.

On the other hand, we can keep it for the sake of simplicity. We'll use a different package if that happens :).

petrutlucian94 (commented):
Looks good overall.

coriolis/tests/integration/deployments/test_osmorphing.py
claudiubelu: Done

coriolis/tests/integration/deployments/test_osmorphing.py
claudiubelu: Agreed. In this example, we're basically using a container image, which is supposed to be a minimal base image to be further used for for other purposes; it shouldn't have a situational package as `jq`.

I've added this assertion as a sanity check, so we don't have a potentially false positive test (expect package to be installed by OS morphing, but it already existed anyways).

I'd keep `jq` for now, as it's very small. We can change it if needed later.

coriolis/tests/integration/base.py
claudiubelu: I'm afraid we'll be using those for now. loopback devices are fine on the importer (`coriolis-writer` doesn't care what type of device it writes to), but on the exporter, `coriolis-replicator` ignores loopback devices.

petrutlucian94 (approved):

PR #431, opened 2026-05-15
Switch from using flake8 to using flake8 + ruff

fabi200123:
This PR contains 2 commits:
- First one adds ruff + flake8 instead of just flake8 for verifying the coding style
- Second commit contains the formatted code from `tox -e fmt`, plus the fix of changing  of the following error:
```
coriolis git:(switch-to-ruff) tox -e fmt                                     
fmt: commands[0]> ruff check --fix /Users/fabian/Documents/CBSL/coriolis
E741 Ambiguous variable name: `l`
   --> coriolis/osmorphing/base.py:341:26
    |
340 |         add = [p[0] for p in itertools.chain.from_iterable(
341 |                [l for k, l in self._packages.items() if k in k_add])]
    |                          ^
342 |
343 |         # If the second value in the tuple is True, the package must
    |

E741 Ambiguous variable name: `l`
   --> coriolis/osmorphing/base.py:346:29
    |
344 |         # be uninstalled on export
345 |         remove = [p[0] for p in itertools.chain.from_iterable(
346 |                   [l for k, l in self._packages.items()])
    |                             ^
347 |                   if p[1]]
    |

Found 234 errors (232 fixed, 2 remaining).
fmt: exit 1 (0.45 seconds) /Users/fabian/Documents/CBSL/coriolis> ruff check --fix /Users/fabian/Documents/CBSL/coriolis pid=45407
  fmt: FAIL code 1 (0.55=setup[0.10]+cmd[0.45] seconds)
  evaluation failed :( (0.68 seconds)
  ```
  Changing it to `pkgs for k, pkgs in self._packages.items() if k in k_add`

### CODE REVIEW

File: tox.ini [hunk 1]
```diff
@@ -48,6 +68,9 @@ omit = coriolis/tests/*
 [flake8]
 # E125 is deliberately excluded. See https://github.com/jcrocholl/pep8/issues/126
 # E251 Skipped due to https://github.com/jcrocholl/pep8/issues/301
+#
+# E501 (line too long) is left to ruff
+# E203 (whitespace before ':') conflicts with how ruff formats slices.
```
claudiubelu, line 80: What about H301, H306?

ruff.toml
```diff
@@ -0,0 +1,26 @@
...
+    "tools",
+    "*openstack/common*",
+    "*lib/python*",
+    "*egg",
+]
+
+[lint]
+select = ["E", "W", "F"]
+
+ignore = [
```
claudiubelu, line 15: Should probably add the explanations for these ignores too.

tox.ini
```diff
@@ -48,6 +68,9 @@ omit = coriolis/tests/*
 [flake8]
 # E125 is deliberately excluded. See https://github.com/jcrocholl/pep8/issues/126
 # E251 Skipped due to https://github.com/jcrocholl/pep8/issues/301
+#
+# E501 (line too long) is left to ruff
```
claudiubelu: E501 is left to ruff, but E501 is ignored in ruff, as seen in `ruff.toml`.

fabi200123: I updated the comment, the idea was that it was left to the `line-length = 88` option in ruff.toml

tox.ini
fabi200123: Added comments for those as well

ruff.toml
fabi200123: Done!

PR #434, opened 2026-05-19
cleanup: Removes redundant retry decorators from exec_ssh_cmd callers

claudiubelu:
`exec_ssh_cmd` already retries up to 5 times. Functions that also have the `@retry_on_error()` decorator and only call `exec_ssh_cmd` (or other already-retried helpers) are compounding retries to 25x or even more.

Remove the decorator from:
- `utils.py`: `create_service`, `restart_service`, `start_service`, `stop_service` (all inner calls go through `test_ssh_path` / `exec_ssh_cmd`, both retried)
- `replicator.py`: `_copy_replicator_cmd`, `_exec_replicator`, `_setup_replicator` (`_setup_replicator` is also wrapped externally by `init_replicator`, making its own decorator triply redundant)

Updates tests that used `get_wrapped_function` to bypass the (now removed) decorator. They now call the methods directly.

### DIFF
--- a/coriolis/providers/replicator.py
+++ b/coriolis/providers/replicator.py
@@ -570,7 +570,6 @@ class Replicator(object):
             ssh, "sudo mv %s %s" % (tmp, remotePath), get_pty=True)
         sftp.close()
 
-    @utils.retry_on_error()
     def _copy_replicator_cmd(self, ssh):
         local_path = os.path.join(
             utils.get_resources_bin_dir(), 'replicator')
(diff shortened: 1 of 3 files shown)


petrutlucian94:
https://github.com/cloudbase/coriolis/pull/433 went a different route, it modifies `exec_ssh_cmd` itself to only retry ssh errors.

In my opinion `exec_ssh_cmd` shouldn't retry arbitrary commands because it can't distinguish expected exceptions, which shouldn't be retried.

Dany9966 (approved):

PR #435, opened 2026-05-19
Extend user scripts API, allowing the phase to be specified

petrutlucian94:
When deploying replicas, Coriolis users can specify scripts that will be executed during os-morphing, right after the OS partition is mounted on the minion instance.

In some situations, this is too late since user scripts may be needed in order to be able to mount the OS disk, for example if it’s encrypted.

In other situations it’s too early. Some scripts may need to be executed on the replica instance. This may require provider assistance.

To accommodate these use cases, we’ll extend the deployment API, allowing the user to specify when a given script should be executed. Each script may specify one of the following execution phases:

* osmorphing-pre-os-mount
* osmorphing-post-os-mount (default)
* replica-initial-boot (TBD)

We'll continue to support the basic script format as well. See the API samples for a few examples.

Note that we can't use the OS morphing tools to run pre-os-mount scripts. A few approaches have been considered:

* initialize the OS morphing tools against the minion root partition
    * hacky
    * unnecessarily complicated and brittle
* use the base Linux/Windows OS morphing classes
    * can't be instantiated since not all abstract methods are implemented by
      the base classes
* extend the OS mount tools to allow running pre-os-mount user scripts
    * easiest and most natural approach, implemented by this patch

The PR also fixes Windows user script execution, ensuring that errors get propagated.

### DIFF
--- a/coriolis/api-refs/api_samples/deployment/deployment-from-transfer-req.json
+++ b/coriolis/api-refs/api_samples/deployment/deployment-from-transfer-req.json
@@ -6,6 +6,20 @@
     "skip_os_morphing": false,
     "instance_osmorphing_minion_pool_mappings": {
       "instance1": "morphing_pool"
+    },
+    "user_scripts": {
+      "global": {
+        "linux": [
+          {
+              "phase": "osmorphing_pre_os_mount",
+              "payload": "echo 'unlocking encrypted OS partion'"
+          },
+          {
+              "phase": "osmorphing_post_os_mount",
+              "payload": "echo 'modifying replica OS filesystem'"
+          }
+        ]
+      }
     }
   }
 }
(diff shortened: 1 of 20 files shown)


petrutlucian94:
Rebased on top of https://github.com/cloudbase/coriolis/pull/433, we should merge that one first.

coriolis/tests/test_utils.py
```diff
@@ -83,6 +83,26 @@ def test_retry_on_error_exception_keyboard_interrupt(self):
 
         self.assertEqual(self.mock_func.call_count, 1)
 
+    def test_retry_on_error_not_in_list_of_retried(self):
```
Dany9966, line 86: Is this a left-over from the other PR maybe? Can you please rebase?

Dany9966 (approved):

PR #436, opened 2026-05-22
Add LUKS disk encryption support for Linux OS morphing

claudiubelu:
The bulk of the work lives in the new `LinuxLuksMixin` class
(`osmount/luks_mixin.py`), which is then included in `BaseLinuxOSMountTools`:

- `mount_os()`: check `osmorphing_info["disk_luks_passphrases"]` for each block device. confirmed LUKS containers are opened via `cryptsetup luksOpen` and the resulting `/dev/mapper/<name>` path is used in place of the raw device. `dismount_os()` closes them again after all filesystems have been unmounted.

- `remove_encryption_artifacts`: after OS morphing, stale TPM2 LUKS tokens and their keyslots are killed and the corresponding `tpm2-*` options are stripped from `/etc/crypttab`. The source TPM does not exist on the destination, so leaving these in place would cause the initramfs to hang or fail on first boot.

- `install_encryption_firstboot_setup`: a temporary migration keyfile is injected into the guest, `/etc/crypttab` is updated to reference it, the initramfs is rebuilt so the migrated VM can boot, GRUB is patched to use the crypttab mapper names instead of the osmount-time names, and a systemd one-shot service is installed to re-enroll TPM2 and remove the migration keyfile on the first boot of the destination VM.

The firstboot shell scripts live in `coriolis/osmorphing/osmount/resources`:

- `luks_firstboot_initramfs_tools.sh` and targets `update-initramfs`-based systems (Debian / Ubuntu).
- `luks_firstboot_dracut.sh`: targets dracut-based systems.

Note that the scripts include a reboot; a migration with LUKS-encrypted disks is not considered completed until a reboot is performed and the disks get unlocked and get mounted properly.


coriolis/osmorphing/osmount/resources/luks_firstboot_dracut.sh
```diff
@@ -0,0 +1,163 @@
+#!/bin/bash
+
+# Coriolis LUKS firstboot cleanup for dracut (RHEL / Fedora / SUSE).
+# Runs once on first boot to re-enroll TPM2, remove migration keyslots, and
+# rebuild the initramfs so the embedded keyfile is gone.
+
+set -e
```
petrutlucian94, line 7: Would it make sense to use `set -x`, making this script easier to debug?

coriolis/osmorphing/osmount/luks_mixin.py
```diff
@@ -0,0 +1,686 @@
...
+                "Could not determine LUKS UUIDs for any passphrase device; "
+                "migration keyfiles not written.")
+
+        self._update_crypttab_keyfile(os_root_dir, uuid_to_keyfile)
+        self._configure_dracut_keyfiles(os_root_dir, uuid_to_keyfile)
+        self._configure_initramfs_tools_keyfiles(os_root_dir)
+
+    def _configure_dracut_keyfiles(self, os_root_dir, uuid_to_keyfile):
+        """Write a dracut.conf.d snippet to embed keyfiles in the initramfs."""
+        for dracut_bin in ["usr/bin/dracut", "usr/sbin/dracut", "sbin/dracut"]:
```
petrutlucian94: I'd just update the caller to use `_detect_initramfs_tool` and remove this check. The caller would then use either `_configure_dracut_keyfiles` or `_configure_initramfs_tools_keyfiles`.

coriolis/osmorphing/osmount/luks_mixin.py
```diff
@@ -0,0 +1,686 @@
...
+        if utils.test_ssh_path(self._ssh, path):
+            return "openrc"
+
+        path = os.path.join(os_root_dir, "sbin/initctl")
+        if utils.test_ssh_path(self._ssh, path):
+            return "upstart"
+
+        return "sysvinit"
+
+    def _register_firstboot_script_systemd(self, os_root_dir):
```
petrutlucian94: We'll also need this for user provided "firstboot" scripts. I guess we'll end up with a separate public method as part of `BaseOSMorphingTools`.

coriolis/tests/integration/deployments/test_osmorphing.py
```diff
@@ -37,6 +63,94 @@ def test_deployment_with_os_morphing(self):
...
+            self._key_file = fh.name
+            fh.write(_LUKS_PASSPHRASE)
+
+        super().setUp()
+
+    def _prepare_src_device(self):
+        disk_id = os.path.basename(self._src_device)
+
+        # Write a minimal Linux OS on the device, encrypted with LUKS.
+        test_utils.make_luks_device(
```
petrutlucian94: Later on, it might be useful to test the scenario in which there are multiple disk attachments. We'd have to ensure that Coriolis doesn't error out if the user didn't specify the keys for secondary encrypted disks.

coriolis/osmorphing/osmount/luks_mixin.py
```diff
@@ -0,0 +1,686 @@
...
+        super().__init__(*args, **kwargs)
+        self._luks_opened = []
+
+    def _unlock_luks_devices(self, dev_paths):
+        """Open any LUKS-encrypted devices listed in dev_paths in-place.
+
+        Reads osmorphing_info["disk_luks_passphrases"], containing mappings
+        {"device_path": "passphrase"}.
+
+        For each device that has a passphrase entry and is confirmed LUKS by
```
petrutlucian94: So the user will have to specify the luks passphrase for each disk path, right?

Do we expect those to be udev links based on the disk WWN or serial ID? For example:

```
/dev/disk/by-uuid/4306c488-13c8-447a-825d-3e9216876081
/dev/disk/by-id/wwn-0x6d094660793802002afcbbe61cfbcd38
/dev/disk/by-id/scsi-36d094660793802002afcbbe61cfbcd38
```

Do we need `coriolis-web` changes to accommodate this?

coriolis/osmorphing/osmount/resources/luks_firstboot_dracut.sh
claudiubelu: Done

coriolis/tests/integration/deployments/test_osmorphing.py
claudiubelu: Indeed, but that is going to be more relevant in more complex tests, in which we actually spin up a replica and assert that the first boot scripts also worked as intended, and then we can also assert that additional disks are also accessible. Not much we can do here, in these tests specifically.

coriolis/osmorphing/osmount/luks_mixin.py
petrutlucian94: As discussed during the call, the user can't know these paths beforehand since the original WWN / SCSI ID won't be preserved.

The consensus was that we're going to use the same key for all LUKS encrypted disks.

coriolis/osmorphing/osmount/luks_mixin.py
```diff
@@ -0,0 +1,686 @@
...
+
+
+_LUKS_FIRSTBOOT_SCRIPTS = {
+    "update-initramfs": _load_script("luks_firstboot_initramfs_tools.sh"),
+    "dracut": _load_script("luks_firstboot_dracut.sh"),
+}
+
+_SYSTEMD_UNIT = """\
+[Unit]
+Description=Coriolis LUKS migration firstboot cleanup
```
petrutlucian94: Won't it run at every boot, not just the first one?

petrutlucian94: Nevermind, the invoked scripts are supposed to disable the service.

coriolis/tests/integration/deployments/test_osmorphing.py
```diff
@@ -185,7 +185,7 @@ def setUp(self):
         super().setUp()
 
     def _prepare_src_device(self):
-        disk_id = os.path.basename(self._src_device)
+        os.path.basename(self._src_device)
```
petrutlucian94: Leftover? It's using `os.path.basename` without assigning it.

coriolis/tests/integration/deployments/test_osmorphing.py
```diff
@@ -232,14 +232,14 @@ class LUKSRockyLinuxOSMorphingDeploymentTest(LUKSOSMorphingDeploymentTest):
     """LUKS + dracut OS morphing test using Rocky Linux 9."""
 
     def _prepare_src_device(self):
-        disk_id = os.path.basename(self._src_device)
+        os.path.basename(self._src_device)
```
petrutlucian94: Same here.

coriolis/tests/integration/deployments/test_osmorphing.py
claudiubelu: Indeed, I missed this part.

coriolis/tests/integration/test_provider/imp.py
```diff
@@ -240,7 +240,13 @@ def restore_replica_disk_snapshots(
     def deploy_replica_instance(
             self, ctxt, connection_info, target_environment, instance_name,
             export_info, volumes_info, clone_disks):
-        return {"instance_deployment_info": {}}
+        instance_deployment_info = {}
+
+        luks_passphrase = target_environment.get("luks_passphrase")
+        if luks_passphrase:
+            instance_deployment_info["luks_passphrase"] = luks_passphrase
```
petrutlucian94: We discussed about renaming `luks_passphrase` to something that isn't LUKS specific since the backend might change. We may also want to reuse it for bitlocker, so maybe `encrypted_disk_passphrase`?

coriolis/osmorphing/osmount/base.py
```diff
@@ -58,6 +61,12 @@ def dismount_os(self, root_dir):
     def set_proxy(self, proxy_settings):
         pass
 
+    def remove_encryption_artifacts(self, os_root_dir):
+        pass
+
+    def install_encryption_firstboot_setup(self, os_root_dir):
```
petrutlucian94: Would it make sense to move this to the os morphing phase (outside of "os-mount")? We'd reuse `register_firstboot_script` introduced here: https://github.com/cloudbase/coriolis/pull/448/changes

coriolis/osmorphing/osmount/resources/luks_firstboot_dracut.sh
```diff
@@ -0,0 +1,167 @@
...
+        sed -i "s|$keyfile|none|g" /etc/crypttab
+        [ "$tpm2_enrolled" = "1" ] && add_tpm2_crypttab_option "$dev"
+    done
+}
+
+deregister_service() {
+    # Only disable, do NOT delete the unit file, or daemon-reload while the
+    # service is still running. systemd would detect "Current command vanished"
+    # and kill this process immediately.
+    systemctl disable coriolis-luks-firstboot.service 2>/dev/null || true
```
petrutlucian94: Possible leftover, we're no longer using `coriolis-luks-firstboot.service`.

claudiubelu: Done

coriolis/osmorphing/osmount/resources/luks_firstboot_dracut.sh
```diff
@@ -0,0 +1,158 @@
...
+
+remove_migration_keyslots
+
+echo "Deleting migration keyfiles."
+rm -f "${keyfiles[@]}"
+rmdir "$KEYFILE_DIR" 2>/dev/null || true
+
+echo "Rebuilding initramfs."
+# Embed the updated crypttab, so systemd-cryptsetup-generator uses the crypttab
+# mapper name and tpm2-device=auto for auto-unlock. The dracut.conf.d is deleted
```
petrutlucian94: > The dracut.conf.d is deleted

nit: this is a bit misleading, it implies that the entire `dracut.conf.d` dir gets deleted, however we're only deleting the Coriolis entry.

I'd rephrase it to "The Coriolis dracut.conf.d entry is deleted".

coriolis/osmorphing/osmount/resources/luks_firstboot_initramfs_tools.sh
```diff
@@ -0,0 +1,138 @@
...
+
+remove_migration_keyslots
+
+echo "Deleting migration keyfiles."
+rm -f "${keyfiles[@]}"
+rmdir "$KEYFILE_DIR" 2>/dev/null || true
+
+echo "Rebuilding initramfs."
+# Suppress needrestart: it reboots the VM after initramfs rebuild, which breaks
+# the firstboot SSH session before the service finishes.
```
petrutlucian94: nit: firstboot SSH session?

coriolis/osmorphing/osmount/base.py
```diff
@@ -422,16 +432,27 @@ def _get_volume_block_devices(self):
         #   where 'ln -s /dev/dm-N /dev/<VG-name>/<LV-name>'
         # Querying for the kernel device name (KNAME) should ensure we get the
         # device names we desire both for physical and logical volumes.
-        volume_devs = self._exec_cmd("lsblk -lnao KNAME").splitlines()
-        LOG.debug("All block devices: %s", str(volume_devs))
+        raw = self._exec_cmd("lsblk -lao KNAME,TYPE --json")
```
petrutlucian94: I'm a bit concerned about the "--json" flag, it seems like RHEL/centos/oracle 7 do not have it.

coriolis/osmorphing/osmount/luks_mixin.py
```diff
@@ -0,0 +1,608 @@
...
+                    (os.path.basename(mapper_path), dev_path))
+        except Exception:
+            self._close_luks_devices()
+            raise
+
+    def _unlock_luks_device(self, dev_path, passphrase):
+        """Open a single LUKS-encrypted device.
+
+        :raises: CoriolisException if the device/passphrase combination is
+          inconsistent (LUKS without passphrase, or passphrase for non-LUKS).
```
petrutlucian94: nit: we should update the docstring, we aren't really raising an exception if a password was received but there aren't any luks encrypted devices.

coriolis/osmorphing/osmount/resources/luks_firstboot_dracut.sh
claudiubelu: Done

coriolis/osmorphing/osmount/luks_mixin.py
```diff
@@ -0,0 +1,608 @@
...
+        :returns: the device mapper name, or None if the device is not LUKS
+          and no passphrase was given.
+        """
+        is_luks = self._is_luks(dev_path)
+
+        if is_luks and passphrase is None:
```
petrutlucian94: nit: I usually avoid `is None` checks to also account for empty strings.

coriolis/osmorphing/osmount/resources/luks_firstboot_initramfs_tools.sh
claudiubelu: Done

coriolis/osmorphing/osmount/luks_mixin.py
claudiubelu: Indeed, it's remnant from when we have multiple passphrases.

coriolis/osmorphing/osmount/luks_mixin.py
claudiubelu: Done

coriolis/osmorphing/osmount/base.py
claudiubelu: Done

petrutlucian94 (approved):
LGTM. We can merge it once the integration tests finish.

PR #438, opened 2026-05-22
osmorphing: dismount OS on failure

claudiubelu:
Currently, the mounted OS is only dismounted after the entire morphing process. If any exception occurs in between, the OS remains mounted.

This may be a problem for minion pools, in which the minions are supposed to remain active after an OS morphing job.


petrutlucian94 (approved):
LGTM, waiting for the integration tests to pass.

PR #440, opened 2026-05-25
integration: Various test fixes and cleanup

claudiubelu:
- Refactor Importer / Exporter class references. They are now referenced through the harness. This will allow them to be swapped in the future.
- Refactor Importer / Exporter connection info.
- Correctly skip Import provider that does not support minion pools. This needs to be done earlier, so the minion pool isn't created on a provider that do not support them.
- raise `AssertionError` instead of using `cls.fail`.
- On transfer execution, we now cancel the execution before deleting it.
- On deployments, we now cancel the deployment before deleting it.
- Updates integration tests' `README.md` file with up-to-date information.

### DIFF
--- a/coriolis/tests/integration/base.py
+++ b/coriolis/tests/integration/base.py
@@ -329,6 +329,31 @@ class ReplicaIntegrationTestBase(CoriolisIntegrationTestBase):
             transfer_id, shutdown_instances=False)
         self.assertExecutionCompleted(execution.id, timeout=timeout)
 
+    def _cleanup_execution(self, transfer_id, execution_id):
+        """Cancel a running execution if needed, then delete it.
+
+        Cancels the transfer execution if it has not yet reached a terminal
+        state, waits up to 60s for it to finish, then deletes it.
+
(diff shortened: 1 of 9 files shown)


coriolis/tests/integration/base.py
```diff
@@ -389,6 +416,32 @@ def assertExecutionErrored(self, execution_id, timeout=300):
...
+        waits up to 60s for it to finish, then deletes it.
+
+        This avoids the 400 HTTP "cannot delete a RUNNING deployment" error
+        that occurs when a deployment is still in-flight at cleanup time, which
+        can happen with slow providers when a test fails or times out before
+        the deployment completes.
+        """
+        ctxt = self._get_db_context()
+        try:
+            deployment = db_api.get_deployment(ctxt, deployment_id)
```
petrutlucian94: Mixing DB requests with Coriolis API requests feels a bit odd but I guess it's reasonable in this case.

petrutlucian94 (approved):

PR #441, opened 2026-05-26
Allow filtering results by status

petrutlucian94:
We're adding a "status" search option that can be specified when listing the following resources:

* deployments
    * status -> last_execution_status db field
* transfers
* transfer executions

The filtering will be performed on the db side.

Client PR: https://github.com/cloudbase/python-coriolisclient/pull/100


coriolis/tests/integration/test_api_listing.py
```diff
@@ -295,3 +298,78 @@ def test_transfer_list_pagination(self):
...
+        self.assertRaises(
+            http_exc.BadRequest,
+            self._client.transfers.list,
+            filters={"status": "fake-status"})
+
+        self.assertRaises(
+            http_exc.BadRequest,
+            self._client.transfer_executions.list,
+            self._completed_transfer.id,
+            filters={"status": "fake-status"})
```
claudiubelu, line 375: maybe also include a bad filter case?

coriolis/tests/integration/test_api_listing.py
```diff
@@ -295,3 +298,78 @@ def test_transfer_list_pagination(self):
...
+        self.assertEqual(exp_summary, ret_summary)
+
+    def test_transfer_filter(self):
+        executions = self._client.transfers.list(
+            filters={"status": constants.TASK_STATUS_COMPLETED})
+        ret_summary = [self._get_record_summary(e) for e in executions]
+        exp_summary = [self._get_record_summary(self._completed_transfer)]
+        self.assertEqual(exp_summary, ret_summary)
+
+    def test_transfer_deployment(self):
```
claudiubelu: nit: since this is testing the deployments, `test_deployment_filter`?

coriolis/deployer_manager/rpc/server.py
```diff
@@ -123,10 +123,10 @@ def _loop(self):
             try:
                 deployments = self._rpc_conductor_client.get_deployments(
                     self._admin_ctx, include_tasks=False,
-                    include_task_info=False)
-                for d in deployments:
-                    if d['last_execution_status'] == PENDING_STATUS:
-                        self._check_deployer_status(d['id'])
+                    include_task_info=False,
+                    filters={'last_execution_status': PENDING_STATUS})
```
claudiubelu: `last_execution_status`? is it not `status`? from what I see in `coriolis.db.api`, you do pop `status`, not `last_execution_status`.

coriolis/tests/integration/test_api_listing.py
petrutlucian94: Whops, right.

coriolis/deployer_manager/rpc/server.py
petrutlucian94: Whops, my bad. Thanks for noticing. Wondering how come the tests passed then :).

coriolis/tests/integration/test_api_listing.py
petrutlucian94: It turns out that the Coriolis API silently ignores unknown GET parameters, just like Nova did until a few years ago. Fixing it would require a larger change and it's outside the scope of this PR.

claudiubelu (approved):

PR #443, opened 2026-05-27
conductor: fix on-error cleanup skipped when source task cancelled-after-completion

claudiubelu:
When concurrently deploying resources on the source (deploy_source) and on the target (deploy_target), if deploy_target fails while deploy_source is still running, the conductor marks deploy_source as `CANCELLING` and sends it a cancel signal.

If deploy_source finishes its work before the signal kills it, `task_completed()` is sent for a task in the `CANCELLING` state. The conductor saves the result (so source_resources, including the resource ID is saved in the DB) and marks the task `CANCELED_AFTER_COMPLETION`.

The on-error cleanup scheduling logic in _advance_execution_state checks:

```
elif TASK_STATUS_COMPLETED in non_error_parents.values():
```

`CANCELED_AFTER_COMPLETION` != `COMPLETED`, so the check is False. `DELETE_TRANSFER_SOURCE_RESOURCES` is unscheduled and the source resource is leaked.

Adding `CLEANUP_TASK_TRIGGER_STATUSES`, covering both `COMPLETED` and `CANCELED_AFTER_COMPLETION`, since both mean the task ran to completion and may have created resources that need cleaning up. Use that constant in the on-error scheduling check.

Adding assertions in `coriolis/tests/integration/test_failure_recovery.py`, that check in the DB that `DELETE_TRANSFER_SOURCE_RESOURCES` / `DELETE_TRANSFER_TARGET_RESOURCES` reached a completed status, and that the source /target resources were zeroed out in the action info.

### DIFF
--- a/coriolis/tests/integration/test_failure_recovery.py
+++ b/coriolis/tests/integration/test_failure_recovery.py
@@ -4,32 +4,123 @@
 """
 Integration test for failure handling and cleanup.
 
-Patches the active import provider's deploy_replica_target_resources to raise
-an exception, triggers a transfer execution, and asserts that:
-  1. The execution reaches ERROR status.
-  2. Cleanup tasks (delete_replica_source_resources) ran so the replicator
-     process is no longer alive.
+Tests two symmetrical error-path scenarios:
+
+- deploy_replica_target_resources is delayed then raises -> source task
+  completes first (creating real source resources) -> execution errors ->
+  source cleanup (DELETE_TRANSFER_SOURCE_RESOURCES) must run,
+  source_resources zeroed out.
+
+- deploy_replica_source_resources is delayed then raises -> target task
+  completes first (creating real target resources) -> execution errors ->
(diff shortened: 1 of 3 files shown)


petrutlucian94 (approved):

coriolis/tests/integration/test_failure_recovery.py
```diff
@@ -4,32 +4,127 @@
...
+            "%s task not found in execution %s" % (task_type, execution_id),
+        )
+
+        completed_statuses = (
+            constants.TASK_STATUS_COMPLETED,
+            constants.TASK_STATUS_CANCELED_AFTER_COMPLETION,
+        )
+        self.assertIn(
+            cleanup_task.status,
+            completed_statuses,
```
Dany9966: Why not use `CLEANUP_TASK_TRIGGER_STATUSES` here?

claudiubelu: good point, done.

Dany9966 (approved):

PR #444, opened 2026-05-27
Change set_dhcp option to create nmconnection files for Redhat 9+

fabi200123:
Currently when the set_dhcp option is set to true, it creates ifcfg files for all redhat versions.
From Rhel8, we also require to have the `NM_CONTROLLED=yes`, since in this version there is support for nmconnection files.
From Rhel9, it has changed from ifcfg files to nmconnection files, thus this PR adds a template and creates nmconnection files for all distros that have minimum version 9.

### DIFF
--- a/coriolis/osmorphing/netpreserver/ifcfg.py
+++ b/coriolis/osmorphing/netpreserver/ifcfg.py
@@ -1,7 +1,6 @@
 # Copyright 2025 Cloudbase Solutions Srl
 # All Rights Reserved.
 
-import os
 import re
 
 from oslo_log import log as logging
@@ -13,23 +12,24 @@ LOG = logging.getLogger(__name__)
 
 class IfcfgNetPreserver(base.BaseNetPreserver):
 
-    def __init__(self, osmorphing_tool):
-        super(IfcfgNetPreserver, self).__init__(osmorphing_tool)
-        self.network_scripts_path = "etc/sysconfig/network-scripts"
-
     def check_net_preserver(self):
-        if self.osmorphing_tool._test_path(self.network_scripts_path):
-            ifcfg_files = self._get_net_config_files(self.network_scripts_path)
+        network_scripts_path = (
+            self.osmorphing_tool._NETWORK_SCRIPTS_PATH)
+        if self.osmorphing_tool._test_path(network_scripts_path):
+            ifcfg_files = self.osmorphing_tool._get_net_config_files(
+                network_scripts_path)
             if ifcfg_files:
-                ifcfgs_ethernet = self._get_ifcfgs_by_type(
-                    "Ethernet", self.network_scripts_path)
+                ifcfgs_ethernet = self.osmorphing_tool._get_ifcfgs_by_type(
+                    "Ethernet", network_scripts_path)
                 if ifcfgs_ethernet:
                     return True
         return False
 
     def parse_network(self):
-        ifcfgs_ethernet = self._get_ifcfgs_by_type(
-            "Ethernet", self.network_scripts_path)
+        network_scripts_path = (
+            self.osmorphing_tool._NETWORK_SCRIPTS_PATH)
+        ifcfgs_ethernet = self.osmorphing_tool._get_ifcfgs_by_type(
+            "Ethernet", network_scripts_path)
         if ifcfgs_ethernet:
             for ifcfg_file, ifcfg in ifcfgs_ethernet:
                 name = ifcfg.get("DEVICE")
@@ -42,20 +42,3 @@ class IfcfgNetPreserver(base.BaseNetPreserver):
                     "mac_address": mac_address,
                     "ip_addresses": [ip_address] if ip_address else []
                 }
-
-    def _get_net_config_files(self, network_scripts_path):
-        dir_content = self.osmorphing_tool._list_dir(network_scripts_path)
-        return [os.path.join(network_scripts_path, f) for f in
-                dir_content if re.match("^ifcfg-(.*)", f)]
-
-    def _get_ifcfgs_by_type(self, ifcfg_type, network_scripts_path):
-        ifcfgs = []
-        for ifcfg_file in self._get_net_config_files(network_scripts_path):
-            ifcfg = self.osmorphing_tool._read_config_file_sudo(ifcfg_file)
-            detected_type = ifcfg.get('TYPE')
-            if not detected_type:
-                detected_type = "Ethernet"
-                ifcfg["TYPE"] = detected_type
-            if detected_type.lower() == ifcfg_type.lower():
-                ifcfgs.append((ifcfg_file, ifcfg))
-        return ifcfgs
--- a/coriolis/osmorphing/netpreserver/nmconnection.py
+++ b/coriolis/osmorphing/netpreserver/nmconnection.py
@@ -1,7 +1,6 @@
 # Copyright 2025 Cloudbase Solutions Srl
 # All Rights Reserved.
 
-import os
 import re
 
 from oslo_log import log as logging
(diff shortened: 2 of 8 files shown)


coriolis/osmorphing/redhat.py
```diff
@@ -172,7 +253,10 @@ def _comment_keys_from_ifcfg_files(
     def set_net_config(self, nics_info, dhcp):
         if dhcp:
             self.disable_predictable_nic_names()
-            self._write_nic_configs(nics_info)
+            if self._uses_nmconnection_net_config():
```
Dany9966: Since we have some duplicate logic here and in netpreserver classes, I think we should refactor that code into osmorphing.
Then you could do something like this:
```python
# this would be the reliable source of truth for a machine using nmconnections or not
# looking for version alone may not be enough.
ethernet_keyfiles = None
if self._test_path(self._NM_CONNECTIONS_PATH):
    ethernet_keyfiles = self._get_keyfiles_by_type("ethernet", self._NM_CONNECTIONS_PATH)
if ethernet_keyfiles:
    self._write_nmconnection_configs(nics_info)
else:
    self._write_nic_configs(nics_info)
```

Dany9966 (changes requested):

coriolis/osmorphing/redhat.py
```diff
@@ -119,10 +153,14 @@ def _set_dhcp_net_config(self, ifcfgs_ethernet):
             del network_cfg["GATEWAY"]
             self._write_config_file(network_cfg_file, network_cfg)
 
+    def _uses_nmconnection_net_config(self):
+        return bool(
+            self._get_nmconnection_net_preserver().get_nmconnection_files())
```
Dany9966: No. This is not what was asked. What was asked is to move the common methods from netpreserver into the osmorphing class, NOT the other way around. As the netpreservers are only meant to be used when actually setting up preservation, I'd like to not instantiate them when not actually needed, to avoid eventual spagetti code.

Please make sure you do the following for a proper refactor:
* Move all the code that will be used in both the osmorphing tool and netpreserver into osmorphing tool. (specifically what we need moved is: `ifcfg._get_net_config_files`, `ifcfg._get_ifcfgs_by_type`, `nmconnection._get_nmconnection_files`, `nmconnection._get_keyfiles_by_type` - it's alright if you move these to `osmorphing.base`);
* Inside netpreserver, use `self.osmorphing_tool` to reference those common methods;
* Make sure to ONLY use a netpreserver object inside of `_setup_network_preservation`, no other need for it.

Dany9966 (changes requested):

coriolis/osmorphing/netpreserver/ifcfg.py
Dany9966: Please refactor this value and the one in redhat module into base, so we have a single common value for everything.

coriolis/osmorphing/netpreserver/ifcfg.py
```diff
@@ -19,16 +18,17 @@ def __init__(self, osmorphing_tool):
 
     def check_net_preserver(self):
         if self.osmorphing_tool._test_path(self.network_scripts_path):
-            ifcfg_files = self._get_net_config_files(self.network_scripts_path)
+            ifcfg_files = self.osmorphing_tool._get_net_config_files(
+                self.network_scripts_path)
             if ifcfg_files:
-                ifcfgs_ethernet = self._get_ifcfgs_by_type(
+                ifcfgs_ethernet = self.osmorphing_tool._get_ifcfgs_by_type(
```
Dany9966, line 22: As a future improvement (in another PR), can you please make the netpreservers NOT read through the files twice? Maybe just re-use the object in the factory instead of the class, or return this collection of interface files instead of a bool value, and pass it down to `parse_network`.

coriolis/osmorphing/netpreserver/nmconnection.py
Dany9966: Same as for the value for ifcfgs.

coriolis/osmorphing/redhat.py
```diff
@@ -119,10 +143,34 @@ def _set_dhcp_net_config(self, ifcfgs_ethernet):
...
+            return
+        nmconnection_files = self._get_nmconnection_files(
+            self._NM_CONNECTIONS_PATH)
+        for cfg_path in nmconnection_files:
+            self._exec_cmd_chroot(
+                'mv "%s" "%s%s"' % (cfg_path, cfg_path, backup_file_suffix))
+            LOG.debug("Backed up nmconnection profile '%s'", cfg_path)
+
+    def _backup_ifcfg_configs(self, device_names, backup_file_suffix=".bak"):
+        for dev_name in device_names:
```
Dany9966: Why do we need a list of device_names here? Shouldn't we back up every interface file?

coriolis/osmorphing/redhat.py
```diff
@@ -131,7 +179,29 @@ def _write_nic_configs(self, nics_info):
...
+                })
+
+    def _write_nmconnection_configs(self, nics_info):
+        nics_info = nics_info or []
+        if not nics_info:
+            return
+
+        self._backup_nmconnection_files()
+        device_names = ["eth%d" % idx for idx, _ in enumerate(nics_info)]
+        self._backup_ifcfg_configs(device_names)
```
Dany9966: Please add an inline comment here stating that redhat-based machines may have both file types, so we need to back up both.

coriolis/osmorphing/redhat.py
```diff
@@ -119,10 +143,34 @@ def _set_dhcp_net_config(self, ifcfgs_ethernet):
...
 
+    def _uses_nmconnection_net_config(self):
+        if not self._test_path(self._NM_CONNECTIONS_PATH):
+            return False
+        return bool(self._get_nmconnection_files(self._NM_CONNECTIONS_PATH))
+
+    def _backup_nmconnection_files(self, backup_file_suffix=".bak"):
+        if not self._test_path(self._NM_CONNECTIONS_PATH):
+            return
+        nmconnection_files = self._get_nmconnection_files(
```
Dany9966: Can we please avoid having to read through files multiple times?

Dany9966 (changes requested):

coriolis/osmorphing/redhat.py
```diff
@@ -119,19 +141,67 @@ def _set_dhcp_net_config(self, ifcfgs_ethernet):
             del network_cfg["GATEWAY"]
             self._write_config_file(network_cfg_file, network_cfg)
 
+    def _get_existing_nmconnection_files(self):
+        if not self._test_path(self._NM_CONNECTIONS_PATH):
+            return []
+        return self._get_nmconnection_files(self._NM_CONNECTIONS_PATH)
+
+    def _uses_nmconnection_net_config(self):
```
Dany9966: Is this method used anywhere? Please remove it if it's not.

coriolis/osmorphing/redhat.py
```diff
@@ -119,19 +141,67 @@ def _set_dhcp_net_config(self, ifcfgs_ethernet):
...
+
+    def _write_nmconnection_configs(self, nics_info, nmconnection_files):
+        nics_info = nics_info or []
+        if not nics_info:
+            return
+
+        # Red Hat-based systems may have both nmconnection keyfiles and legacy
+        # ifcfg profiles; back up both so stale source configs cannot override
+        # the freshly written DHCP profiles.
+        self._backup_nmconnection_files(nmconnection_files)
```
Dany9966, line 191: Since we only create `Ethernet` type connections, I think we should only back up those types of files. The rest may remain the same. The current code backs up all of them, which is wrong and can break bridge/bond/vpn/etc. type files after migration.

coriolis/osmorphing/redhat.py
```diff
@@ -119,19 +141,67 @@ def _set_dhcp_net_config(self, ifcfgs_ethernet):
...
+            return []
+        return self._get_nmconnection_files(self._NM_CONNECTIONS_PATH)
+
+    def _uses_nmconnection_net_config(self):
+        return bool(self._get_existing_nmconnection_files())
+
+    def _backup_nmconnection_files(self, nmconnection_files=None,
+                                   backup_file_suffix=".bak"):
+        if nmconnection_files is None:
+            nmconnection_files = self._get_existing_nmconnection_files()
```
Dany9966: Please make this code reflect the fact that we only have to back up `Ethernet` type files.

coriolis/osmorphing/redhat.py
```diff
@@ -119,19 +141,67 @@ def _set_dhcp_net_config(self, ifcfgs_ethernet):
...
+        for cfg_path in nmconnection_files:
+            self._exec_cmd_chroot(
+                'mv "%s" "%s%s"' % (cfg_path, cfg_path, backup_file_suffix))
+            LOG.debug("Backed up nmconnection profile '%s'", cfg_path)
+
+    def _backup_all_ifcfg_configs(self, backup_file_suffix=".bak"):
+        if not self._test_path(self._NETWORK_SCRIPTS_PATH):
+            return
+        for cfg_path in self._get_net_config_files(self._NETWORK_SCRIPTS_PATH):
```
Dany9966: Same here. Please make sure to ONLY back up `Ethernet` type files.

coriolis/tests/osmorphing/test_redhat.py
```diff
@@ -149,32 +149,325 @@ def test__set_dhcp_net_config(self, mock_read_config_file,
...
+        mock_get_existing_nmconnection_files.return_value = []
+        nics_info = [{
+            'mac_address': mock.sentinel.mac_address,
+        }]
+        dhcp = True
+
+        self.morphing_tools.set_net_config(nics_info, dhcp)
+
+        mock_disable_predictable_nic_names.assert_called_once()
+        mock_write_nic_configs.assert_called_once_with(nics_info)
```
Dany9966, line 405: Can you please mock `_get_existing_nmconnection_files` and assert not called?

coriolis/osmorphing/redhat.py
```diff
@@ -170,7 +240,15 @@ def _comment_keys_from_ifcfg_files(
     def set_net_config(self, nics_info, dhcp):
         if dhcp:
             self.disable_predictable_nic_names()
```
Dany9966: I think you should move this call below the `nics_info` check. If it's empty, there's no point in calling this.

Dany9966 (changes requested):

Dany9966 (approved):

PR #445, opened 2026-05-27
Add `SLE` module activation into `BaseSUSEMorphingTools`

mihaelabalutoiu:
Providers override `_get_sle_modules()` to declare which SLE modules are needed.

### DIFF
--- a/coriolis/osmorphing/suse.py
+++ b/coriolis/osmorphing/suse.py
@@ -129,6 +129,23 @@ class BaseSUSEMorphingTools(base.BaseLinuxOSMorphingTools):
         except Exception:
             return False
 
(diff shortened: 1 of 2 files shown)


petrutlucian94 (approved):

coriolis/osmorphing/suse.py
```diff
@@ -127,6 +129,23 @@ def _has_systemd(self):
         except Exception:
             return False
 
+    def _get_sle_modules(self):
```
petrutlucian94, line 132: The naming is a bit inconsistent: `_get_sle_modules` vs `_enable_sles_module`.

Anyway, since SUSE refers to them as "SLE modules", I guess it's reasonable.

PR #447, opened 2026-05-28
integration: Allow running with external import providers

claudiubelu:
Currently, we create a destination device in the test base. However, the import provider is supposed to create the disks for transferred as required.

Moving this part into the test provider will make it easier for us to swap in other providers later on.

Introduces an `BaseTestImportProvider` ABC that contains provider-specific logic not currently defined in the import providers, meant to be used for testing-only purposes:
- detect leaked resources
- delete deployed replicas

Adds an implementation for it, in `TestImportProvider`.

Adds `providers.yaml` support for external providers:

- `_load_providers_config()`: parses the YAML file present at the `CORIOLIS_PROVIDERS_YAML` environment variable, which describes the destination provider, its connection info, target environment, storage mappings. If `CORIOLIS_PROVIDERS_YAML` is unset, the default test provider will be used instead.
- `CORIOLIS_CONFIG_FILE`: optional `oslo.config` file path, so provider packages can register and read their own config options.
- `providers.yaml.sample`: sample file for declaring providers.

Changes needed for using other providers:
- `tox.ini`: pass all `CORIOLIS_*` env vars through; set `PBR_VERSION` to avoid pbr version-detection failures.
- Randomizes `instance_name` used for transfer tests to avoid conflict.
- Fixes `connection_info` reference in `test_minion_pools.py`.
- Machine Pool allocation may take longer for other providers. 600s gives sufficient headroom.

### DIFF
--- a/coriolis/tests/integration/harness.py
+++ b/coriolis/tests/integration/harness.py
@@ -25,6 +25,7 @@ import shutil
 import socket
 import subprocess
 import tempfile
+from unittest import mock
 import uuid
 
 from cheroot.workers import threadpool as cheroot_threadpool
@@ -50,6 +51,7 @@ from coriolis.deployer_manager.rpc import server as deployer_manager_rpc_server
 from coriolis import exception
 from coriolis.minion_manager.rpc import server as minion_manager_rpc_server
 from coriolis import policy as policy_module
+from coriolis.providers import factory as providers_factory
 from coriolis import rpc as rpc_module
 from coriolis.scheduler.rpc import server as scheduler_rpc_server
 from coriolis import service
@@ -311,10 +313,16 @@ class _IntegrationHarness:
 
         self.imp_provider_class = _get_provider(_TEST_IMPORT_PROVIDER)
         self.imp_provider_platform = self.imp_provider_class.platform
+        self.imp_provider = providers_factory.get_provider(
+            self.imp_provider_platform,
+            constants.PROVIDER_TYPE_TRANSFER_IMPORT,
+            event_handler=mock.MagicMock(),
+        )
         self.imp_conn_info = {
             "pkey_path": self.ssh_key_path,
             "role": "destination",
         }
+        self.imp_env_options = {}
 
         self._wsgi_server = None
         self._wsgi_server_thread = None
--- a/coriolis/tests/integration/harness.py
+++ b/coriolis/tests/integration/harness.py
@@ -57,6 +57,7 @@ from coriolis.scheduler.rpc import server as scheduler_rpc_server
 from coriolis import service
 from coriolis.taskflow import runner as taskflow_runner
 from coriolis.tasks import factory as task_runners_factory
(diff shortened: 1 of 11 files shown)


coriolis/tests/integration/provider_harness.py
```diff
@@ -0,0 +1,69 @@
...
+
+        Returns a dict merged into ``destination_environment`` for this
+        transfer. Must register ``addCleanup()`` on *test_instance* for
+        every resource created, so it is released even on test failure.
+        """
+        return {}
+
+    @property
+    def are_disks_local(self) -> bool:
+        """True if destination disks are local block devices on the test host.
```
petrutlucian94: I don't like this design. It doesn't really matter if the disks are local, provider agnostic tests shouldn't directly touch resources created by the providers, otherwise they're no longer provider agnostic.

We should clearly separate provider agnostic tests from those that rely on the "test provider" and interact with the created resources (e.g. scsi debug devices).

Without a proper design and clear limitations, our tests are going to become a mess and we'll refactor them every other week. We don't want that.

For what is worth, Openstack's Tempest tests are completely driver agnostic. They do have "functional tests" that call directly into the provider, but hose are completely separate.

https://github.com/openstack/nova/tree/master/nova/tests/functional/libvirt

coriolis/tests/integration/providers.yaml.sample
```diff
@@ -0,0 +1,28 @@
...
+  provider: "coriolis.tests.integration.test_provider.imp.TestImportProvider"
+
+  # connection_info is passed to the destination endpoint.
+  connection_info: null
+
+  # destination_environment options forwarded to each transfer / deployment.
+  environment: {}
+
+  # Storage backend mapping (source identifier -> destination pool / datastore).
+  storage_mappings: {}
```
petrutlucian94, line 28: I think we'll also need source/target environment settings.

coriolis/tests/integration/harness.py
```diff
@@ -514,3 +580,10 @@ def _teardown(self):
             test_utils.destroy_scsi_debug()
         except Exception:
             pass
+
+    def uses_test_import_provider(self):
```
petrutlucian94: It's a bit confusing since we'd always use a test provider that implements `BaseTestImportProvider`.

So I guess even for Openstack for example we'd need to have a test provider.

Would `uses_core_test_import_provider` make sense?

claudiubelu: fair point. Done.

coriolis/tests/integration/providers.yaml.sample
claudiubelu: we already have the target environment settings above (`destination.settings`). Source environment settings will have to be addressed later, when we'll essentially do the same thing for the source provider.

petrutlucian94 (approved):
It's much cleaner now, lgtm.

PR #448, opened 2026-05-29
Allow registering replica-first-boot scripts

petrutlucian94:
We've recently extended the Coriolis API to allow specifying *when* a given user script should be executed. We currently support the following phases:

* osmorphing_pre_os_mount
* osmorphing_post_os_mount

For convenience, we'll also add the following phase:

* replica_first_boot

Sample:

```
coriolis deployment create fc85dc97-8e6d-40bf-8f44-1884b0b4251f  \
    --user-script-global windows=/root/user_scripts/bitlocker/suspend_using_pwd.ps1,phase=osmorphing_pre_os_mount
    --user-script-global windows=/root/user_scripts/resume_bitlocker.ps1,phase=replica_first_boot
```

These scripts will be executed when the replica VM boots for the first time. We'll inject them during os-morphing. As you may have noticed, the user can already do this but it's quite inconvenient to pass a script that injects another first-boot script.

We'll rely on systemd on Linux and cloudbase-init on Windows. All the Linux distributions that we support have switched to Systemd about 10 years ago. Also, as per this commit [1], we rely on the fact that cloudbase-init will be available all the time and that we can use it to run first-boot scripts.

Note that we did consider using scheduled tasks on Windows (as opposed to Cloudbase-init). We'd need to use an xml task definition and register it using registry keys, however it seems like we lack the privileges to create entries such as the following:

```
$HIVE\Microsoft\Windows NT\CurrentVersion\Schedule\TaskCache\$TaskGUID"
```

[1] https://github.com/cloudbase/coriolis/commit/b5f93fdb87c2c0735fb065be5a1c378c0bf505d6

### DIFF
--- a/coriolis/osmorphing/base.py
+++ b/coriolis/osmorphing/base.py
@@ -29,14 +29,24 @@ CLOUD_INIT_SERVICE_UNIT_NAME = "cloud-init"
 CLOUD_INIT_SERVICE_UNIT_NAME_FALLBACK = "cloud-init-main"
 
 FIRST_BOOT_SCRIPT_RUNNER = """#!/bin/bash
+
+first_error=0
+
 function run_scripts {
     script_dir=$1
 
-    for f in $script_dir/*.sh; do
+    for f in "$script_dir"/*.sh; do
         if [ -x "$f" ]; then
             echo "Invoking script: $f"
             "$f"
-            echo "Exit code: $?"
+            rc=$?
+            echo "Exit code: $rc"
+
+            if [ $rc -ne 0 ] && [ $first_error -eq 0 ]; then
+                first_error=$rc
+            fi
+        else
+            echo "Ignoring script, not executable: $f"
         fi
     done
 }
@@ -47,9 +57,17 @@ run_scripts /usr/lib/coriolis/firstboot/service
 # Run user provided scripts.
 run_scripts /usr/lib/coriolis/firstboot/user
 
-mkdir -p /var/lib/coriolis
-touch /var/lib/coriolis/firstboot-complete
+if [ $first_error -eq 0 ]; then
+    echo "All the scripts completed successfully, creating /var/lib/coriolis/firstboot-complete"
+    mkdir -p /var/lib/coriolis
+    touch /var/lib/coriolis/firstboot-complete
+else
+    echo "One of the scripts failed, won't create /var/lib/coriolis/firstboot-complete"
+fi
+
+exit $first_error
 """
+
 FIRST_BOOT_SCRIPT_RUNNER_PATH = "/usr/lib/coriolis/firstboot/run-firstboot.sh"
 FIRST_BOOT_SYSTEMD_UNIT = """
 [Unit]
@@ -62,10 +80,13 @@ ConditionPathExists=!/var/lib/coriolis/firstboot-complete
 Type=oneshot
 ExecStart=/usr/lib/coriolis/firstboot/run-firstboot.sh
 RemainAfterExit=yes
+StandardOutput=journal+console
+StandardError=journal+console
 
 [Install]
 WantedBy=multi-user.target
 """
+
 FIRST_BOOT_SYSTEMD_UNIT_NAME = "coriolis-firstboot.service"
 FIRST_BOOT_SYSTEMD_UNIT_PATH = (
     f"/etc/systemd/system/{FIRST_BOOT_SYSTEMD_UNIT_NAME}")
--- a/coriolis/osmorphing/base.py
+++ b/coriolis/osmorphing/base.py
@@ -28,6 +28,48 @@ DEFAULT_CLOUD_USER = "cloud-user"
 CLOUD_INIT_SERVICE_UNIT_NAME = "cloud-init"
 CLOUD_INIT_SERVICE_UNIT_NAME_FALLBACK = "cloud-init-main"
 
+FIRST_BOOT_SCRIPT_RUNNER = """#!/bin/bash
+function run_scripts {
+    script_dir=$1
+
+    for f in $script_dir/*.sh; do
+        if [ -x "$f" ]; then
+            echo "Invoking script: $f"
+            "$f"
+            echo "Exit code: $?"
+        fi
+    done
+}
+
+# Run Coriolis provided scripts.
+run_scripts /usr/lib/coriolis/firstboot/service
+
+# Run user provided scripts.
+run_scripts /usr/lib/coriolis/firstboot/user
+
+mkdir -p /var/lib/coriolis
+touch /var/lib/coriolis/firstboot-complete
+"""
+FIRST_BOOT_SCRIPT_RUNNER_PATH = "/usr/lib/coriolis/firstboot/run-firstboot.sh"
+FIRST_BOOT_SYSTEMD_UNIT = """
+[Unit]
+Description=Coriolis replica first-boot scripts.
+After=network-online.target
+Wants=network-online.target
+ConditionPathExists=!/var/lib/coriolis/firstboot-complete
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/coriolis/firstboot/run-firstboot.sh
+RemainAfterExit=yes
+
+[Install]
+WantedBy=multi-user.target
+"""
+FIRST_BOOT_SYSTEMD_UNIT_NAME = "coriolis-firstboot.service"
+FIRST_BOOT_SYSTEMD_UNIT_PATH = (
+    f"/etc/systemd/system/{FIRST_BOOT_SYSTEMD_UNIT_NAME}")
+
 
 class BaseOSMorphingTools(object, with_metaclass(abc.ABCMeta)):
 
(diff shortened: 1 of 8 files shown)


petrutlucian94:
Client PR: https://github.com/cloudbase/python-coriolisclient/pull/101

coriolis/osmorphing/base.py
```diff
@@ -28,6 +28,48 @@
 CLOUD_INIT_SERVICE_UNIT_NAME = "cloud-init"
 CLOUD_INIT_SERVICE_UNIT_NAME_FALLBACK = "cloud-init-main"
 
+FIRST_BOOT_SCRIPT_RUNNER = """#!/bin/bash
+function run_scripts {
+    script_dir=$1
+
+    for f in $script_dir/*.sh; do
+        if [ -x "$f" ]; then
```
claudiubelu, line 39: It seems that we're only executing scripts that have the executable flag. Should we log a warning if it's not executable? It might be useful to at least know that a script has been skipped. Nonetheless, if a script is not executable, we might have done something wrong in the first place.

coriolis/osmorphing/base.py
```diff
@@ -28,6 +28,48 @@
...
+            echo "Invoking script: $f"
+            "$f"
+            echo "Exit code: $?"
```
claudiubelu: So, we're running all the scripts in a best effort kind of way. But how will the user know something went wrong? Technically, we never have any non-zero exit anywhere, so technically, at first glance, you'd be inclined to believe everything executed properly. You'd only notice that something went wrong when checking the logs. In addition to that, we still touch `/var/lib/coriolis/firstboot-complete`, so the failed script is never retried anyways.

To give LUKS first boot script as an example, this might be an important failure to flag, as the migration is not considered complete if it failed.

coriolis/osmorphing/base.py
```diff
@@ -28,6 +28,48 @@
...
+
+# Run Coriolis provided scripts.
+run_scripts /usr/lib/coriolis/firstboot/service
+
+# Run user provided scripts.
+run_scripts /usr/lib/coriolis/firstboot/user
+
+mkdir -p /var/lib/coriolis
+touch /var/lib/coriolis/firstboot-complete
+"""
```
claudiubelu, line 71: newline after this?

coriolis/osmorphing/base.py
```diff
@@ -28,6 +28,48 @@
...
+ConditionPathExists=!/var/lib/coriolis/firstboot-complete
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/coriolis/firstboot/run-firstboot.sh
+RemainAfterExit=yes
+
+[Install]
+WantedBy=multi-user.target
+"""
```
claudiubelu, line 90: ditto

coriolis/osmorphing/base.py
```diff
@@ -28,6 +28,48 @@
...
+[Unit]
+Description=Coriolis replica first-boot scripts.
+After=network-online.target
+Wants=network-online.target
+ConditionPathExists=!/var/lib/coriolis/firstboot-complete
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/coriolis/firstboot/run-firstboot.sh
+RemainAfterExit=yes
```
claudiubelu, line 84: It might also be useful to send the output to the serial console, especially if something failed. This has been useful for me, debugging the LUKS first boot script without having to log in into the VM. Though, we'll have to be extra careful with `set -x` scripts then, so sensitive data in the scripts are not leaked:

```
StandardOutput=journal+console
StandardError=journal+console
```

coriolis/osmorphing/base.py
```diff
@@ -719,3 +770,53 @@ def _setup_network_preservation(self, nics_info) -> None:
...
 
         return
+
+    def register_firstboot_script(
+        self,
+        script: str,
+        index: int = 0,
+        user_provided=True,
+    ):
+        if len(script) == 0:
```
claudiubelu: In what scenario is this true?

coriolis/osmorphing/base.py
```diff
@@ -719,3 +770,53 @@ def _setup_network_preservation(self, nics_info) -> None:
...
+            LOG.debug("Empty first-boot script, skipping...")
+            return
+
+        if user_provided:
+            script_dir = "/usr/lib/coriolis/firstboot/user"
+        else:
+            script_dir = "/usr/lib/coriolis/firstboot/service"
+        unique_id = str(uuid.uuid4()).split("-")[0]
+        script_path = os.path.join(script_dir, f"{index:02d}-{unique_id}.sh")
```
claudiubelu: This might be a bit confusing: if you check the logs, you'd see that script '10-deadbeef.sh` failed; you'd only know what failed if you explicitly check what that script contains. Do we not have the script names?

coriolis/tests/integration/deployments/test_osmorphing.py
```diff
@@ -135,3 +137,54 @@ def test_os_morphing_global_script_extended_format(self):
...
+                    first_boot_script_dir, file_name)
+                first_boot_script = test_utils.read_file_from_device(
+                    self._dst_device,
+                    first_boot_script_path)
+                if payload == first_boot_script:
+                    found = True
+
+        if not found:
+            raise AssertionError(
+                "Couldn't find the expected first boot script.")
```
claudiubelu, line 193: Can you also assert that the windows one is not here? Just in case.

coriolis/tests/osmorphing/test_manager.py
```diff
@@ -29,6 +29,10 @@ def setUp(self):
                 "phase": constants.PHASE_OSMORPHING_POST_OS_MOUNT,
                 "payload": "post-os-mount-script",
             },
+            {
+                "phase": constants.PHASE_REPLICA_FIRST_BOOT,
+                "payload": "fist-boot-script",
```
claudiubelu: nit: `✊-👢-📜` (`fist-boot-script`) -> `first-boot-script`? :)

coriolis/osmorphing/base.py
petrutlucian94: Sure, we can log a warning.

coriolis/osmorphing/base.py
petrutlucian94: It's debatable. I'm not sure if retrying the failed scripts would actually help. We'd also end up retrying scripts that did succeed.

We could return a non-zero exit code so that the "coriolis-firstboot" unit ends up in failed state, telling the user that something went wrong.

For what is worth, the `LocalScripts` cloudbase-init plugin seems to ignore these failures, only checking if a reboot is necessary based on the return code: https://github.com/cloudbase/cloudbase-init/blob/master/cloudbaseinit/plugins/common/localscripts.py

coriolis/osmorphing/base.py
petrutlucian94: That would also apply to user provided scripts, which may end up leaking sensitive data.

Worst case scenario, we can make it configurable.

coriolis/osmorphing/base.py
petrutlucian94: The user may provide an empty script. I thought that a sanity check wouldn't hurt.

coriolis/osmorphing/base.py
petrutlucian94: Good point.

The same method is used for user provided scripts, in which case we don't have a script name.

Well, I can add a "script_filename" optional argument.

coriolis/osmorphing/base.py
petrutlucian94: All the `FIRST_BOOT_*` constants are defined one after the other, which is why I didn't put additional newlines. I can add them if you think that it improves readability.

coriolis/osmorphing/base.py
petrutlucian94: I've updated the script to propagate errors and skip creating the "firstboot-complete" file in case of failures.

coriolis/osmorphing/base.py
petrutlucian94: For what is worth, I've updated the check to use `if not script` so that we also cover `None`, just in case.

claudiubelu: Ye, but... why would we accept empty scripts? That sounds like a mistake we should raise on the client side; the user might have made a mistake (e.g.: forgot to save the file before uploading).

coriolis/osmorphing/base.py
claudiubelu: Indeed, and the same problem applies there. This problem didn't exist before, as we only supported one user script. 

I think the `script_filename` can be deduced from the file passed in.

claudiubelu (approved):

coriolis/osmorphing/base.py
petrutlucian94: > I think the script_filename can be deduced from the file passed in.

For user provided scripts? We'd have to extend the API to also receive the script file name, in addition to the script contents and execution phase. We'd also have to be careful with potential duplicate file names when receiving inputs such as:

```
coriolis deployment create $transfer  \
    --user-script-global windows=pre_mount/script.ps1,phase=osmorphing_pre_os_mount
    --user-script-global windows=first_boot/script.ps1,phase=replica_first_boot
```

It would probably have to prepend the index (priority) on Windows so that it won't get executed before the Coriolis scripts, if any.

We can have a follow-up PR if deemed necessary.

coriolis/osmorphing/base.py
petrutlucian94: Well, I can send a separarate patch to reject empty user scripts. We'd need it on the API side to also cover requests coming from `coriolis-web`.

I just wanted to be careful about this since users can omit the script path when updating transfers in order to remove user scripts.

petrutlucian94: Here's a separate PR for that: https://github.com/cloudbase/coriolis/pull/451

PR #449, opened 2026-06-03
Move common utils to Coriolis core

petrutlucian94:
There are a few utility functions that are shared by most providers:

* initializing the backup writers and corresponding constants
* polling for an instance to become reachable using SSH or WinRM

That being considered, we'll move them to Coriolis core, aiming to reduce provider code duplication.


coriolis/utils.py
```diff
@@ -1118,3 +1118,97 @@ def start_thread(target, args=(), kwargs=None, daemon=True):
...
+def _poll_instance_until_reachable_winrm(
+    connection_info: dict,
+    timeout: int = 600,
+    poll_interval: int = 10,
+):
+    start = time.time()
+    while (time.time() - start) < timeout:
+        try:
+            # Avoid a circular dependency.
+            from coriolis import wsman
```
Dany9966: To avoid this kind of stuff, maybe you could move these polling functions into `providers.provider_utils`?

petrutlucian94: Sounds good, I'll move it there.

petrutlucian94:
The comment was addressed, I'm merging the PR.

PR #450, opened 2026-06-03
Unlock Bitlocker encrypted volumes

petrutlucian94:
Just like in case of LUKS (https://github.com/cloudbase/coriolis/pull/436), we'll let Coriolis users specify a BitLocker recovery password.

At the very least it should unlock the OS volume, however it may be used for other encrypted volumes as well. If no encrypted volume could be unlocked using the specified key, Coriolis will error out.

In addition to that, we'll temporarily suspend BitLocker on the specified volumes. It won't decrypt the volumes, it merely adds a publicly accessible protector that allows the replica instance to boot.

Once the replica instance boots, BitLocker will be resumed automatically and the TPM protector will be reconfigured.

While at it, we'll use `oslo_utils.strutils.mask_password` and `oslo_utils.strutils.mask_dict_password` in various places to avoid logging sensitive information, such as the Bitlocker recovery key.

### DIFF
--- a/coriolis/osmorphing/osmount/windows.py
+++ b/coriolis/osmorphing/osmount/windows.py
@@ -242,20 +242,6 @@ class WindowsMountTools(base.BaseOSMountTools):
             f'manage-bde -unlock "{volume_id}" '
             f'-RecoveryPassword "{recovery_password}"')
 
-    def _suspend_bitlocker(self, volume_id: str):
-        """Suspend BitLocker until the next reboot for a given volume.
-
-        It doesn't decrypt the device, it just adds a publicly accessible
-        BitLocker protector that automatically unlocks the volume.
-
-        When the replica instance boots, the TPM protector will be reconfigured
-        automatically. Unfortunately the '-RebootCount' parameter isn't
-        honored, perhaps due to the fact that the disks are attached to a
-        separate VM. For this reason, we'll use a first-boot script to resume
-        BitLocker explicitly.
-        """
-        self._conn.exec_ps_command(f'Suspend-BitLocker "{volume_id}"')
-
     def _unlock_encrypted_volumes(self):
         recovery_password = self._osmorphing_info.get(
             constants.ENCRYPTED_DISKS_PASS)
@@ -285,18 +271,29 @@ class WindowsMountTools(base.BaseOSMountTools):
                     "recovery password.",
                     encrypted_volume_id)
                 continue
-
-            # Suspend BitLocker until the replica boots.
-            #
-            # We'll intentionally propagate the failure if we managed to
-            # unlock the volume but failed to suspend BitLocker.
-            self._suspend_bitlocker(encrypted_volume_id)
             self._unlocked_volumes.append(encrypted_volume_id)
 
         if not unlocked:
             raise exception.CoriolisException(
                 "Could not unlock any volume using the specified "
                 "BitLocker recovery password.")
+
+    def _suspend_bitlocker(self, volume_id: str):
+        """Suspend BitLocker until the next reboot for a given volume.
+
+        It doesn't decrypt the device, it just adds a publicly accessible
+        BitLocker protector that automatically unlocks the volume.
+
+        When the replica instance boots, the TPM protector will be reconfigured
+        automatically. Unfortunately the '-RebootCount' parameter isn't
+        honored, perhaps due to the fact that the disks are attached to a
+        separate VM. For this reason, we'll use a first-boot script to resume
+        BitLocker explicitly.
+        """
+        self._conn.exec_ps_command(f'Suspend-BitLocker "{volume_id}"')
+
+    def _resume_bitlocker(self, volume_id: str):
+        self._conn.exec_ps_command(f'Resume-BitLocker "{volume_id}"')
 
     def install_encryption_firstboot_setup(
         self,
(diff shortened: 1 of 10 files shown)


coriolis/osmorphing/osmount/windows.py
```diff
@@ -223,9 +231,105 @@ def _set_volumes_drive_letter(self):
...
+    def _get_encrypted_volume_ids(self):
+        out = self._conn.exec_ps_command(
+            'gwmi -ns "Root\\CIMV2\\Security\\MicrosoftVolumeEncryption" '
+            '-class Win32_EncryptableVolume | % {$_.DeviceID}')
+        return [x for x in out.replace("\r\n", "\n").split("\n") if x]
+
+    def _unlock_encrypted_volume(self, volume_id: str, recovery_password: str):
+        self._conn.exec_ps_command(
+            f'manage-bde -unlock "{volume_id}" '
+            f'-RecoveryPassword "{recovery_password}"')
```
claudiubelu, line 243: we don't log commands in `exec_ps_command` like we do for SSH, right? I wouldn't want to log sensitive information.

coriolis/osmorphing/osmount/windows.py
```diff
@@ -223,9 +231,105 @@ def _set_volumes_drive_letter(self):
...
+        BitLocker explicitly.
+        """
+        self._conn.exec_ps_command(f'Suspend-BitLocker "{volume_id}"')
+
+    def _unlock_encrypted_volumes(self):
+        recovery_password = self._osmorphing_info.get(
+            constants.ENCRYPTED_DISKS_PASS)
+        if not recovery_password:
+            LOG.info("No encrypted disk password specified, "
+                     "skipping BitLocker unlock.")
```
claudiubelu, line 250: For LUKS, we also checked whether the disks are encrypted or not. If they are, and are TPM-bound, they won't unlock on replica start, especially if the root disk is encrypted. IMO, we should raise an exception if the disk is encrypted and no recovery password is given.

coriolis/osmorphing/osmount/windows.py
```diff
@@ -223,9 +231,105 @@ def _set_volumes_drive_letter(self):
...
+
+        When the replica instance boots, the TPM protector will be reconfigured
+        automatically. Unfortunately the '-RebootCount' parameter isn't
+        honored, perhaps due to the fact that the disks are attached to a
+        separate VM. For this reason, we'll use a first-boot script to resume
+        BitLocker explicitly.
+        """
+        self._conn.exec_ps_command(f'Suspend-BitLocker "{volume_id}"')
+
+    def _unlock_encrypted_volumes(self):
```
claudiubelu, line 245: minor nit: `_unlock_encrypted_volumes` should probably be near `_unlock_encrypted_volume`.

coriolis/osmorphing/osmount/windows.py
```diff
@@ -223,9 +231,105 @@ def _set_volumes_drive_letter(self):
...
+                LOG.info(
+                    "Successfully unlocked BitLocker encrypted volume: %s",
+                    encrypted_volume_id)
+                unlocked = True
+            except Exception:
+                LOG.info(
+                    "Could not unlock volume %s using the specified "
+                    "recovery password.",
+                    encrypted_volume_id)
+                continue
```
claudiubelu, line 273: we are talking about encrypted data here. Shouldn't we panic / raise a bit more if this is the case? If this is the case (this exception occured for the additional disks), and we proceed with the OS morphing and replica start, we'll see the VM start and consider that a success, but the other disks are still locked. If the disks are TPM-locked, can they still be recovered / unlocked, if a recovery password was not set up beforehand?

coriolis/osmorphing/osmount/windows.py
```diff
@@ -223,9 +231,105 @@ def _set_volumes_drive_letter(self):
...
+
+            # Suspend BitLocker until the replica boots.
+            #
+            # We'll intentionally propagate the failure if we managed to
+            # unlock the volume but failed to suspend BitLocker.
+            self._suspend_bitlocker(encrypted_volume_id)
+            self._unlocked_volumes.append(encrypted_volume_id)
```
claudiubelu, line 274: I'd move this above `self._suspend_bitlocker`. The current volume is still unlocked, whether bitlocker is suspended or not, and it should still be locked back on cleanup.

Do we lock the volumes back up in case of failure though?

coriolis/osmorphing/osmount/windows.py
petrutlucian94: It's using `debug` logging. We had a discussion about this and the consensus was debug logging may leak sensitive information. Release environments should not use debug logging.

Removing it altogether would make Coriolis difficult to debug.

There's also the option of log sanitization and while it could be helpful, it's difficult to guarantee that no sensitive data will be logged.

petrutlucian94: It won't help much since target environment options would get logged anyway, but I'll sanitize the output of `exec_ps_command`.

coriolis/osmorphing/osmount/windows.py
petrutlucian94: The user may customize the unlock process through os-morphing user scripts and omit the recovery password here.

coriolis/osmorphing/osmount/windows.py
petrutlucian94: Not really, we mainly care about the OS drive. In most cases, Windows group policies won't even allow re-using the same password for multiple disks.

coriolis/osmorphing/osmount/windows.py
petrutlucian94: We don't need to lock the volumes explicitly, the minion VM will be cleaned up immediately afterwards. So will the replica disks if cloning is enabled.

petrutlucian94: On the other hand, we need need to ensure that BitLocker is resumed in case of failures, I've moved the `_suspend_bitlocker`  call to `install_encryption_firstboot_setup` and updated it to resume BitLocker in case of failures.

coriolis/osmorphing/osmount/windows.py
petrutlucian94: I've started using `oslo_utils.strutils.mask_password` and `oslo_utils.strutils.mask_dict_password`, covering the exec commands, task info dicts and osmorphing info dicts.

coriolis/osmorphing/osmount/windows.py
```diff
@@ -223,9 +231,120 @@ def _set_volumes_drive_letter(self):
...
+            LOG.exception("First-boot preparation failed, attempting to "
+                          "resume BitLocker.")
+            try:
+                for encrypted_volume_id in self._unlocked_volumes:
+                    self._resume_bitlocker(encrypted_volume_id)
+            except Exception:
+                LOG.exception(
+                    "Unable to resume BitLocker for volume: %s" %
+                    encrypted_volume_id)
+            raise
```
claudiubelu, line 342: Hmm, this basically means that we raise on the first `_resume_bitlocker` exception, leaving the rest suspended.

petrutlucian94: Whops, my bad. Fixed, also added a test to cover this.

petrutlucian94:
The comments have been addressed, I'm merging the PR.

PR #451, opened 2026-06-04
Reject empty user scripts

petrutlucian94:
Coriolis allows removing user scripts by specifying "None" instead of a user script.

As mentioned here [1], an empty string is most likely an user error though and we should reject the request.

[1] https://github.com/cloudbase/coriolis/pull/448#discussion_r3350559746


petrutlucian94:
Trivial PR, no replies in one week. I'll go ahead and merge it.

PR #452, opened 2026-06-04
Add suse network config support

mihaelabalutoiu:
This PR adds:

- Complete SUSE network configuration support during OS morphing, fixing post-migration network connectivity issues (notably on `SLES 16`).
- Implementations of `set_net_config` and `disable_predictable_nic_names` for SUSE, with support for both wicked-based (SLES/openSUSE 12/15) and `NetworkManager-based` (SLES 16) systems.
- A refactor that moves common `ifcfg/NetworkManager` configuration logic into the shared base class, reducing code duplication and ensuring existing network profiles are backed up before new ones are written.

### DIFF
--- a/coriolis/osmorphing/base.py
+++ b/coriolis/osmorphing/base.py
@@ -18,6 +18,39 @@ from coriolis import utils
 GRUB2_SERIAL = "serial --word=8 --stop=1 --speed=%d --parity=%s --unit=0"
 LOG = logging.getLogger(__name__)
 
+IFCFG_TEMPLATE = """
+TYPE=Ethernet
+BOOTPROTO=dhcp
+DEFROUTE=yes
+IPV4_FAILURE_FATAL=no
+IPV6INIT=yes
+IPV6_AUTOCONF=yes
+IPV6_DEFROUTE=yes
+IPV6_FAILURE_FATAL=no
+NAME=%(device_name)s
+DEVICE=%(device_name)s
+ONBOOT=yes
+NM_CONTROLLED=%(nm_controlled)s
+"""
+
+NMCONNECTION_TEMPLATE = """[connection]
+id=%(device_name)s
+uuid=%(connection_uuid)s
+type=ethernet
+interface-name=%(device_name)s
+autoconnect=true
+
+[ethernet]
+
+[ipv4]
+method=auto
+may-fail=false
+
+[ipv6]
+method=auto
+addr-gen-mode=default
+"""
+
 
 # Required OS release fields which are expected from the OSDetect tools.
 # 'schemas.CORIOLIS_DETECTED_OS_MORPHING_INFO_SCHEMA' schema:
@@ -222,6 +255,12 @@ class BaseLinuxOSMorphingTools(BaseOSMorphingTools):
     _packages = {}
     _NETWORK_SCRIPTS_PATH = "etc/sysconfig/network-scripts"
     _NM_CONNECTIONS_PATH = "etc/NetworkManager/system-connections"
+    # ifcfg profile template written for DHCP NICs. Subclasses may override it
+    # when the target uses a different ifcfg format (e.g. wicked-based SUSE).
+    _IFCFG_TEMPLATE = IFCFG_TEMPLATE
+    # Minimum OS major version at which ifcfg profiles should be marked as
+    # NetworkManager-controlled (NM_CONTROLLED=yes).
+    _IFCFG_NM_CONTROLLED_MIN_VERSION = None
 
     def __init__(self, conn, os_root_dir, os_root_dev, hypervisor,
                  event_manager, detected_os_info, osmorphing_parameters,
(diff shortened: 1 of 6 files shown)


coriolis/osmorphing/suse.py
```diff
@@ -11,8 +11,12 @@
 from coriolis import exception
 from coriolis.osmorphing import base
 from coriolis.osmorphing.osdetect import suse as suse_detect
+from coriolis.osmorphing import redhat as redhat_osmorphing
 from coriolis import utils
 
+IFCFG_TEMPLATE = redhat_osmorphing.IFCFG_TEMPLATE
+NMCONNECTION_TEMPLATE = redhat_osmorphing.NMCONNECTION_TEMPLATE
```
Dany9966: I think you should move these templates into base if you need them here as well.

coriolis/osmorphing/suse.py
```diff
@@ -29,6 +33,8 @@
 
 class BaseSUSEMorphingTools(base.BaseLinuxOSMorphingTools):
 
+    _NETWORK_SCRIPTS_PATH = "etc/sysconfig/network-scripts"
+    _NM_CONNECTIONS_PATH = "etc/NetworkManager/system-connections"
```
Dany9966: These were moved into `base`. You can remove these.

coriolis/osmorphing/suse.py
```diff
@@ -61,12 +67,107 @@ def check_os_supported(cls, detected_os_info):
...
+
+    def _write_dhcp_net_config(self, nics_info):
+        self.disable_predictable_nic_names()
+        ethernet_keyfiles = self._get_ethernet_keyfiles()
+        if ethernet_keyfiles:
+            self._write_nmconnection_configs(nics_info)
+        else:
+            self._write_nic_configs(nics_info)
 
     def set_net_config(self, nics_info, dhcp):
```
Dany9966, line 96: Please rechek how this was implemented in redhat, and re-use that functionality. Most of the methods above this were moved into base, so there's no need to be here as well.

Dany9966 (changes requested):

coriolis/osmorphing/suse.py
mihaelabalutoiu: Done

coriolis/osmorphing/suse.py
mihaelabalutoiu: Done, I moved them.

coriolis/osmorphing/suse.py
mihaelabalutoiu: Done

coriolis/osmorphing/suse.py
```diff
@@ -61,12 +61,102 @@ def check_os_supported(cls, detected_os_info):
...
+            {"opt_type": "key_val", "opt_key": "net.ifnames", "opt_val": 0})
+        cfg.append_to_option(
+            "GRUB_CMDLINE_LINUX",
+            {"opt_type": "key_val", "opt_key": "biosdevname", "opt_val": 0})
+        self._write_file_sudo("etc/default/grub", cfg.dump())
+        self._execute_update_grub()
+
+    def _get_ifcfg_nm_controlled(self):
+        if self._version_supported_util(self._version, minimum=15):
+            return "yes"
```
Dany9966: Did you check on suse machines 12 and 15? Does only 15 need nm_controlled to be enabled?

coriolis/osmorphing/suse.py
```diff
@@ -61,12 +61,102 @@ def check_os_supported(cls, detected_os_info):
...
+        return "no"
+
+    def _backup_nmconnection_files(self, backup_file_suffix=".bak"):
```
Dany9966: The `redhat._backup_nm_connection_files` should be refactored into `base` and re-used instead of overriding. As a matter of fact, any method that can be re-used from redhat should be refactored into base and used here. There's no need to override everything here, unless there's something that really is different.

IMO the only methods added in this PR that need to be implemented here are `disable_predictable_nic_names`, and `set_net_config`. The rest should come from base.

coriolis/osmorphing/suse.py
```diff
@@ -61,12 +61,102 @@ def check_os_supported(cls, detected_os_info):
...
+            "GRUB_CMDLINE_LINUX_DEFAULT",
+            {"opt_type": "key_val", "opt_key": "biosdevname", "opt_val": 0})
+        cfg.append_to_option(
+            "GRUB_CMDLINE_LINUX",
+            {"opt_type": "key_val", "opt_key": "net.ifnames", "opt_val": 0})
+        cfg.append_to_option(
+            "GRUB_CMDLINE_LINUX",
+            {"opt_type": "key_val", "opt_key": "biosdevname", "opt_val": 0})
+        self._write_file_sudo("etc/default/grub", cfg.dump())
+        self._execute_update_grub()
```
Dany9966: Please replace this with `self._schedule_grub2_update()`.

Dany9966 (changes requested):

coriolis/osmorphing/base.py
```diff
@@ -467,6 +503,75 @@ def _get_keyfiles_by_type(self, nmconnection_type, network_scripts_path):
...
+                                   backup_file_suffix=".bak"):
+        if nmconnection_files is None:
+            nmconnection_files = (
+                self._get_existing_ethernet_nmconnection_files())
+        for cfg_path in nmconnection_files:
+            self._exec_cmd_chroot(
+                'mv "%s" "%s%s"' % (cfg_path, cfg_path, backup_file_suffix))
+            LOG.debug("Backed up nmconnection profile '%s'", cfg_path)
+
+    def _backup_all_ifcfg_configs(self, backup_file_suffix=".bak"):
```
Dany9966: Can we please rename this into `_backup_ethernet_ifcfg_configs`?

mihaelabalutoiu:
I made the changes, please take a look now. Thank you!

coriolis/osmorphing/suse.py
```diff
@@ -61,12 +68,55 @@ def check_os_supported(cls, detected_os_info):
...
+        cfg.append_to_option(
+            "GRUB_CMDLINE_LINUX",
+            {"opt_type": "key_val", "opt_key": "biosdevname", "opt_val": 0})
+        self._write_file_sudo("etc/default/grub", cfg.dump())
+        self._schedule_grub2_update()
+
+    def _write_nic_configs(self, nics_info):
+        # NOTE: overrides the base implementation because
+        # wicked-based SUSE uses a different ifcfg location and format.
+        self._backup_all_ifcfg_configs()
```
Dany9966: Since this looks for ethernet type ifcfgs to backup mainly, and SUSE machines don't identify types in ifcfg files, there's a risk here that we may back up files that are of other types, which we shouldn't. In this case, I think we should not call this method, and instead only back up files that would conflict with what Coriolis creates. 
For example, if we want to create `ifcfg-eth0` and that already exists, first back it up, then create it.

Dany9966 (changes requested):

coriolis/osmorphing/suse.py
mihaelabalutoiu: Done

coriolis/osmorphing/base.py
mihaelabalutoiu: Done

coriolis/osmorphing/suse.py
```diff
@@ -61,12 +68,57 @@ def check_os_supported(cls, detected_os_info):
...
+        cfg.append_to_option(
+            "GRUB_CMDLINE_LINUX",
+            {"opt_type": "key_val", "opt_key": "biosdevname", "opt_val": 0})
+        self._write_file_sudo("etc/default/grub", cfg.dump())
+        self._schedule_grub2_update()
+
+    def _write_nic_configs(self, nics_info):
+        # NOTE: overrides the base implementation because
+        # wicked-based SUSE uses a different ifcfg location and format.
+        network_cfg_file = "etc/sysconfig/network"
```
petrutlucian94: Shouldn't we back up the original script files like the base implementation does?

Also, I think we can avoid duplicating it by using:
* `_NETWORK_SCRIPTS_PATH` for the config file location -> already defined by the base class, could be overridden by the Suse provider
* another class attribute for the config template

fabi200123:
LGTM

coriolis/osmorphing/base.py
```diff
@@ -467,6 +503,75 @@ def _get_keyfiles_by_type(self, nmconnection_type, network_scripts_path):
...
+            return
+        for cfg_path, _ in self._get_ifcfgs_by_type(
+                "Ethernet", self._NETWORK_SCRIPTS_PATH):
+            if os.path.basename(cfg_path) == "ifcfg-lo":
+                continue
+            self._exec_cmd_chroot(
+                'mv "%s" "%s%s"' % (cfg_path, cfg_path, backup_file_suffix))
+            LOG.debug("Backed up ifcfg profile '%s'", cfg_path)
+
+    def _write_nic_configs(self, nics_info):
```
petrutlucian94, line 543: I see that the `nics_info` format is entirely provider specific, all we need is the number of nics:

https://github.com/cloudbase/coriolis/blob/7ea1e243ce953170615cca8b6c54c286f4fec8dd/coriolis/schemas/os_morphing_resources_schema.json#L38-L43

I guess it's reasonable since providers can override this and handle provider specific nic info.

petrutlucian94:
Oh, and please add a PR description.

mihaelabalutoiu:
> Oh, and please add a PR description.

Done, I updated the code, please take a look now and sorry about that.

petrutlucian94 (approved):

PR #453, opened 2026-06-08
Move duplicate os-morphing code to core

petrutlucian94:
There's a significant amount of os-morphing code duplication among Coriolis providers.

This PR moves the following to Coriolis core:

* Windows virtio installation
* Wheezy backports on Debian 7
* UEFI fallback bootloader for Debian
* initramfs update helper for Ubuntu

### DIFF
--- a/coriolis/osmorphing/debian.py
+++ b/coriolis/osmorphing/debian.py
@@ -162,6 +162,25 @@ class BaseDebianMorphingTools(base.BaseLinuxOSMorphingTools):
             cfg_name = "%s/coriolis_netplan.yaml" % self.netplan_base
             self._write_file_sudo(cfg_name, new_cfg)
 
+    def _run_update_initramfs(self):
+        # env LC_ALL=C suppresses Perl/shell locale warnings that would
+        # otherwise appear in stdout and get mixed into the version list.
+        # Using 'env' rather than a shell assignment prefix because
+        # _exec_cmd_chroot runs the command directly via chroot (no shell),
+        # so "LC_ALL=C cmd" would be treated as the binary name.
+        raw = self._exec_cmd_chroot("env LC_ALL=C linux-version list")
+
+        kernel_versions = [v for v in raw.splitlines() if v and v[0].isdigit()]
+        if not kernel_versions:
+            LOG.warning(
+                "No kernel versions found via 'linux-version list'; "
+                "skipping update-initramfs. Raw output was: %r",
+                raw,
+            )
+            return
+        for version in kernel_versions:
+            self._exec_cmd_chroot(f"update-initramfs -k {version} -u")
+
     def get_installed_packages(self):
         cmd = "dpkg-query -f '${binary:Package}\\n' -W"
         try:
@@ -196,6 +215,7 @@ class BaseDebianMorphingTools(base.BaseLinuxOSMorphingTools):
 
     def post_packages_install(self, package_names):
         self._configure_cloud_init()
+        self._run_update_initramfs()
         super(BaseDebianMorphingTools, self).post_packages_install(
             package_names)
 
--- a/coriolis/osmorphing/debian.py
+++ b/coriolis/osmorphing/debian.py
@@ -169,7 +169,15 @@ class BaseDebianMorphingTools(base.BaseLinuxOSMorphingTools):
         # Using 'env' rather than a shell assignment prefix because
         # _exec_cmd_chroot runs the command directly via chroot (no shell),
         # so "LC_ALL=C cmd" would be treated as the binary name.
-        raw = self._exec_cmd_chroot("env LC_ALL=C linux-version list")
+        try:
+            raw = self._exec_cmd_chroot("env LC_ALL=C linux-version list")
+        except exception.SSHCommandNotFoundException as ex:
+            # Container images such as the ones used by the integration tests
+            # do not include the "linux-version" binary.
+            LOG.warning(
+                "Unable to enumerate kernel versions, skipping "
+                "update-initramfs. Exception: %s", ex)
+            return
 
         kernel_versions = [v for v in raw.splitlines() if v and v[0].isdigit()]
         if not kernel_versions:
--- a/coriolis/osmorphing/debian.py
+++ b/coriolis/osmorphing/debian.py
@@ -2,6 +2,8 @@
 # All Rights Reserved.
 
 from io import StringIO
+import os
+import re
 
 from oslo_log import log as logging
 import yaml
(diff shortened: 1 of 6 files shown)


coriolis/tests/osmorphing/test_debian.py
```diff
@@ -312,3 +331,130 @@ def test_uninstall_packages_with_exception(self, mock_exec_cmd_chroot):
...
+            "",
+            # update-grub,
+        ]
+        mock_test_path_chroot.return_value = True
+
+        self.morpher._install_uefi_fallback_bootloader()
+
+        mock_exec_cmd_chroot.assert_has_calls([
+            mock.call("uname -m"),
+        ])
```
claudiubelu, line 504: according to : ``There can be extra calls before or after the specified calls.``. Because of this, there's technically no difference between this test and ``test_install_uefi_fallback_bootloader``. We could add an exception to the `mock_exec_cmd_chroot.side_effect` above: if too many calls are being done to `_exec_cmd_chroot`, an exception would be raised.

[1] [There can be extra calls before or after the specified calls.](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_has_calls)

coriolis/tests/osmorphing/test_ubuntu.py
```diff
@@ -170,3 +170,34 @@ def test__set_netplan_ethernet_configs_no_ethernet_configs(
                 self.os_root_dir, self.os_root_dir))
         mock_write_file_sudo.assert_called_once_with(
             'etc/netplan/file1.yaml', ubuntu.yaml.dump(config_data))
+
+    @mock.patch.object(base.BaseLinuxOSMorphingTools, '_exec_cmd_chroot')
+    def test_update_ramfs(self, mock_exec_cmd_chroot):
```
claudiubelu: nit: `test_update_initramfs`? ditto to the second test below too.

coriolis/tests/osmorphing/test_windows.py
```diff
@@ -989,3 +998,130 @@ def test_register_firstboot_script_explicit_fname(
...
+        self.morphing_tools._conn.download_file.assert_called_once_with(
+            self._FAKE_VIRTIO_ISO_URL, exp_iso_path)
+        mock_mount_disk_image.assert_called_once_with(exp_iso_path)
+        mock_dismount.assert_called_once_with(exp_iso_path)
+
+        exp_repo_path = "C:\\Windows\\System32\\DriverStore\\FileRepository"
+        mock_grant_permissions.assert_called_once_with(
+            exp_repo_path, f"*{null_sid}")
+        mock_revoke_permissions.assert_called_once_with(
+            exp_repo_path, f"*{null_sid}")
```
claudiubelu, line 1125: also assert `_dismount_disk_image` is called?

also for completion's sake, could we also add a test for the `if not virtio_dir:` scenario?

coriolis/tests/osmorphing/test_debian.py
petrutlucian94: Good point.

coriolis/osmorphing/debian.py
```diff
@@ -97,6 +99,42 @@ def _compose_netplan_cfg(self, nics_info):
...
+        regex = b"^" + backports
+
+        repo_found = re.search(regex, sources, flags=re.MULTILINE)
+        if not repo_found:
+            self._event_manager.progress_update("Adding wheezy backports")
+            sources_updated = sources + b"\n" + backports + b"\n"
+            self._write_file_sudo("etc/apt/sources.list", sources_updated)
+            self._exec_cmd_chroot("apt-get update -y")
+
+    def _install_uefi_fallback_bootloader(self):
```
Dany9966, line 118: I think  this method should be called in post_packages_install. There's a bug amongst providers where the firmware of the migrated VM is not properly being checked, therefore this method gets called anyway (even if using BIOS).
Adding this logic in core will solve the problem.
if self._osmorphing_parameters.get('firmware_type') == constants.FIRMWARE_TYPE_EFI:
     self._install_uefi_fallback_bootloader()
I guess we can add that in a separate set of PRs where this also gets removed from the providers.

coriolis/tests/osmorphing/test_windows.py
petrutlucian94: We do assert that the disk is unmounted, it's just that the dismount assertion is next to the "mount" assertion.

Image

Good point about the missing virtio dir scenario, I'll add another test.

coriolis/osmorphing/debian.py
```diff
@@ -97,6 +99,42 @@ def _compose_netplan_cfg(self, nics_info):
...
+        if self._test_path_chroot(os.path.join(bootloader_dir,
+                                  arch_map[arch])):
+            LOG.info("Fallback bootloader exists. Skipping")
+            return
+
+        self._exec_cmd_chroot(
+            f"grub-install --removable --target={arch}-efi "
+            "--efi-directory=/boot/efi --uefi-secure-boot"
+        )
+        self._exec_cmd_chroot("update-grub")
```
Dany9966, line 137: Also for the future, I'd like to see some kind of logic of post-hooks for stuff like `update-grub` or `update-initramfs` (or similar), so we don't do that multiple times in the same OS_MORPHING process, as some of those could take some time. Doing it multiple times will just make it worse.

coriolis/osmorphing/ubuntu.py
```diff
@@ -122,3 +122,22 @@ def _set_netplan_ethernet_configs(
                 "Writing following configuration to '%s': %s" % (
                     config_path_chroot, config_data))
             self._write_file_sudo(config_path_chroot, yaml.dump(config_data))
+
+    def _run_update_initramfs(self):
```
Dany9966: I think this should go in the `debian` class instead. The `ubuntu` one will inherit it. It may be needed in debian as well. This is also something we could run in `post_packages_install` by default.

coriolis/osmorphing/windows.py
```diff
@@ -705,6 +706,114 @@ def set_net_config(self, nics_info, dhcp):
...
+            "vioscsi",
+            "vioserial",
+            "viostor",
+            "viogpudo",
+        ]
+
+        self._event_manager.progress_update("Downloading virtio-win drivers")
+
+        virtio_iso_path = "c:\\virtio-win.iso"
+        virtio_iso_url = self._osmorphing_parameters["windows_virtio_iso_url"]
```
Dany9966, line 760: This key value `windows_virtio_iso_url` needs to be checked amongst all providers that will use it, make sure that it's being used correctly (i.e. they don't use a different key name for some stupid reason).

claudiubelu (approved):

coriolis/osmorphing/debian.py
petrutlucian94: It's idempotent, so we can safely add the call on the core side right now.

coriolis/osmorphing/debian.py
petrutlucian94: Sounds reasonable, it can be done in a follow-up PR.

coriolis/osmorphing/windows.py
petrutlucian94: Well, I guess there will be a set of patches that removes this duplicate method from each individual provider. Before merging those PRs, we'll need to ensure that this field is set.

PR #455, opened 2026-06-18
Fix issue of trying to decode a str

MiningPickaxe:
This fixes an error where _parse_iface_file tries to decode a str. The output of the command getting executed is already getting decoded in coriolis/utils.py exec_cmd_ssh introduced in commit 539fabb9d302b3915cda19357ba0d0ea6d33d25f


petrutlucian94:
Thanks for sending the fix! It seems like this was indeed an oversight in https://github.com/cloudbase/coriolis/commit/539fabb9d302b3915cda19357ba0d0ea6d33d25f.

There's an unit test that will need to be updated before being able to merge the fix but I'll take care of that.

PR #456, opened 2026-06-18
Optimize transfer execution lookups

mihaelabalutoiu:
Avoid loading full transfer execution histories and enable efficient DB-side filtering for execution lookups.

### DIFF
--- a/coriolis/conductor/rpc/server.py
+++ b/coriolis/conductor/rpc/server.py
@@ -1416,34 +1416,43 @@ class ConductorServerEndpoint(object):
                     "Transfer '%s' is currently being deployed" % transfer_id)
 
     @staticmethod
-    def _check_running_executions(action):
-        running_executions = [
-            e.id for e in action.executions
-            if e.status in constants.ACTIVE_EXECUTION_STATUSES]
+    def _check_running_executions(ctxt, transfer):
+        running_executions = db_api.get_transfer_tasks_executions(
+            ctxt, transfer.id,
+            filters={"status": constants.ACTIVE_EXECUTION_STATUSES})
         if running_executions:
             raise exception.InvalidActionTasksExecutionState(
                 "Another tasks execution is in progress: %s" % (
-                    running_executions))
+                    [e.id for e in running_executions]))
 
     def _check_transfer_running_executions(self, ctxt, transfer):
-        self._check_running_executions(transfer)
+        self._check_running_executions(ctxt, transfer)
         self._check_running_transfer_deployments(ctxt, transfer.id)
 
     @staticmethod
-    def _check_valid_transfer_tasks_execution(transfer, force=False):
-        sorted_executions = sorted(
-            transfer.executions, key=lambda e: e.number, reverse=True)
-        if not sorted_executions and not force:
-            raise exception.InvalidTransferState(
-                "The Transfer has never been executed.")
+    def _get_last_transfer_execution(ctxt, transfer_id, status=None,
+                                     execution_type=None):
+        filters = {}
+        if status is not None:
+            filters["status"] = status
+        if execution_type is not None:
+            filters["type"] = execution_type
+        executions = db_api.get_transfer_tasks_executions(
+            ctxt, transfer_id, filters=filters,
+            sort_keys=["number"], sort_dirs=["desc"], limit=1)
+        return executions[0] if executions else None
 
-        if not [e for e in sorted_executions
-                if e.type == constants.EXECUTION_TYPE_TRANSFER_EXECUTION and (
-                    e.status == constants.EXECUTION_STATUS_COMPLETED)]:
-            if not force:
-                raise exception.InvalidTransferState(
-                    "A transfer must have been executed successfully at least "
-                    "once in order to be deployed")
+    def _check_transfer_deploy_ability(self, ctxt, transfer, force=False):
+        if force:
+            return
+
+        completed_execution = self._get_last_transfer_execution(
+            ctxt, transfer.id, status=constants.EXECUTION_STATUS_COMPLETED,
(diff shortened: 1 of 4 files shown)


coriolis/conductor/rpc/server.py
```diff
@@ -1416,34 +1416,48 @@ def _check_running_transfer_deployments(self, ctxt, transfer_id):
...
+            return
+
+        completed_execution = self._get_last_completed_transfer_execution(
+            ctxt, transfer.id)
+        if completed_execution is not None:
+            return
+
+        any_execution = db_api.get_transfer_tasks_executions(
+            ctxt, transfer.id, limit=1)
+        if not any_execution:
```
Dany9966: This check can be removed. If no `completed_execution` is found, we can make use of the last execution message below. There's also no point to make almost the same DB call twice.

coriolis/conductor/rpc/server.py
```diff
@@ -1416,34 +1416,47 @@ def _check_running_transfer_deployments(self, ctxt, transfer_id):
...
-        self._check_running_executions(transfer)
+        self._check_running_executions(ctxt, transfer)
         self._check_running_transfer_deployments(ctxt, transfer.id)
 
-    @staticmethod
-    def _check_valid_transfer_tasks_execution(transfer, force=False):
-        sorted_executions = sorted(
-            transfer.executions, key=lambda e: e.number, reverse=True)
-        if not sorted_executions and not force:
+    def _get_last_completed_transfer_execution(self, ctxt, transfer_id):
```
Dany9966: This can be a `staticmethod`

coriolis/conductor/rpc/server.py
```diff
@@ -1416,34 +1416,47 @@ def _check_running_transfer_deployments(self, ctxt, transfer_id):
...
+        executions = db_api.get_transfer_tasks_executions(
+            ctxt, transfer_id,
+            filters={
+                "status": constants.EXECUTION_STATUS_COMPLETED,
+                "type": constants.EXECUTION_TYPE_TRANSFER_EXECUTION,
+            },
+            sort_keys=["number"], sort_dirs=["desc"], limit=1)
+        return executions[0] if executions else None
+
+    def _check_valid_transfer_tasks_execution(self, ctxt, transfer,
```
Dany9966: This method name is also not descriptive enough to what it actually does.
Can you please rename it into something like `_check_transfer_deployability`.

coriolis/conductor/rpc/server.py
```diff
@@ -1875,16 +1888,14 @@ def _update_task_info_for_minion_allocations(
                 ctxt, action.id, instance, action.info[instance])
 
     def _get_last_execution_for_transfer(self, ctxt, transfer, requery=False):
```
Dany9966: This method and `_get_last_completed_transfer_execution` are almost the same. I think you could merge these into one.

Dany9966 (changes requested):

mihaelabalutoiu:
Done, thank you for that!

coriolis/conductor/rpc/server.py
```diff
@@ -1416,34 +1416,40 @@ def _check_running_transfer_deployments(self, ctxt, transfer_id):
...
-        sorted_executions = sorted(
-            transfer.executions, key=lambda e: e.number, reverse=True)
-        if not sorted_executions and not force:
-            raise exception.InvalidTransferState(
-                "The Transfer has never been executed.")
+    def _get_last_transfer_execution(ctxt, transfer_id, status=None):
+        filters = {}
+        if status is not None:
+            filters["status"] = status
+            filters["type"] = constants.EXECUTION_TYPE_TRANSFER_EXECUTION
```
Dany9966: `type` filter should be parametrized separately imo

Dany9966 (changes requested):

coriolis/tests/conductor/rpc/test_server.py
```diff
@@ -2266,62 +2265,88 @@ def test_check_transfer_running_executions(
...
+        self.assertEqual(execution, result)
+        mock_get_executions.assert_called_once_with(
+            mock.sentinel.context, mock.sentinel.transfer_id,
+            filters={
+                "status": constants.EXECUTION_STATUS_COMPLETED,
+                "type": constants.EXECUTION_TYPE_TRANSFER_EXECUTION,
+            },
+            sort_keys=["number"], sort_dirs=["desc"], limit=1)
+
+        # with only a status filter (type not constrained)
```
Dany9966, line 2324: Please also add a `type` only filter, no status.

coriolis/tests/conductor/rpc/test_server.py
```diff
@@ -5043,7 +5033,8 @@ def test_update_transfer(
             mock.sentinel.context,
             mock_get_transfer.return_value,
         )
-        mock_check_valid_transfer_tasks_execution.assert_called_once_with(
+        mock_check_transfer_deploy_ability.assert_called_once_with(
```
Dany9966: I think we shouldn't check for transfer deployability on a transfer update operation tbh.
Can you please handle this in a separate PR? This will allow users to update Transfers that are in `UNEXECUTED` status as well.

Dany9966 (changes requested):

coriolis/tests/conductor/rpc/test_server.py
mihaelabalutoiu: Sure, I will open a separate PR for that. Thank you!

coriolis/tests/conductor/rpc/test_server.py
mihaelabalutoiu: Done

Dany9966 (approved):

PR #457, opened 2026-06-19
Fix scheduled transfers not triggering

fabi200123:
The `transfer-cron` service started its cron loop in the `TransferCronServerEndpoint` constructor, which runs in the parent process before `oslo_service` forks its worker. Since oslo_service forks a worker even with `workers=1`, the cron loop ran in the parent while the RPC register/unregister handlers ran in the forked child. Each process kept its own in-memory job registry, so schedules created or updated at runtime were added to the child's registry while the parent's loop kept checking an empty one, and scheduled transfers never started.

This was caused by the migration from eventlet to threading, which introduced the parent/child process split for this.

### DIFF
--- a/coriolis/tests/integration/transfers/test_schedules.py
+++ b/coriolis/tests/integration/transfers/test_schedules.py
@@ -34,6 +34,20 @@ class _TransferScheduleTestBase(base.ReplicaIntegrationTestBase):
 
         return sched
 
+    def _wait_for_scheduled_execution(self, timeout=180):
+        """Poll until the schedule triggers a transfer execution.
+
+        Returns the first execution that appears, or None on timeout.
+        """
+        deadline = time.monotonic() + timeout
+        while time.monotonic() < deadline:
+            executions = self._client.transfer_executions.list(
+                self._transfer.id, limit=1)
+            if executions:
+                return executions[0]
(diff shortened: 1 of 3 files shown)


coriolis/tests/integration/transfers/test_schedules.py
```diff
@@ -34,6 +34,20 @@ def _create_schedule(self, **overrides):
...
 
+    def _wait_for_scheduled_execution(self, timeout=180):
+        """Poll until the schedule triggers a transfer execution.
+
+        Returns the first execution that appears, or None on timeout.
+        """
+        deadline = time.monotonic() + timeout
+        while time.monotonic() < deadline:
+            executions = self._client.transfer_executions.list(
+                self._transfer.id)
```
Dany9966: We should set a `limit` here, just in case we ever create a test with a lot of executions.

Dany9966 (changes requested):

coriolis/tests/integration/transfers/test_schedules.py
fabi200123: Added a `limit=1` (we are using executions[0] afterwards)

Dany9966 (approved):

PR #458, opened 2026-06-22
Allow updating transfers in `UNEXECUTED` status

mihaelabalutoiu:
Remove the `transfer_deploy_ability` check from `update_transfer` so that transfers that have not been executed yet (UNEXECUTED status) can still be updated.


Dany9966 (approved):
Finally, some great removal of code! :D

PR #459, opened 2026-06-23
Add post-hook functionality when finalizing osmorphing

fabi200123:
This PR updates the logic of execute grub update in osmorphing so that this is only ran in the `post_package_install` based on a flag configured via `self._schedule_grub2_update()`. 

Before, any grub related modification is followed by an update grub command (or grub-mkconfig). If there were multiple of them, we would basically run redundant updates and only make the osmorphing process take longer, as some of these processes can take up to minutes (especially when doing update-initramfs or dracut updates).

### DIFF
--- a/coriolis/tests/osmorphing/test_base.py
+++ b/coriolis/tests/osmorphing/test_base.py
@@ -262,10 +262,14 @@ class BaseLinuxOSMorphingToolsTestBase(test_base.CoriolisBaseTestCase):
         self.os_morphing_tools.pre_packages_install(mock.sentinel.package_name)
         mock_copy_resolv_conf.assert_called_once_with()
 
+    @mock.patch.object(
+        base.BaseLinuxOSMorphingTools, '_run_scheduled_grub2_update')
     @mock.patch.object(base.BaseLinuxOSMorphingTools, '_restore_resolv_conf')
-    def test_post_packages_install(self, mock_restore_resolv_conf):
+    def test_post_packages_install(self, mock_restore_resolv_conf,
+                                   mock_run_scheduled_grub2_update):
         self.os_morphing_tools.post_packages_install(
             mock.sentinel.package_name)
+        mock_run_scheduled_grub2_update.assert_called_once_with()
         mock_restore_resolv_conf.assert_called_once_with()
 
     @mock.patch.object(base.BaseLinuxOSMorphingTools, '_copy_resolv_conf')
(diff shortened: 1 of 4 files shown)


coriolis/tests/osmorphing/test_base.py
```diff
@@ -1066,9 +1071,42 @@ def test__execute_update_grub(self, mock_get_update_grub2_command,
...
+        self.os_morphing_tools._grub2_update_scheduled = True
+
+        self.os_morphing_tools._run_scheduled_grub2_update()
+
+        mock_execute_update_grub.assert_called_once_with()
+        # The pending flag must be cleared so subsequent calls are no-ops.
+        self.assertFalse(self.os_morphing_tools._grub2_update_scheduled)
+
+        self.os_morphing_tools._run_scheduled_grub2_update()
+        mock_execute_update_grub.assert_called_once_with()
```
Dany9966: I'm a bit confused. How would `_execute_update_grub` run here if `_grub2_update_scheduled` is False?

coriolis/tests/osmorphing/test_debian.py
```diff
@@ -87,7 +89,9 @@ def test_disable_predictable_nic_names(
         mock_read_file_sudo.assert_called_once_with('etc/default/grub')
         mock_write_file_sudo.assert_called_once_with(
             "etc/default/grub", mock_grub2_cfg_editor.return_value.dump())
-        mock_exec_cmd_chroot.assert_called_once_with("/usr/sbin/update-grub")
+        # The grub update must be deferred (scheduled) rather than run now.
+        mock_schedule_grub2_update.assert_called_once_with()
+        mock_exec_cmd_chroot.assert_not_called()
```
Dany9966: I think you should remove this exec_cmd_chroot mock, as it could mess with future modifications. If not mocked and exec_cmd_chroot does run, the test should fail either way.

Dany9966 (changes requested):

coriolis/tests/osmorphing/test_base.py
fabi200123: It doesnt. The `mock_execute_update_grub.assert_called_once_with()` is doing `called_once_with`, which is expecting this to have run exactly once when the flag was on true.
Added some comments + changed it to ```self.assertEqual(1, mock_execute_update_grub.call_count)```

Dany9966 (approved):

PR #464, opened 2026-07-01
Refactor osdetect for rhel family OSes

fabi200123:
This PR refactors the osdetect for the OSes that are RHEL related. It adds a `redhat_common.py` where the `detect_os_from_os_release` is defined to return the proper info from `/etc/os-release`:
```
    return {
        "os_type": constants.OS_TYPE_LINUX,
        "distribution_name": distribution_name,
        "release_version": version,
        "friendly_release_name": "%s Version %s" % (
            distribution_name, version)}
```

**NOTE**: Due to the use of `/etc/os-release`, which was **_introduced_** starting with OSes like `RedHat version 7` (e.g. RHEL 7, CentOS 7, OL7), this implementation makes the previous `RedHat version 6` OSes to not work with the current logic for detecting OSes.


coriolis/osmorphing/osdetect/common.py
```diff
@@ -0,0 +1,55 @@
...
+OS_RELEASE_ID_MAP = {
+    "rocky": ROCKY_LINUX_DISTRO_IDENTIFIER,
+    "centos": CENTOS_DISTRO_IDENTIFIER,
+    "almalinux": CENTOS_DISTRO_IDENTIFIER,
+    "rhel": RED_HAT_DISTRO_IDENTIFIER,
+    "ol": ORACLE_DISTRO_IDENTIFIER,
+}
+
+
+def detect_os_from_os_release(os_release, *, match_ids):
```
petrutlucian94: Let's use type hints and document the "os_release" parameter, stating that it's a dict populated by /etc/os-release values.

coriolis/osmorphing/osdetect/common.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2026 Cloudbase Solutions Srl
+# All Rights Reserved.
+
+"""Shared /etc/os-release parsing for RHEL-family osdetect tools."""
```
petrutlucian94: I'd rename the module to "redhat_common.py"

coriolis/osmorphing/osdetect/common.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def detect_os_from_os_release(os_release, *, match_ids):
+    """Detect a RHEL-family distro from /etc/os-release.
+
+    Each osdetect module passes the ``ID`` value(s) it handles, e.g.
+    ``match_ids={"rocky"}`` or ``match_ids={"centos", "almalinux"}``.
+    """
+    if not os_release:
+        return {}
```
petrutlucian94: At the very least, we should log a warning if we haven't received the os_release dict.

coriolis/osmorphing/osdetect/common.py
```diff
@@ -0,0 +1,55 @@
...
+    if os_id not in match_ids:
+        return {}
+
+    default_name = OS_RELEASE_ID_MAP.get(os_id)
+    if not default_name:
+        return {}
+
+    version = os_release.get("VERSION_ID")
+    if not version:
+        return {}
```
petrutlucian94: Same here, the OS id matches but it doesn't include a version. Let's log a warning.

coriolis/osmorphing/osdetect/common.py
```diff
@@ -0,0 +1,55 @@
...
+    if not os_release:
+        return {}
+
+    os_id = (os_release.get("ID") or "").lower()
+    if os_id not in match_ids:
+        return {}
+
+    default_name = OS_RELEASE_ID_MAP.get(os_id)
+    if not default_name:
+        return {}
```
petrutlucian94: Warning? The id matches but we don't have a default distribution name for that given identifier.

coriolis/osmorphing/osdetect/common.py
```diff
@@ -0,0 +1,55 @@
...
+
+    version = os_release.get("VERSION_ID")
+    if not version:
+        return {}
+
+    name = os_release.get("NAME") or default_name
+    if os_id == "centos" and "Stream" in name:
```
petrutlucian94: Do we really need this special check for centos stream? I tried out a docker image and got `NAME="CentOS Stream"`, which already matches `CENTOS_STREAM_DISTRO_IDENTIFIER`

Speaking of which, won't there always be a `NAME` entry? Do we actually need a dict of distro names?

coriolis/osmorphing/osdetect/centos.py
```diff
@@ -1,46 +1,17 @@
...
-                        distribution_name = CENTOS_STREAM_DISTRO_IDENTIFIER
-                    info = {
-                        "os_type": constants.OS_TYPE_LINUX,
-                        "distribution_name": distribution_name,
-                        "release_version": version,
-                        "friendly_release_name": "%s Version %s" % (
-                            distribution_name, version)}
-        return info
+        return common.detect_os_from_os_release(
+            self._get_os_release(),
```
petrutlucian94: RHEL 6 and derivate OS-es do not have the /etc/os-release file. Are we going to drop RHEL 6 support? If so, the PR description must clearly state it. We can also remove RHEL 6 specific os-morphing code.

coriolis/osmorphing/osdetect/common.py
fabi200123: The Stream was added starting with CentOS 8, regarding the NAME entry, I agree on using `NAME`, this was the previous name that was used (didn't want to diverge too much)

petrutlucian94: My point is that the default will probably never be used, making `OS_RELEASE_ID_MAP ` redundant.

coriolis/osmorphing/osdetect/centos.py
fabi200123: Left a **NOTE** in the PR description mentioning the `/etc/os-release` was added starting Redhat version 7, anything before not having it and thus the logic not working for them.

petrutlucian94: As discussed during the daily call, it turns out that we still have customers using RHEL 6, so we can't drop it yet.

petrutlucian94: LE: apparently our docs specify that we only support RHEL 7+, so this decision got overturned again and we can proceed to remove RHEL 6 support.

https://cloudbase.it/coriolis-getting-started/#Supported_guest_operating_systems

fabi200123: Per last discussion, we will support just the OSes listed here: https://cloudbase.it/coriolis-getting-started/#Supported_guest_operating_systems

Thus, we will drop support for rhel6 and related distros

coriolis/osmorphing/osdetect/redhat_common.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 2026 Cloudbase Solutions Srl
+# All Rights Reserved.
+
+"""Shared /etc/os-release parsing for RHEL-family osdetect tools."""
+
+from typing import Dict
```
petrutlucian94: nit, just for future reference: no need to use Dict from typing, you can declare the type annotation as `dict[str, str]`

coriolis/osmorphing/osdetect/redhat_common.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def detect_os_from_os_release(
+        os_release: Optional[Dict[str, str]],
+        *,
+        match_ids: Set[str]) -> Dict[str, str]:
+    """Detect a RHEL-family distro from /etc/os-release.
+
+    :param os_release: dict populated by ``/etc/os-release`` values, as
+        returned by ``BaseLinuxOSDetectTools._get_os_release()``.
```
petrutlucian94: At this point, can't we just call `get_os_release` here and drop the parameter?

coriolis/osmorphing/centos.py
```diff
@@ -22,7 +22,11 @@ class BaseCentOSMorphingTools(redhat.BaseRedHatMorphingTools):
     @classmethod
     def check_os_supported(cls, detected_os_info):
         supported_oses = [
-            CENTOS_STREAM_DISTRO_IDENTIFIER, CENTOS_DISTRO_IDENTIFIER]
+            CENTOS_STREAM_DISTRO_IDENTIFIER,
+            CENTOS_DISTRO_IDENTIFIER,
+            "CentOS Linux",
```
petrutlucian94: We should probably define constants for "CentOS Linux", "AlmaLinux" and "Oracle Linux Server".

petrutlucian94:
Since we're now dropping RedHat 6 support, the following line should be updated as well: https://github.com/cloudbase/coriolis/blob/e7cc804089b5256c9943ee0f9b62fd330bd03033/coriolis/osmorphing/redhat.py#L36-L37

coriolis/osmorphing/osdetect/redhat_common.py
```diff
@@ -0,0 +1,68 @@
...
+CENTOS_STREAM_DISTRO_IDENTIFIER = "CentOS Stream"
+RED_HAT_DISTRO_IDENTIFIER = "Red Hat Enterprise Linux"
+ORACLE_DISTRO_IDENTIFIER = "Oracle Linux"
+CENTOS_LINUX_DISTRO_IDENTIFIER = "CentOS Linux"
+ALMALINUX_DISTRO_IDENTIFIER = "AlmaLinux"
+ORACLE_LINUX_SERVER_DISTRO_IDENTIFIER = "Oracle Linux Server"
+
+
+class RedHatOSReleaseMixin:
+    """Mixin providing RHEL-family /etc/os-release detection.
```
petrutlucian94: Why limit the mixin to /etc/os-release parsing? It can be used for any logic that's reused across RHEL derivatives.

fabi200123: Fair point! Updated it

coriolis/osmorphing/osdetect/centos.py
```diff
@@ -1,46 +1,17 @@
...
-                    if CENTOS_STREAM_DISTRO_IDENTIFIER in distro:
-                        distribution_name = CENTOS_STREAM_DISTRO_IDENTIFIER
-                    info = {
-                        "os_type": constants.OS_TYPE_LINUX,
-                        "distribution_name": distribution_name,
-                        "release_version": version,
-                        "friendly_release_name": "%s Version %s" % (
-                            distribution_name, version)}
-        return info
+        return self._detect_os_from_os_release({"centos", "almalinux"})
```
petrutlucian94: I'm wondering if we actually need the `match_ids` filter. without it, it would be able to detect the OS info for any distro that provides `/etc/os-release` (most modern releases that include systemd).

If we drop the `match_ids` parameter, `_detect_os_from_os_release` can become `detect_os` and it no longer needs to be defined by every RHEL deriviative class.

petrutlucian94: That being said, there's also the option to move `_detect_os_from_os_release` to `BaseLinuxOSDetectTools` and eventually use it for Debian based distros as well, once we drop the support for pre-systemd Debian/Ubuntu releases.

petrutlucian94: By going this route, we can have a single `osdetect` class for all RedHat derivatives.

petrutlucian94: We could even have a single `LinuxOSDetectUsingOSRelease` class for all Linux distros that include `/etc/os-release`, eventually it would be used for most if not all Linux distros.

petrutlucian94 (approved):

PR #467, opened 2026-07-03
Support migrating volumes_info entries with legacy disk ids

fabi200123:
This PR adds the logic required to not delete and replicate again the disks if the disk_id gets updated, but referencing the same disk that can be checked via the legacy_ids options.

### DIFF
--- a/coriolis/tasks/replica_tasks.py
+++ b/coriolis/tasks/replica_tasks.py
@@ -74,6 +74,60 @@ def _check_ensure_volumes_info_ordering(export_info, volumes_info):
     return ordered_volumes_info
 
 
+def _update_legacy_disk_ids_in_volumes_info(export_info, volumes_info):
+    """Migrates volumes_info entries keyed on legacy disk identifiers.
+
+    Source providers may change their disk identification methodology
+    over time (e.g. the VMware provider moving from per-VM device keys
+    to virtual disk UUIDs). Replicas created before such a change have
+    volumes_info entries whose 'disk_id' no longer matches any disk 'id'
+    in a freshly-fetched export_info, which would lead the destination
+    provider to consider the existing volumes orphaned (deleting them)
+    and to create brand new ones.
+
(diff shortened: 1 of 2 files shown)


coriolis/tasks/replica_tasks.py
```diff
@@ -74,6 +74,60 @@ def _check_ensure_volumes_info_ordering(export_info, volumes_info):
...
+    existing_volume_disk_ids = {
+        str(v.get('disk_id')) for v in volumes_info}
+    for volume_info in volumes_info:
+        disk_id = str(volume_info.get('disk_id'))
+        if disk_id in current_ids:
+            continue
+        new_disk_id = legacy_ids_map.get(disk_id)
+        if not new_disk_id:
+            continue
+        if new_disk_id in existing_volume_disk_ids:
```
Dany9966, line 113: This is an odd edge-case, basically pretending like a volume can be attached twice to the same VM. Or there's a UUID conflict. It's fine for now.

coriolis/tasks/replica_tasks.py
```diff
@@ -291,6 +345,12 @@ def _run(self, ctxt, instance, origin, destination, task_info,
         connection_info = base.get_connection_info(ctxt, destination)
 
         volumes_info = task_info.get("volumes_info", [])
+        # NOTE: migrate any volumes_info entries of pre-existing replicas
+        # which are still keyed on a legacy disk identifier (e.g. VMware
+        # device keys before the switch to virtual disk UUIDs) before the
+        # provider matches them against the current export_info disk ids:
+        _update_legacy_volumes_info_disk_ids(export_info, volumes_info)
```
Dany9966: Can we rename this into `_update_legacy_disk_ids_in_volumes_info` please?

Dany9966 (changes requested):

Dany9966 (approved):

PR #469, opened 2026-07-06
Fix guest grub bootloader entries by temporarily disabling os-prober

Cristi1324:
Prevent os-prober from adding host boot entries during guest GRUB regeneration by temporarily setting `GRUB_DISABLE_OS_PROBER=true`.


petrutlucian94 (approved):

petrutlucian94:
For context: superfluous boot entries were added during the os-morphing process:
image

PR #470, opened 2026-07-07
Avoid UEFI GRUB entries on BIOS-booted guests

mihaelabalutoiu:
When OSMorphing runs from a UEFI-booted minion while migrating a BIOS-booted guest, the guest chroot still exposes the minion's `/sys/firmware/efi` (the minion's `/proc /sys /dev /run` are bind-mounted into the chroot). Red Hat–family bootloader tooling detects that path and assumes the guest is UEFI-booted, so it generates UEFI-specific GRUB entries (`linuxefi`/`initrdefi`) for a BIOS guest.

Since the BIOS GRUB (i386-pc) has no such commands, the guest fails to boot after migration:

`error: can't find command 'linuxefi'`
`error: can't find command 'initrdefi'`

## Solution

1. **Mask the minion's firmware in the guest chroot.** When the migrated OS firmware type is BIOS and the minion exposes EFI firmware, mount a tmpfs over `/sys/firmware` inside the mounted guest root. This prevents tools running in  the chroot (e.g. `grub2-mkconfig` via `/etc/grub.d/10_linux`) from incorrectly detecting UEFI and emitting UEFI boot entries, so a BIOS guest gets `linux16`/`initrd16`.

2. **Avoid unnecessary GRUB regeneration.** The cloud-init handling now regenerates `grub.cfg` only when `cloud-init=disabled` is actually present in `/etc/default/grub`, removing a gratuitous `grub2-mkconfig` invocation that widened exposure to this issue.

### DIFF
--- a/coriolis/osmorphing/base.py
+++ b/coriolis/osmorphing/base.py
@@ -550,9 +550,11 @@ class BaseLinuxOSMorphingTools(BaseOSMorphingTools):
             self._exec_cmd_chroot(
                 "sed -i '/cloud-init=disabled/d' %s" % system_conf_disabler)
         if self._test_path(grub_conf_disabler):
-            self._exec_cmd_chroot(
-                "sed -i '/cloud-init=disabled/d' %s" % grub_conf_disabler)
-            self._schedule_grub2_update()
+            contents = self._read_file_sudo(grub_conf_disabler)
+            if "cloud-init=disabled" in contents:
+                self._exec_cmd_chroot(
+                    "sed -i '/cloud-init=disabled/d' %s" % grub_conf_disabler)
+                self._schedule_grub2_update()
 
     def _reset_cloud_init_run(self):
         self._exec_cmd_chroot("cloud-init clean --logs")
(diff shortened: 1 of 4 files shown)


coriolis/osmorphing/base.py
```diff
@@ -550,9 +550,11 @@ def _ensure_cloud_init_not_disabled(self):
             self._exec_cmd_chroot(
                 "sed -i '/cloud-init=disabled/d' %s" % system_conf_disabler)
         if self._test_path(grub_conf_disabler):
-            self._exec_cmd_chroot(
-                "sed -i '/cloud-init=disabled/d' %s" % grub_conf_disabler)
-            self._schedule_grub2_update()
+            contents = self._read_file_sudo(grub_conf_disabler)
```
petrutlucian94, line 553: Unrelated change but it seems useful, avoiding an unnecessary grub update.

petrutlucian94:
Please add a PR description, describing the problem and the solution.

mihaelabalutoiu:
> Please add a PR description, describing the problem and the solution.

Done

PR #471, opened 2026-07-14
Add real_time_clock_utc option to cloudbase_init_conf (default=False)

fabi200123:
This PR adds the option to configure `real_time_clock_utc` in the cloudbase-init config (default value is `False`, as per cloudbase-init docs) to fix issues when Windows VMs are migrated and the time is changes to UTC for a few moments. 

The issues were seen when migrating Windows VMs to LXD, and the VM was firstly setting time to UTC and just after a few moments, it got changed back to what was on the source.

### DIFF
--- a/coriolis/osmorphing/windows.py
+++ b/coriolis/osmorphing/windows.py
@@ -498,7 +498,8 @@ class BaseWindowsMorphingTools(base.BaseOSMorphingTools):
 
     def _write_cloudbase_init_conf(self, cloudbaseinit_base_dir,
                                    local_base_dir, com_port="COM1",
-                                   metadata_services=None, plugins=None):
+                                   metadata_services=None, plugins=None,
+                                   real_time_clock_utc: bool = False):
         if metadata_services is None:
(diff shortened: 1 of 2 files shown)


coriolis/osmorphing/windows.py
```diff
@@ -498,7 +498,8 @@ def _write_local_script(self, base_dir, script_path, priority=50):
 
     def _write_cloudbase_init_conf(self, cloudbaseinit_base_dir,
                                    local_base_dir, com_port="COM1",
-                                   metadata_services=None, plugins=None):
+                                   metadata_services=None, plugins=None,
+                                   real_time_clock_utc: bool = False):
```
Dany9966, line 502: Please expose this argument to `_install_cloudbase_init` method, that's the one that gets called by providers actually. Then you can just add that arg to the call instead of overriding in the provider code.

Dany9966 (changes requested):

Dany9966 (approved):

PR #474, opened 2026-07-20
Add version-specific support utilities for RHEL-family tools

mihaelabalutoiu:
This PR introduces dynamic version detection to RHEL-family OSMorphing tools enabling support for RHEL 7-10+ through unified implementations.

### DIFF
--- a/coriolis/osmorphing/base.py
+++ b/coriolis/osmorphing/base.py
@@ -11,6 +11,9 @@ from oslo_log import log as logging
 from six import with_metaclass
 import yaml
 
+from packaging.version import InvalidVersion
+from packaging.version import Version
+
 from coriolis import exception
 from coriolis.osmorphing.netpreserver import factory
 from coriolis import utils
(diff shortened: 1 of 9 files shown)


coriolis/osmorphing/base.py
```diff
@@ -334,6 +334,17 @@ def _version_supported_util(cls, version, minimum, maximum=None):
...
 
+    @classmethod
+    def _parse_major_version_util(cls, version):
+        if not version:
+            raise ValueError("Empty or missing version: %r" % (version,))
+        match = re.match(r"^([0-9]+)", str(version))
+        if not match:
+            raise ValueError(
+                "Could not parse major version from release string: %r"
+                % (version,))
```
petrutlucian94: Passing a tuple when formatting a single value is a bit unusual but still correct. There's also the option to use format strings, e.g. `f"Could not parse major version from release string: {version}"`

petrutlucian94 (approved):
This trivial helper is worth having. If we'll ever need to parse semver and retrieve the minor version or patch version, we can use `packaging.version`: https://packaging.pypa.io/en/latest/version.html

coriolis/osmorphing/base.py
```diff
@@ -334,6 +336,19 @@ def _version_supported_util(cls, version, minimum, maximum=None):
 
         return True
 
+    @classmethod
+    def _parse_major_version_util(cls, version):
```
petrutlucian94: At this point, I'd to the following:
* rename this to `_parse_version_util`
* return the `Version` object so that the caller can easily retrieve the major, minor or patch number. It also allows comparing versions
* set a type hint: `def _parse_version_util(cls, version: str) -> Version:`

petrutlucian94 (changes requested):

coriolis/osmorphing/base.py
mihaelabalutoiu: Thank you a lot for that, I updated the code.

petrutlucian94 (approved):

PR #476, opened 2026-07-22
Add a method to know if providers support or not shared disks in clustered transfers

fabi200123:
This PR adds the following:
- a method for providers (`supports_shared_disks`) to know if they have implemented the support for shared disks or not

### DIFF
--- a/coriolis/providers/base.py
+++ b/coriolis/providers/base.py
@@ -22,6 +22,23 @@ class BaseProvider(object, with_metaclass(abc.ABCMeta)):
         """Platform type."""
         raise NotImplementedError("Missing provider platform attribute.")
 
+    @classmethod
+    def supports_shared_disks(cls) -> bool:
+        """Whether this provider supports shared (multi-writer) disks.
+
+        Destination import providers that can attach the same volume to
+        multiple instances (e.g. Libvirt ``<shareable/>``) should override
+        this to return ``True``. Defaults to ``False``.
+
+        In order to coordinate shared disk operations, the conductor will mark
+        one of the instances as owner of the shared disk.
+
+        As such, operations such as disk creation, deletion, resize or
+        data migration should be handled by tasks that are targeting the
+        owner instance.
+        """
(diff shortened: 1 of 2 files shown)


coriolis/providers/base.py
```diff
@@ -22,6 +22,16 @@ def platform(self) -> str:
...
         raise NotImplementedError("Missing provider platform attribute.")
 
+    @classmethod
+    def supports_shared_disks(cls) -> bool:
+        """Whether this provider supports shared (multi-writer) disks.
+
+        Destination import providers that can attach the same volume to
+        multiple instances (e.g. Libvirt ``<shareable/>``) should override
+        this to return ``True``. Defaults to ``False``.
+        """
```
petrutlucian94, line 39: I'd also add the following since the ability to attach the volumes to multiple instances isn't all that's required on the provider side.

```
In order to coordinate shared disk operations, the conductor will mark one of the instances
as owner of the shared disk.

As such, operations such as disk creation, deletion, resize or data migration should be handled
by tasks that are targeting the owner instance.
```

fabi200123: Done

petrutlucian94 (approved):

coriolis/tasks/replica_tasks.py
```diff
@@ -1072,6 +1072,15 @@ def _run(self, ctxt, instance, origin, destination, task_info,
                 "Replica Import validation for destination platform "
                 "'%s'" % destination_type)
 
+        disks = export_info.get("devices", {}).get("disks") or []
+        if any(d.get("owner") and d.get("owner") != instance
+               for d in disks if d):
+            if not destination_provider.supports_shared_disks():
+                raise exception.NotSupportedOperation(
+                    operation=(
+                        "Destination platform '%s' does not support shared "
```
Dany9966: Please mention that the `Destination platform {destination_type} does not support synchronized transfer of instances that have shared disks.`

PR #478, opened 2026-07-28
db: Port database migration scripts from sqlalchemy-migrate to alembic

claudiubelu:
`oslo_db` has deprecated `sqlalchemy-migrate` in favor of `alembic`, and the old facade no longer exists in current `oslo.db` releases.

Adds the alembic config, env, and template scaffolding.

Relocates the existing `migrate_repo` scripts under `coriolis/db/sqlalchemy/alembic/versions/`.

Updates database migrations to alembic. Most updates are straight-forward and repetitive, which some exceptions: `001`, `016`, `018`, `020`.

In some scripts (e.g.: `001`), the code was shifted around to account for the proper table creation order (e.g.: table Y might have a foreign key from table X).

Scripts `014` and `016` deviate from the alembic db migration script pattern in order to keep diffs to a minimum.

Downgrades are left unimplemented (`raise NotImplementedError`), matching the original scripts which never defined a `downgrade()` either.

Rewrites `coriolis/db/sqlalchemy/migration.py` to drive the alembic revision chain added instead of `oslo_db.sqlalchemy.migration`, which no longer exists in current `oslo.db` releases.

`db_sync` auto-detects a pre-existing sqlalchemy-migrate `migrate_version` table left over from the old `migrate_repo`. It stamps the database onto the equivalent alembic revision instead of re-running already-applied migrations, then proceeds to upgrade normally.

The `oslo.db` upper constraint is removed.


coriolis/db/sqlalchemy/migration.py
```diff
@@ -3,25 +3,74 @@
...
+    inspector = sqlalchemy.inspect(engine)
+    if LEGACY_VERSION_TABLE not in inspector.get_table_names():
+        return
+
+    with engine.connect() as conn:
+        legacy_version = conn.execute(
+            sqlalchemy.text(
+                f"SELECT version FROM {LEGACY_VERSION_TABLE}")).scalar()
+
+    if legacy_version != LEGACY_FINAL_VERSION:
```
Dany9966: Is there any reason why we do this? Does it really matter that it's not the last DB version?

claudiubelu: Indeed, it should work with any version, since the scripts have been ported 1:1. Done.

Dany9966 (approved):

PR #479, opened 2026-07-29
Refactor Ubuntu osdetect logic

fabi200123:
This PR further refactors the osdetect logic, moving ubuntu detection to be based on the `/etc/os-release` check from `class LinuxOSDetectUsingOSRelease`.


petrutlucian94:
LGTM. Note that all LTS releases from 16.04 onwards have `/etc/os-release`, dropping 12.04 /14.04 support seems reasonable at this point. We recently dropped RHEL 6 support.

petrutlucian94 (approved):

coriolis/osmorphing/ubuntu.py
petrutlucian94: We can bump this to 16.04.

fabi200123: Agreed! Updated it

PR #485, opened 2026-08-04
Remove CD-ROM from block devices

Cristi1324:
image

### DIFF
--- a/coriolis/osmorphing/osmount/base.py
+++ b/coriolis/osmorphing/osmount/base.py
@@ -22,6 +22,9 @@ LOG = logging.getLogger(__name__)
 
 MAJOR_COLUMN_INDEX = 4
 
+# lsblk TYPE values that represent migration volumes.
+_VOLUME_BLOCK_DEVICE_TYPES = frozenset({"disk"})
+
 
 class BaseOSMountTools(object, with_metaclass(abc.ABCMeta)):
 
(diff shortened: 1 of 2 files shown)


coriolis/osmorphing/osmount/base.py
```diff
@@ -22,6 +22,9 @@
 
 MAJOR_COLUMN_INDEX = 4
 
+# lsblk TYPE values that are not migration volumes (CD-ROM, loopback, etc.).
+_NON_VOLUME_BLOCK_DEVICE_TYPES = frozenset({"part", "rom", "loop"})
```
claudiubelu: hm, considering that we don't migrate loopback devices in the first place, I wonder how `"loop"` can be encountered during OS morphing.

Dany9966: Maybe instead of filtering out the types, filter in? 
```suggestion
_VOLUME_BLOCK_DEVICE_TYPES = frozenset({"disk", "part"})
```

```python
            if (len(parts) >= 2
                    and parts[1] in _VOLUME_BLOCK_DEVICE_TYPES):
                volume_devs.append("/dev/%s" % parts[0])

Dany9966: Actually that method should return only disk block devices. So only return `disk` types.

Dany9966: > hm, considering that we don't migrate loopback devices in the first place, I wonder how `"loop"` can be encountered during OS morphing.

No, but the `lsblk` command returns the devices available on the minion, so it apperently returns the loop devices on the minion machine. Not sure why it does that, iirc this didn't happen in the past.

Dany9966 (approved):

PR #486, opened 2026-08-06
Add Ubuntu 26.04 support for OSMorphing minions

mihaelabalutoiu:
This PR implements the following changes:
- Remove the nested shell wrapper from the `apt-get` command;
- Improve shell quoting for GRUB configuration commands:
   - `set_grub_value` placed quotes by hand when building its sed scripts, so values containing spaces or quotes could be split into separate options instead of reaching sed as a single argument.
- Improve environment variable handling in OSMount tools:
    - the OSMount setup commands relied on `sudo -E` to forward the proxy configuration to the package managers, which does not work on `sudo` implementations that do not support the flag.
- Improve environment variable handling for chroot commands:
    - `sudo-rs`, the default sudo on Ubuntu 26.04, does not implement the `-E` flag, so proxy variables no longer reached commands run inside the chroot. They are now passed through `env(1)`, which `chroot(1)` preserves.
- Add `env` command availability check on OSMorphing hosts;
- Disable `pty` allocation for SSH commands
- Sever mount propagation on OSMorphing chroot binds:
    -  Systemd uses shared mount propagation, allowing mounts created under `/proc`, `/sys`, `/dev` or `/run` on the minion to propagate into the OSMorphing chroot.
    - On [Ubuntu 26.04](https://documentation.ubuntu.com/release-notes/26.04/summary-for-lts-users/#systemd-259) this can leave systemd credential mounts under `/run` busy and cause the final `umount -R` to fail with `target is busy`.

### DIFF
--- a/coriolis/osmorphing/debian.py
+++ b/coriolis/osmorphing/debian.py
@@ -33,6 +33,21 @@ class BaseDebianMorphingTools(base.BaseLinuxOSMorphingTools):
 
     netplan_base = "etc/netplan"
 
+    def __init__(self, conn, os_root_dir, os_root_dev, hypervisor,
+                 event_manager, detected_os_info, osmorphing_parameters,
+                 operation_timeout=None):
+        super(BaseDebianMorphingTools, self).__init__(
+            conn, os_root_dir, os_root_dev, hypervisor, event_manager,
+            detected_os_info, osmorphing_parameters, operation_timeout)
+
+        # NOTE: every dpkg invocation may run maintainer scripts which prompt
+        # through debconf (e.g. asking for a keyboard layout, or acknowledging
+        # a pending kernel upgrade), which would hang the OSMorphing operation
+        # indefinitely. Both the install and the uninstall paths are affected,
+        # so the frontend is declared non-interactive for all the commands run
+        # by these tools instead of being set on individual operations.
+        self.set_environment({'DEBIAN_FRONTEND': 'noninteractive'})
+
     @classmethod
     def check_os_supported(cls, detected_os_info):
         if detected_os_info['distribution_name'] != (
@@ -242,9 +257,8 @@ class BaseDebianMorphingTools(base.BaseLinuxOSMorphingTools):
             self._exec_cmd_chroot(deb_reconfigure_cmd)
 
             apt_get_cmd = (
-                '/bin/bash -c "DEBIAN_FRONTEND=noninteractive '
                 'apt-get install %s -y '
-                '-o Dpkg::Options::=\'--force-confdef\'"' % (
+                '-o Dpkg::Options::=--force-confdef' % (
                     " ".join(package_names)))
             self._exec_cmd_chroot(apt_get_cmd)
         except Exception as err:
--- a/coriolis/osmorphing/debian.py
+++ b/coriolis/osmorphing/debian.py
@@ -242,10 +242,10 @@ class BaseDebianMorphingTools(base.BaseLinuxOSMorphingTools):
             self._exec_cmd_chroot(deb_reconfigure_cmd)
 
             apt_get_cmd = (
-                '/bin/bash -c "DEBIAN_FRONTEND=noninteractive '
                 'apt-get install %s -y '
-                '-o Dpkg::Options::=\'--force-confdef\'"' % (
+                '-o Dpkg::Options::=--force-confdef' % (
                     " ".join(package_names)))
+            self._environment['DEBIAN_FRONTEND'] = 'noninteractive'
             self._exec_cmd_chroot(apt_get_cmd)
         except Exception as err:
             raise exception.FailedPackageInstallationException(
--- a/coriolis/osmorphing/osmount/base.py
+++ b/coriolis/osmorphing/osmount/base.py
@@ -102,6 +102,7 @@ class BaseSSHOSMountTools(BaseOSMountTools):
         self._ssh = ssh
 
     def setup(self):
+        utils.check_env_command(self._ssh)
         if self._allow_ssh_env_vars():
             self._ssh.close()
             self._connect()
--- a/coriolis/osmorphing/osmount/base.py
+++ b/coriolis/osmorphing/osmount/base.py
@@ -128,6 +128,31 @@ class BaseSSHOSMountTools(BaseOSMountTools):
             raise exception.OSMorphingSSHOperationTimeout(
                 cmd=cmd, timeout=timeout) from ex
 
+    def _exec_sudo_env_cmd(self, cmd, timeout=None):
+        """
+        Runs a sudo command that also passes all the environment variables to
+        the underlying command. Replaces sudo's -E flag, which is not currently
+        supported in all shipped sudo variants (like sudo-rs).
+        """
+
+        if not timeout:
+            timeout = self._osmount_operation_timeout
+        env_cmd = "sudo %s%s" % (
+            utils.get_env_command_prefix(self._environment), cmd)
+        try:
+            return utils.exec_ssh_cmd(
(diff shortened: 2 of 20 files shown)


coriolis/osmorphing/osmount/base.py
```diff
@@ -128,6 +128,34 @@ def _exec_cmd(self, cmd, timeout=None):
...
+        """
+        Runs a sudo command that also passes all the environment variables to
+        the underlying command. Replaces sudo's -E flag, which is not currently
+        supported in all shipped sudo variants (like sudo-rs).
+        """
+
+        if not timeout:
+            timeout = self._osmount_operation_timeout
+        environment = self._environment
+        if extra_env:
```
Dany9966: Why set a `extra_env` parameter instead of setting the additional env in `self._environment`? 
This is just to pass in the extra env once?

coriolis/osmorphing/osmount/base.py
```diff
@@ -128,6 +128,34 @@ def _exec_cmd(self, cmd, timeout=None):
...
+            environment = {**(environment or {}), **extra_env}
+        env_cmd = "sudo %s%s" % (
+            utils.get_env_command_prefix(environment), cmd)
+        try:
+            return utils.exec_ssh_cmd(
+                self._ssh,
+                env_cmd,
+                environment=environment,
+                get_pty=True,
```
Dany9966: We should also set this to `False` from now on.

coriolis/utils.py
```diff
@@ -316,6 +317,20 @@ def list_ssh_dir(ssh, remote_path):
...
+def get_env_command_prefix(environment):
+    """
+    Each variable is shell-quoted individually, so values containing
+    spaces or other shell-sensitive characters are passed through
+    verbatim.
+    """
+
+    if not environment:
+        return ""
+    return "env %s " % " ".join(
```
Dany9966, line 330: I first avoided `env` because it's not guaranteed to be there on every minion machine, but it's a fair requirement after all. 
Therefore, can you please also add a validation command and error out in case `env` command doesn't exist?

coriolis/osmorphing/osmount/base.py
Dany9966: Can you please make sure `get_pty` is set to False everywhere? 
Setting this to `True` will put stderr output into stdout and may break some command parsing in the osmorphing code. 
pty was required in the past as some older minion images didn't work otherwise, but we can revert safely now.

coriolis/utils.py
```diff
@@ -390,9 +405,16 @@ def wrapper():
 
 def exec_ssh_cmd_chroot(ssh, chroot_dir, cmd, environment=None, get_pty=False,
                         timeout=None):
-    return exec_ssh_cmd(ssh, "sudo -E chroot %s %s" % (chroot_dir, cmd),
-                        environment=environment, get_pty=get_pty,
-                        timeout=timeout)
+    return exec_ssh_cmd(
+        ssh,
+        "sudo %schroot %s %s" % (
```
Dany9966, line 431: I'm not sure this works (as I've tried this before). Please retest with an actual `sudo chroot install_cmd`.

Dany9966 (changes requested):

coriolis/utils.py
mihaelabalutoiu: I retested and confirm it works as expected as is.

coriolis/utils.py
mihaelabalutoiu: I added `utils.check_env_command()` method called from `BaseSSHOSMountTools.setup()` so it fails early rather than partway through OSMorphing.

coriolis/osmorphing/osmount/base.py
mihaelabalutoiu: You're right, I've set it directly on `self._environment` and dropped the `extra_env` parameter. Thank you for that!

coriolis/osmorphing/osmount/base.py
mihaelabalutoiu: Gotcha, I made a separate commit for that: https://github.com/cloudbase/coriolis/pull/486/changes/57d13133dda65641ee0070feb8152fa4efd29fe1

coriolis/utils.py
```diff
@@ -312,10 +313,43 @@ def list_ssh_dir(ssh, remote_path):
...
+    """
+
+    try:
+        exec_ssh_cmd(ssh, "command -v env", get_pty=False)
+    except (exception.SSHCommandFailed,
+            exception.SSHCommandNotFoundException) as ex:
+        raise exception.CoriolisException(
+            "The 'env' command is unavailable on the OSMorphing host. It is "
+            "required in order to pass environment variables (such as proxy "
+            "settings) to the commands run during OSMorphing.") from ex
```
Dany9966: Let's please replace the error text with the following:

```suggestion
            "The 'env' command is unavailable on the OSMorphing minion machine. "
            "It is required in order to forward environment variables to commands "
            "run on the migrated machine (such as proxy settings). Please switch the "
            "minion machine image/template to one that has this command in place."
        ) from ex
```

Dany9966 (changes requested):

coriolis/utils.py
mihaelabalutoiu: Done, thank you for that!

coriolis/osmorphing/debian.py
```diff
@@ -242,10 +242,10 @@ def install_packages(self, package_names):
             self._exec_cmd_chroot(deb_reconfigure_cmd)
 
             apt_get_cmd = (
-                '/bin/bash -c "DEBIAN_FRONTEND=noninteractive '
                 'apt-get install %s -y '
-                '-o Dpkg::Options::=\'--force-confdef\'"' % (
+                '-o Dpkg::Options::=--force-confdef' % (
                     " ".join(package_names)))
+            self._environment['DEBIAN_FRONTEND'] = 'noninteractive'
```
petrutlucian94: `osmount` tools set the `DEBIAN_FRONTEND` env variable as part of the `setup` call. The `osmorphing` tools do not have a `setup` method but I think we can move it to the `__init__` method.

Image

`uninstall_packages` needs it as well and it's cleaner to set this as part of the constructor as opposed to `install_packages` / `uninstall_packages`. Modifying the shared `_environment` field as a side effect here isn't a clean approach.

mihaelabalutoiu: Gotcha,  I updated the code,  moved it to the constructor, so it covers `uninstall_packages` too and the side effect is gone from `install_packages`. Please take a look now, thank you!

coriolis/tests/osmorphing/test_base.py
```diff
@@ -99,9 +100,57 @@ def test_check_os_supported_not_implemented(self):
...
+
+    def test_set_environment_does_not_share_the_given_mapping(self):
+        """The OSMount environment is shared, so it must not be mutated."""
+        environment = {'http_proxy': 'http://10.0.0.1:3128'}
+
+        self.os_morphing_tools.set_environment(environment)
+        self.os_morphing_tools._environment['DEBIAN_FRONTEND'] = (
+            'noninteractive')
+
+        self.assertEqual({'http_proxy': 'http://10.0.0.1:3128'}, environment)
```
Dany9966: I don't think we care for testing this. It's basically testing what `dir.update()` does, not the scope of `set_environment`.

Dany9966 (changes requested):

coriolis/tests/osmorphing/test_base.py
mihaelabalutoiu: Done

Dany9966 (approved):

PR #487, opened 2026-08-06
Add SAP licence support

fabi200123:
This PR adds the following:

* New constants for `sap_migration`, `sap_replica` and `sap_libvirt` (endpoint type)
* Provider type overrides for `SCENARIO_TYPE_TO_LICENSING_RESERVATION_MAP`: a new `ENDPOINT_TYPE_TO_LICENSING_RESERVATION_OVERRIDES` map and `_get_reservation_type` helper resolve the reservation type based on the destination endpoint type, falling back to the scenario-based map. Transfers targeting `sap_libvirt` endpoints get SAP reservation types
* The destination endpoint type is now passed down to all `_create_reservation_for_transfer` calls

Note: `MigrationLicenceFulfilledException` is no longer raised by the conductor, since the equivalent check is now enforced server-side. The conductor no longer performs the fulfilled-migration check itself.


Dany9966 (approved):

coriolis/conductor/rpc/server.py
```diff
@@ -394,16 +417,8 @@ def _check_reservation_for_transfer(self, transfer):
                 "Attempting to check reservation with ID '%s' for transfer "
                 "action '%s'", reservation_id, action_id)
             try:
-                reservation = self._licensing_client.get_reservation(
-                    reservation_id)
-
-                fulfilled_at = reservation.get("fulfilled_at", None)
-                if scenario == constants.TRANSFER_SCENARIO_LIVE_MIGRATION and (
-                        fulfilled_at):
-                    raise exception.MigrationLicenceFulfilledException(
```
Dany9966, line 403: I think an exception like this should still be forwarded to the user when a 403 is thrown.

coriolis/exception.py
```diff
@@ -556,6 +554,6 @@ class ChecksumMismatch(CoriolisException):
 class MigrationLicenceFulfilledException(Invalid):
     message = (
         "The Live Migration operation with ID '%(action_id)s' (licensing "
```
Dany9966: I would also replace this ("Live Migration") with `scenario`, just in case there will be multiple scenarios that can throw 403.
I'd add something like this:
`"The %s operation with ID .... " % transfer.scenario.replace('_', ' ').title()`

Dany9966 (changes requested):

coriolis/exception.py
fabi200123: In this case, also renamed it to `LicenceReservationFulfilledException`

Dany9966 (approved):

PR #488, opened 2026-08-06
luks: Adds tpm2_pcrs config option for LUKS-encrypted devices

claudiubelu:
Adds the mentioned config option. The default value for it is 7, which is the default value used by `systemd-cryptenroll`.

### DIFF
--- a/coriolis/cmd/worker.py
+++ b/coriolis/cmd/worker.py
@@ -9,6 +9,7 @@ from oslo_reports import guru_meditation_report as gmr
 from oslo_reports import opts as gmr_opts
 
 from coriolis import constants
+from coriolis.osmorphing.osmount import luks_mixin
 from coriolis import service
 from coriolis import utils
 from coriolis.worker.rpc import server as rpc_server
@@ -25,6 +26,7 @@ CONF.register_opts(worker_opts, 'worker')
 def main():
     worker_count, args = service.get_worker_count_from_args(sys.argv)
     CONF(args[1:], project='coriolis', version="1.0.0")
+    luks_mixin.validate_config()
     if not worker_count:
         worker_count = CONF.worker.worker_count
     utils.setup_logging()
--- a/coriolis/osmorphing/osmount/luks_mixin.py
+++ b/coriolis/osmorphing/osmount/luks_mixin.py
@@ -6,16 +6,33 @@ import json
 import os
 import re
 
+from oslo_config import cfg
+from oslo_config import types
 from oslo_log import log as logging
 
 from coriolis import constants
 from coriolis import exception
 from coriolis import utils
 
+luks_opts = [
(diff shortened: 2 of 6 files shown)


coriolis/cmd/worker.py
```diff
@@ -25,6 +26,7 @@
 def main():
     worker_count, args = service.get_worker_count_from_args(sys.argv)
     CONF(args[1:], project='coriolis', version="1.0.0")
+    luks_mixin.validate_config()
```
Dany9966, line 29: Do we have to fail the whole service for this?

coriolis/osmorphing/osmount/luks_mixin.py
```diff
@@ -41,6 +55,36 @@ def _load_script(filename):
 }
 
 
+def validate_config():
+    """Validate LUKS-related config options.
+
+    Called at worker service startup, so a misconfigured [luks] config section
+    fails fast, rather than surfacing into an actual LUKS migration.
+    """
+    if not CONF.luks.tpm2_pcrs:
```
Dany9966, line 67: I don't feel like we should force any PCRs. The only reason I asked for a validation method was to cover for user errors when building dracut/clevis strings, but if the int value is all the user has to pass and everything else is handled by `_render_tpm2_pcrs`, I guess we could also validate whether the pcr values are valid and restrict to only values that we can support (maybe 0-7, or whatever else you might have tested with or looks like it could work).

coriolis/cmd/worker.py
claudiubelu: Yes. It's a misconfiguration. It's better to fail fast rather than fail later on, when a LUKS VM has to be migrated.

coriolis/osmorphing/osmount/luks_mixin.py
claudiubelu: We are not forcing any specific PCR, we just erroring out if there is **no** PCR configured. But yeah, we can further validate the values as well, making sure they are valid (0-15).

coriolis/cmd/worker.py
Dany9966: Usually we validate options only when they're actually being used. No other option is being validated in `cmd` module. But then again, you would have to specifically edit the option and then restart. If it's invalid, you'll know straight away, but failing the service as a whole seems a bit drastic. I'm ok with this as long as worker logs can still be downloaded and the validation errors are actually contained in the `coriolis-worker.log`. Can you please double-check?

Dany9966: Also, users will have no tell if the options are valid or not unless they start migrating and notice a bunch of `500/503` errors. If you keep this validation in the osmorphing task alone, you can redirect the error message to the task as well and it'll be nicely output in the UI.

PR #494, opened 2026-08-11
tests: Increase transfer integration tests coverage

claudiubelu:
Adds `verify_checksum` `source_environment` flag through the test provider's `replicate_disks` call (False by default). `test_incremental_replica_transfer` will now also check the checksum.

Adds a `backup_writer_backend` `destination_environment` option through `_create_minion`, so a test can request the SSH backend for one transfer. Adds a transfer test using the mentioned SSH backup writer.

`data_transfer.compression_proxy` falls back to in-process gzip / zlib compression when `compressor_address` is unset (`None` by default). That means that the external compressor service scenario is not covered. Spins up a unix-socket HTTP service in-process, points `compressor_address` at it for the duration of one transfer, and asserts it was invoked.

A transfer with more than one instance is considered "clustered". The conductor runs a cross-instance sync barrier and assigns disk owners across instances.

In the added tests, each instance in this has its own disk + a disk shared between the (same disk id in both instances' `export_info`). The conductor will assign the first instance as the shared disk's owner, only its `DEPLOY_TRANSFER_DISKS` task creates a destination volume and replicates data into it, while the other instance's task records a placeholder `volumes_info` entry with `replicate_disk_data` `False` and no `volume_dev`.

Adds integration test for clustered sync-barrier abort on peer error. If one instance's task errors out while a peer instance's task of the same type is stuck waiting at the cross-instance sync barrier, the stuck peer must be aborted rather than left deadlocked forever.

Updates the test providers to support shared disks.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -36,6 +36,12 @@ jobs:
       run: |
         sudo apt-get install -y linux-modules-extra-$(uname -r)
 
+    - name: Build write_data resource binary
+      shell: bash
+      run: |
+        sudo apt-get install -y build-essential zlib1g-dev
+        make -C coriolis/resources
+
     - name: Build Docker image for integration test minions
       shell: bash
       run: |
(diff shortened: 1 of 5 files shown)


fabi200123 (approved):
LGTM!

PR #496, opened 2026-08-13
Replaces sshtunnel with a direct paramiko SSH tunnel

claudiubelu:
`sshtunnel` 0.4.0 (last release in 2021) references `paramiko.DSSKey` in `get_keys()`, which was removed in `paramiko` 3.0.  There is no maintained version of `sshtunnel` that supports `paramiko` 3+.

Replace `SSHTunnelForwarder` with `_SSHTunnel`, a small class that implements local-port-forwarding directly using `paramiko`'s
`Transport.open_channel('direct-tcpip', ...)`.  The public interface (`start` / `stop` / `local_bind_address`) is the same, so `_get_ssh_tunnel` needs only minor keyword-argument changes.

Removes `sshtunnel` from `requirements.txt` and update the unit test to patch `_SSHTunnel` instead of `SSHTunnelForwarder`.

Adds unit tests and an integration test exercising the `_SSHTunnel`.

### DIFF
--- a/coriolis/providers/replicator.py
+++ b/coriolis/providers/replicator.py
@@ -3,15 +3,17 @@
 
 import json
 import os
+import select
 import shutil
+import socket
 import tempfile
+import threading
 import time
 
 from oslo_config import cfg
 from oslo_log import log as logging
 from oslo_utils import units
 import paramiko
-from sshtunnel import SSHTunnelForwarder
 
 from coriolis import exception
 from coriolis.providers import provider_utils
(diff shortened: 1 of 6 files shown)


coriolis/providers/replicator.py
```diff
@@ -44,6 +46,95 @@
...
+
+        t = threading.Thread(target=self._accept_loop, daemon=True)
+        t.start()
+
+    def _accept_loop(self):
+        while True:
+            try:
+                sock, _ = self._listen_sock.accept()
+            except OSError:
+                return
```
Dany9966, line 92: Can we log the reason for the OSError?

coriolis/providers/replicator.py
```diff
@@ -44,6 +46,95 @@
...
+            t.start()
+
+    def _forward(self, local_sock):
+        try:
+            chan = self._transport.open_channel(
+                'direct-tcpip',
+                self._remote_bind_address,
+                local_sock.getpeername(),
+            )
+        except Exception:
```
Dany9966, line 105: Same as above

requirements.txt
```diff
@@ -44,7 +44,6 @@ strict-rfc3339
 sqlalchemy<2.0.0
 taskflow
 webob
-sshtunnel
```
Dany9966, line 47: Feels good to remove dependencies :D

Dany9966 (commented):
LGTM aside from the comments related to logging tunnel setup errors.

coriolis/providers/replicator.py
claudiubelu: Done

coriolis/providers/replicator.py
claudiubelu: Done. Also added in the try finally below.

Dany9966 (approved):

PR #500, opened 2026-08-19
Document `delete_replica_source_snapshots` return value

petrutlucian94:
`delete_replica_source_snapshots` is expected to return the updated volumes info, even if the provider method is a no-op. If nothing is returned, deleting the target replica disks will fail.

As such, we'll document the expected return value.


petrutlucian94:
Trivial PR, merging it.

PR #503, opened 2026-08-19
Refactor provider options

Cristi1324:
This PR moves common provider options into core so each provider does not have to redefine the same option schema.

Initial common keys are `set_dhcp`, `data_transfer_mechanism`, and `cloudbase_init_plugins`. Core merges those schemas into every import provider.

### DIFF
--- a/coriolis/osmorphing/windows.py
+++ b/coriolis/osmorphing/windows.py
@@ -9,6 +9,7 @@ import os
 import re
 import uuid
 
+from oslo_config import cfg
 from oslo_log import log as logging
 from packaging import version
 
@@ -16,6 +17,7 @@ from coriolis import constants, exception, utils
 from coriolis.osmorphing import base
 from coriolis.osmorphing.osdetect import windows as windows_osdetect
 
+CONF = cfg.CONF
 LOG = logging.getLogger(__name__)
 
 WINDOWS_CLIENT_IDENTIFIER = windows_osdetect.WINDOWS_CLIENT_IDENTIFIER
@@ -44,6 +46,19 @@ CLOUDBASE_INIT_DEFAULT_PLUGINS = [
     'ConfigWinRMCertificateAuthPlugin',
     'cloudbaseinit.plugins.common.localscripts.LocalScriptsPlugin',
 ]
+
+CLOUDBASE_INIT_PLUGINS_DESTINATION_OPTION = {
+    "name": "cloudbase_init_plugins",
+    "values": [
+        {
+            "id": item,
+            "name": item.rsplit(".", 1)[-1],
+        }
+        for item in CLOUDBASE_INIT_DEFAULT_PLUGINS
+    ],
+    "config_default": list(CLOUDBASE_INIT_DEFAULT_PLUGINS),
+}
+
 
 CLOUDBASE_INIT_DEFAULT_METADATA_SVCS = [
     'cloudbaseinit.metadata.services.httpservice.HttpService',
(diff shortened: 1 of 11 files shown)


coriolis/osmorphing/windows.py
```diff
@@ -499,18 +502,79 @@ def _write_local_script(self, base_dir, script_path, priority=50):
             script_path,
             remote_script_path)
 
+    def _parse_cloudbase_init_plugins(self, plugins):
```
claudiubelu, line 547: seems duplicated from https://github.com/cloudbase/coriolis/pull/501, I assume that PR is the best location for this, no?

coriolis/schemas/os_morphing_resources_schema.json
```diff
@@ -57,6 +57,18 @@
...
               "default": false
+            },
+            "cloudbase_init_plugins": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              },
+              "description": "Cloudbase-Init plugin class names for Windows guests. When set, this list is written into the guest. When omitted, the provider plugin list is used, then coriolis.conf, then the Windows morphing default."
+            },
+            "set_dhcp": {
```
claudiubelu, line 64: Right, there is also `nics_set_dhcp` above. How does that interact with this?

coriolis/osmorphing/destination_option_overrides.py
```diff
@@ -0,0 +1,54 @@
+# Copyright 2026 Cloudbase Solutions Srl
```
claudiubelu: Right, `coriolis/osmorphing/` mostly contains different target OSes, and I'd like to keep that as much as possible. I'd merge this  and destination_options.py into `coriolis/osmorphing/conf.py` instead.

coriolis/osmorphing/destination_option_overrides.py
```diff
@@ -0,0 +1,54 @@
...
+"""
+
+
+class DestinationOptionOverrides(object):
+    """Hooks for dest-options that need non-generic values or choices."""
+
+    @classmethod
+    def choices(cls, entry):
+        """Return allowed values for a registry entry."""
+        handler = getattr(cls, "_choices_%s" % entry["name"], None)
```
claudiubelu: I'd really rather not rely on this sort of magic too much. It makes the code a lot less grep-able. Why not just let overrides be overrides?

Also, as for the class name, there is no need to name it `DestinationOptionOverrides`, just let it be `DestinationOptions`, let providers override what they need to.

coriolis/osmorphing/destination_option_overrides.py
```diff
@@ -0,0 +1,54 @@
...
+        """Return a custom config_default, or None to use the generic path."""
+        handler = getattr(
+            cls, "_config_default_%s" % entry["name"], None)
+        if handler is None:
+            return None
+        return handler(conf_value, choices)
+
+    @classmethod
+    def _choices_cloudbase_init_plugins(cls):
+        from coriolis.osmorphing import windows
```
claudiubelu: move this import at the top level.

coriolis/osmorphing/destination_options.py
```diff
@@ -0,0 +1,281 @@
...
+        "name": name,
+        "values": values,
+        "config_default": config_default,
+    }
+
+
+def get_core_destination_option(name):
+    """Return the dest-options row for one registered core option."""
+    for entry in CORE_DESTINATION_OPTIONS:
+        if entry["name"] == name:
```
claudiubelu: why not use a dict at this point then?

PR #505, opened 2026-08-22
os-morphing: fix static ip configuration if the mac changes

petrutlucian94:
When "use_dhcp" is disabled, Coriolis tries to apply static IP configuration on Windows, while on Linux it defines udev rules for preserving the interface names, identifying interfaces by MAC address.

However, it expects the MAC address of the source instance to be preserved, which isn't always the case.

This change updates the Linux and Windows OS morphing code to use the new MAC address, if provided. The import provider is expected to set "new_mac_address" within the nic info if the original MAC hasn't been preserved.

Note that Coriolis cannot modify network configuration files at the moment. An exception will be raised if the original MAC address is explicitly set by the network configuration (e.g. netplan).

### DIFF
--- a/coriolis/osmorphing/base.py
+++ b/coriolis/osmorphing/base.py
@@ -1008,6 +1008,9 @@ class BaseLinuxOSMorphingTools(BaseOSMorphingTools):
 
         for nic in nics_info:
             nic_mac = nic.get('mac_address')
+            new_nic_mac = nic.get('new_mac_address') or nic_mac
+            if new_nic_mac != nic_mac:
+                LOG.info("MAC address changed: %s -> %s", nic_mac, new_nic_mac)
             nic_ips = nic.get('ip_addresses')
             if not nic_mac:
                 LOG.warning(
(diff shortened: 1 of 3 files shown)


coriolis/osmorphing/base.py
```diff
@@ -1021,21 +1024,24 @@ def _setup_network_preservation(self, nics_info) -> None:
                 ip_addresses = info.get('ip_addresses', [])
                 if mac_address and mac_address == nic_mac:
                     LOG.info(
-                        "Found matching interface for NIC '%s' with MAC '%s'",
-                        nic.get('name'), nic_mac)
-                    matching_ifaces[iface] = nic_mac
+                        "Found matching interface for NIC '%s' with MAC '%s'. "
+                        "New MAC: '%s'.",
+                        nic.get('name'), nic_mac, new_nic_mac)
+                    matching_ifaces[iface] = new_nic_mac
```
Dany9966: If there's a `mac_address` in the netpreserver info, that means that it had found a mac address defined inside of the interface configuration file. 
You set udev rule for the new mac address, true, but I'm not sure how it will behave if the old mac is in the actual interface file.
Was this tried out?

petrutlucian94: Right, my netplan file didn't have an explicit mac address, it picked the fixed ip instead (`preserve_fixed_ip=true`).

To support explicit mac addresses in the network configuration file, the net preserver would probably have to perform updates. Since this isn't supported at the moment, we can either log a warning or raise a hard error.

Dany9966 (approved):

PR #511, opened 2026-08-28
Use grubby to update kernel command line arguments

mihaelabalutoiu:
Failing to update the kernel console options breaks text console functionality on affected platforms.

The issue is caused by BLS being enabled by default in `/etc/default/grub`. On `RHEL 10`, kernel arguments are managed through `/boot/loader/entries`, and `grub2-mkconfig` does not propagate the updated options there.

This PR implements the following:

- Adds `Grub2ConfigEditor.remove_from_option`, with `grubby --remove-args` semantics: a bare name drops the argument whichever value it holds, a name/value pair drops it only on an exact match.
- Adds `_update_kernel_cmdline_args` to the base `OSMorphing` tools, editing both `GRUB2` cmdline options through `Grub2ConfigEditor`. Debian and SUSE each had their own copy of these edits and now share this one.
- Overrides it on redhat to use `grubby --update-kernel=ALL`, which reaches the BLS boot entries.

### DIFF
--- a/coriolis/osmorphing/base.py
+++ b/coriolis/osmorphing/base.py
@@ -17,6 +17,7 @@ from coriolis import exception, utils
 from coriolis.osmorphing.netpreserver import factory
 
 GRUB2_SERIAL = "serial --word=8 --stop=1 --speed=%d --parity=%s --unit=0"
+GRUB2_CMDLINE_OPTIONS = ["GRUB_CMDLINE_LINUX", "GRUB_CMDLINE_LINUX_DEFAULT"]
 LOG = logging.getLogger(__name__)
 
 IFCFG_TEMPLATE = """
@@ -756,10 +757,7 @@ class BaseLinuxOSMorphingTools(BaseOSMorphingTools):
         if self._test_path(grub_conf_disabler):
             contents = self._read_file_sudo(grub_conf_disabler)
             if "cloud-init=disabled" in contents:
-                self._exec_cmd_chroot(
-                    "sed -i '/cloud-init=disabled/d' %s" % grub_conf_disabler
-                )
-                self._schedule_grub2_update()
+                self._update_kernel_cmdline_args(args_to_remove=["cloud-init=disabled"])
 
     def _reset_cloud_init_run(self):
         self._exec_cmd_chroot("cloud-init clean --logs")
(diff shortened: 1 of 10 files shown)


coriolis/osmorphing/base.py
```diff
@@ -760,6 +760,7 @@ def _ensure_cloud_init_not_disabled(self):
                     "sed -i '/cloud-init=disabled/d' %s" % grub_conf_disabler
                 )
                 self._schedule_grub2_update()
+                self._update_kernel_cmdline_args(args_to_remove=["cloud-init=disabled"])
```
Dany9966, line 763: I don't like that both grub update and grubby are being called. It should be the one or the other, not both. 

An ideal implementation would be to call for `_update_kernel_cmdline_args` indeed, but you only implemented it for redhat based. There should also be  a debian implementation which does the former (reads /etc/default/grub, removes/adds/edits the cmdline, schedules grub2 update).

I propose the following:

1. Use Grub2ConfigEditor whenever possible instead of `sed`-ing the file directly.
This implies implementing a new method to remove cmdline entries, call it `remove_from_option`
This method should pretty much emulate what grubby does on redhat, and that is if you pass a single option
(`--remove-args=cloud-init`), then it will be removed even if it's a key_val, no matter what value cloud-init has 
in cmdline. If you pass a key_val, then only remove the key if the value matches 
(`--remove-args=cloud-init=disabled` only remove cloud-init from cmdline if it's disabled, but won't 
remove `cloud-init=enabled`).

2. We should somehow abstractize this for redhat. On base, when calling for `_update_kernel_cmdline_args`,
it should instantiate a Grub2ConfigEditor, append or remove from GRUB_CMDLINE_LINUX and 
GRUB_CMDLINE_LINUX_DEFAULT options (depending on what args_to_add or args_to_remove are being 
passed). 
If it's redhat, then simply use grubby to handle args_to_add/args_to_remove

3. (only if grubby commands take too long, otherwise treat this as optional) I think the final grubby command should also be run once at the end, so add some `schedule_grubby` 
methods as well when adding args to remove/add. (similar to `_schedule_grub2_update`)

Dany9966 (changes requested):

coriolis/osmorphing/base.py
mihaelabalutoiu: Thank you for that! I updated the code, please take a look now.

coriolis/osmorphing/suse.py
```diff
@@ -69,33 +69,7 @@ def check_os_supported(cls, detected_os_info):
         return False
 
     def disable_predictable_nic_names(self):
```
fabi200123, line 71: Worth taking into account to modify custom implementations of `disable_predictable_nic_names` in the other providers as well.
Besides that LGTM

PR #516, opened 2026-09-02
osmount: avoid stale propagated mounts

petrutlucian94:
We'll need to prevent Coriolis mounts from being propagated to other namespaces belonging to Systemd services.

These would be leaked when running "unmount -R" in this namespace, being especially troublesome if the
minion gets reused. Stale mounts can prevent the FS from being remounted (e.g. if os-morphing gets retried).

### DIFF
--- a/coriolis/osmorphing/osmount/base.py
+++ b/coriolis/osmorphing/osmount/base.py
@@ -655,6 +655,15 @@ class BaseLinuxOSMountTools(luks_mixin.LinuxLUKSMixin, BaseSSHOSMountTools):
         return os_root_dir, os_root_device
 
     def mount_os(self):
+        # Prevent Coriolis mounts from being propagated to other namespaces
+        # belonging to Systemd services. These would be leaked when running
+        # "unmount -R" in this namespace, being especially troublesome if the
+        # minion gets reused. Stale mounts can prevent the FS from being
+        # remounted (e.g. if os-morphing gets retried).
+        #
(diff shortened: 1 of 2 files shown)


coriolis/osmorphing/osmount/base.py
```diff
@@ -655,6 +655,15 @@ def _find_and_mount_root(self, devices):
...
 
     def mount_os(self):
+        # Prevent Coriolis mounts from being propagated to other namespaces
+        # belonging to Systemd services. These would be leaked when running
+        # "unmount -R" in this namespace, being especially troublesome if the
+        # minion gets reused. Stale mounts can prevent the FS from being
+        # remounted (e.g. if os-morphing gets retried).
+        #
+        # The other "--make-(r)private" calls may no longer be required.
+        self._exec_cmd("sudo mount --make-rprivate /")
```
petrutlucian94, line 665: There's also the option of limiting the scope to a given directory, although that would also have to be a mount, e.g.

```
sudo mkdir -p /mnt/coriolis
# bind /mnt/coriolis to itself so that we can modify propagation rules
sudo mount --bind /mnt/coriolis  /mnt/coriolis
sudo mount --make-private /mnt/coriolis
```

But then we'd have to ensure that all Coriolis mounts go under that path and also avoid rebinding it multiple times.

For simplicity we can stick with `sudo mount --make-rprivate /`.

Dany9966 (approved):

